Issue #1, opened 2024-04-26
MIgrate codebase to `github.com/boozallen` repository


d-ryan-ashcraft:
Migrate the aissemble baseline codebase into this github.com/boozallen repository.

This ticket will should allow building of the repository, but will not cover some of the coordination activities activities like the build and release processes.  Those will follow quickly once the codebase is brought over.

d-ryan-ashcraft:
Initial code pushed - we have a few tweaks to make, then this subsequent updates will come as part of other tickets.

d-ryan-ashcraft:
We're working through a few issues that are typical when moving large codebases.  They are:
- some test files were filtered out the commit due to `.gitignore` conflicts (ACTIVELY BEING WORKED)
- documentation links not working until the GitHub Pages job is completed (WORK STARTED - #5)

We'll continue to update this with items and statuses until complete.

d-ryan-ashcraft:
Successful initial build.  This is not indicative of everything being perfect just yet, but the initial code migration is now completed.  Additional issues will be flowed to resolve any outstanding issues in addition to those already added (e.g., #2, #5).
```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for aiSSEMBLE::Foundation::Maven Plugins::Artifact Plugin 1.7.0-SNAPSHOT:
[INFO] 
[INFO] aiSSEMBLE::Foundation::Maven Plugins::Artifact Plugin SUCCESS [ 19.609 s]
[INFO] aiSSEMBLE::Foundation::Core Java ................... SUCCESS [ 12.990 s]
[INFO] aiSSEMBLE::Foundation::Core Python ................. SUCCESS [ 38.247 s]
[INFO] aiSSEMBLE::Foundation::Messaging ................... SUCCESS [  0.099 s]
[INFO] aiSSEMBLE::Foundation::Messaging Java .............. SUCCESS [  8.030 s]
[INFO] aiSSEMBLE::Foundation::Messaging::Python ........... SUCCESS [  0.067 s]
[INFO] aiSSEMBLE::Foundation::Messaging::Python::Service .. SUCCESS [  9.827 s]
[INFO] aiSSEMBLE::Foundation::Messaging Python::Client .... SUCCESS [01:21 min]
[INFO] aiSSEMBLE::Foundation::Lineage ..................... SUCCESS [  0.167 s]
[INFO] aiSSEMBLE::Foundation::Data Lineage ................ SUCCESS [  0.058 s]
[...]

PR #10, opened 2024-04-29
#1 add missed files during migration

ewilkins-csi:
Due to some tricky ordering of exclusions in our .gitignore file many of the test files for the mda-maven-plugin were ignored during our initial migration.  Moved nearly all exclusions to the tail of the .gitignore file to prevent future mishaps except one instance where the ordering was clearly intentional. Also removed .gitignore from the ignored list so updates to it are easily picked up.


csun-cpointe (approved):

d-ryan-ashcraft (approved):
Looks good - thanks @ewilkins-csi!

PR #19, opened 2024-04-30
#9 enable ConfigLoader to read/write config to vault

### LINKED ISSUE #9: Enable ConfigLoader to write/read config properties to vault server
A part of a series to implement the Universal Configurations, we want the foundation-configuration-store module ConfigLoader class be able to read/write loaded configuration properties to the vault server.

**Definition of Done:**
- Add to foundation-configuration-store the PropertyDAO class that abstracts away logic for writing properties to Vault
- Ensure that configurations loaded by the ConfigLoader class can write to the Vault server
- Ensure written configurations can be read from Vault by the ConfigLoader
- Add an unit test that stands up a vault test container as a vault server where we can write/read configuration properties
- BDD Scenario:
```
Feature: Configuration can be stored to Vault
  
  Scenario: The ConfigLoader can read and write to the Vault server
    Given set of properties to be loaded
    When the configurations are loaded
    And the properties can be read from the vault server
```

csun-cpointe:
Closes #9 to enable the ConfigLoader class to read/write configuration properties to the vault store.

### DIFF
--- a/extensions/extensions-configuration-store-vault/pom.xml
+++ b/extensions/extensions-configuration-store-vault/pom.xml
@@ -1,0 +1,93 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>com.boozallen.aissemble</groupId>
+        <artifactId>extensions</artifactId>
+        <version>1.7.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>extensions-configuration-store-vault</artifactId>
+
+    <name>aiSSEMBLE::Extensions::Configuration::Store::Vault</name>
+    <description>Provides centralized secure storage of configurations</description>
+    <packaging>jar</packaging>
+    <properties>
+        <version.vault>5.1.0</version.vault>
+    </properties>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-source-plugin</artifactId>
+            </plugin>
+        </plugins>
+    </build>
+    <profiles>
+        <profile>
+            <id>integration-test</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-failsafe-plugin</artifactId>
+                        <version>${version.failsafe.plugin}</version>
+                        <configuration>
+                            <systemPropertyVariables>
+                                <version.aissemble.vault>${project.version}</version.aissemble.vault>
+                            </systemPropertyVariables>
+                        </configuration>
+                    </plugin>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                            <systemPropertyVariables>
+                                <version.aissemble.vault>${project.version}</version.aissemble.vault>
+                            </systemPropertyVariables>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+    </profiles>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.cucumber</groupId>
+            <artifactId>cucumber-java</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.cucumber</groupId>
+            <artifactId>cucumber-junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.bettercloud</groupId>
+            <artifactId>vault-java-driver</artifactId>
+            <version>${version.vault}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.boozallen.aissemble</groupId>
+            <artifactId>extensions-encryption-vault-java</artifactId>
+            <version>${version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>vault</artifactId>
+            <version>1.19.7</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.boozallen.aissemble</groupId>
+            <artifactId>foundation-configuration-store</artifactId>
+            <version>${version}</version>
+        </dependency>
+    </dependencies>
+</project>
--- a/extensions/extensions-configuration-store-vault/src/main/java/com/boozallen/aissemble/configuration/dao/VaultPropertyDao.java
+++ b/extensions/extensions-configuration-store-vault/src/main/java/com/boozallen/aissemble/configuration/dao/VaultPropertyDao.java
@@ -1,0 +1,155 @@
+package com.boozallen.aissemble.configuration.dao;
+
+/*-
+ * #%L
+ * aiSSEMBLE::Extensions::Configuration::Store::Vault
+ * %%
+ * Copyright (C) 2021 Booz Allen
+ * %%
+ * This software package is licensed under the Booz Allen Public License. All Rights Reserved.
+ * #L%
+ */
+
+import com.bettercloud.vault.Vault;
+import com.bettercloud.vault.VaultConfig;
+import com.bettercloud.vault.VaultException;
+import com.bettercloud.vault.api.mounts.MountPayload;
+import com.bettercloud.vault.api.mounts.MountType;
+import com.bettercloud.vault.response.LogicalResponse;
+import com.bettercloud.vault.response.MountResponse;
+import com.boozallen.aissemble.configuration.exception.PropertyDaoException;
+import com.boozallen.aissemble.configuration.store.Property;
+import com.boozallen.aissemble.data.encryption.VaultUtil;
+import com.boozallen.aissemble.data.encryption.config.DataEncryptionConfiguration;
+import org.aeonbits.owner.KrauseningConfigFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * VaultPropertyDao reads and writes configuration property to Vault server
+ */
+public class VaultPropertyDao implements PropertyDao {
+    private static final Logger logger = LoggerFactory.getLogger(VaultPropertyDao.class);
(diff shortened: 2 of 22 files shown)


foundation/foundation-configuration-store/pom.xml
```diff
@@ -13,13 +13,25 @@
     <name>aiSSEMBLE::Foundation::Configuration::Store</name>
     <description>Provides centralized secure storage of configurations</description>
     <packaging>jar</packaging>
-
```
d-ryan-ashcraft, line 16: S (Substantive): I don't think we want anything solution specific in foundation.  As such, I think we'd want to put this in something like 'extensions-configuration-store-vault`?

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/dao/PropertyDao.java
```diff
@@ -0,0 +1,150 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class PropertyDao implements Dao<Property> {
+    private static final Logger logger = LoggerFactory.getLogger(PropertyDao.class);
+
+    @Override
+    public boolean checkEmpty() {
+        return false;
```
aaron-gary, line 40: Q: I'm not familiar with the full interface, but what is the logic behind always returning false here?

csun-cpointe: yeah.. should have put a comment but this function is not implemented yet so for now, it only returns false.

aaron-gary: Ok, given that - you could add a comment or throw an "unimplemented exception"

aaron-gary (approved):

extensions/extensions-configuration-store-vault/src/main/java/com/boozallen/aissemble/configuration/dao/PropertyDao.java
```diff
@@ -0,0 +1,150 @@
...
+    private void write(Vault vault, List<Property> properties) throws VaultException {
+        for (Property property : properties) {
+            // path: properties/<<group name>>
+            String path = String.format("properties/%s", property.getGroupName());
+            // create a property secret engine
+            createPropertiesSecretEngine(vault, path);
+
+            Map<String, Object> data = new HashMap<>();
+            data.put(property.getName(), property.getValue());
+            logger.info(String.format("Write to vault with properties %s: %s", path, data));
```
ewilkins-csi, line 110: C: I don't think we want to log the property value, in case it's sensitive.

extensions/extensions-configuration-store-vault/src/main/java/com/boozallen/aissemble/configuration/dao/PropertyDao.java
```diff
@@ -0,0 +1,150 @@
...
+        } else {
+            throw new PropertyDaoException("Unable to write to vault with empty properties");
+        }
+
+    }
+
+    private void write(Vault vault, List<Property> properties) throws VaultException {
+        for (Property property : properties) {
+            // path: properties/<<group name>>
+            String path = String.format("properties/%s", property.getGroupName());
```
ewilkins-csi, line 104: A: I think we should use a less generic name for the top-level.  The vault instance may be used for more than just aissemble-based functionality. Maybe `aissemble-properties`.

extensions/extensions-configuration-store-vault/src/main/java/com/boozallen/aissemble/configuration/dao/PropertyDao.java
```diff
@@ -0,0 +1,150 @@
...
+     * @throws PropertyDaoException
+     */
+    @Override
+    public void write(Property property) throws PropertyDaoException {
+        if (property != null) {
+            try {
+                final Vault vault = getVault();
+                this.write(vault, Collections.singletonList(property));
+            } catch (VaultException e) {
+                throw new PropertyDaoException(String.format("Unable to write to vault with property: %s.", property), e);
```
ewilkins-csi, line 74: C: Same note about not logging values.  The `toString` of `Property` includes the value. I think we probably should just remove the value from `toString` all together to reduce the risk of leaking information, but regardless we shouldn't rely on it here.  Just print the property group/name directly.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/dao/Dao.java
```diff
@@ -0,0 +1,43 @@
...
+import com.boozallen.aissemble.configuration.exception.PropertyDaoException;
+
+import java.util.Set;
+
+public interface Dao<T> {
+
+    boolean checkEmpty();
+
+    /**
+     * Read T from Vault with given group name and property name
```
ewilkins-csi, line 22: A: Not Vault-specific

csun-cpointe: yeah.. this has been corrected locally. Let me push it.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/dao/Dao.java
```diff
@@ -0,0 +1,43 @@
...
+     * @param groupName group name
+     * @param propertyName property name
+     * @return T
+     */
+    T read(String groupName, String propertyName);
```
ewilkins-csi, line 27: Q: I'd originally thought this was an interface for _any_ DAO we may need for the config store but this seems specific to properties.  If it's only an interface for properties, why the type parameter? Do we anticipate Property sub-types?

extensions/extensions-configuration-store-vault/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.bettercloud</groupId>
+            <artifactId>vault-java-driver</artifactId>
+            <version>${version.vault}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.boozallen.aissemble</groupId>
+            <artifactId>extensions-encryption-vault-java</artifactId>
```
ewilkins-csi, line 60: Q: Is this for the `VaultUtil` class?

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/dao/Dao.java
csun-cpointe: It should be generic. yeah.. I think so. From the design we do have another LimitationMetaDataDao, which has the groupName and propertyName as well. ref: https://confluence.boozallencsn.com/display/AIOPS/Universal+Configurations?preview=/443073804/458738303/class-diagram.drawio.png

extensions/extensions-configuration-store-vault/src/main/java/com/boozallen/aissemble/configuration/dao/VaultPropertyDao.java
```diff
@@ -0,0 +1,153 @@
...
+
+        return vault;
+    }
+
+    private MountResponse createPropertiesSecretEngine(Vault vault, String path) throws VaultException {
+        MountResponse response = null;
+        try {
+            response = vault.mounts().read(path);
+        } catch (Exception e) {
+            logger.warn("Configuration properties secret engine may not exist on vault");
```
ewilkins-csi, line 142: Q/A: Would this always be expected on the first run, or is the expected case that it's already configured before the service is ever used?  If the former, we might downgrade this to `INFO`.

extensions/extensions-configuration-store-vault/src/test/java/com/boozallen/aissemble/configuration/dao/VaultPropertyDaoSteps.java
```diff
@@ -0,0 +1,189 @@
...
+        // Override secrets.root.key values from the aissemble-vault image for VaultPropertyDao vault access.
+        String rootKey = vaultContainer.copyFileFromContainer("/root_key.txt", this::inputStreamToString);
+        encryptProperties.setProperty("secrets.root.key", rootKey);
+
+        // Override secrets.root.key values from the aissemble-vault image for VaultPropertyDao vault access.
+        String unsealKeysJSONArrayString = vaultContainer.copyFileFromContainer("/unseal_keys.txt", this::inputStreamToString);
+        encryptProperties.setProperty("secrets.unseal.keys", unsealKeysJSONArrayString.replaceAll("[\\s|\\[\\]\"]", ""));
+    }
+
+    private String getSystemArchitectureTag() {
```
ewilkins-csi, line 171: A: Unused?

extensions/extensions-configuration-store-vault/src/test/java/com/boozallen/aissemble/configuration/dao/VaultPropertyDaoSteps.java
```diff
@@ -0,0 +1,189 @@
...
+        properties.add(new Property("messaging", "connector", "smallrye-kafka"));
+        properties.add(new Property("messaging", "serializer", "apache.StringSerializer"));
+        properties.add(new Property("messaging", "topic", "messaging-topic"));
+
+        return properties;
+    }
+
+    private void setupVaultContainer() {
+        final Properties encryptProperties = Krausening.getInstance().getProperties("encrypt.properties");
+        final DockerImageName vaultImage = DockerImageName.parse("aiops-docker-internal.nexus.boozallencsn.com/aiops/boozallen/aissemble-vault:1.7.0-SNAPSHOT-arm64").asCompatibleSubstituteFor("vault");
```
ewilkins-csi, line 154: S: We've had a lot of issues in the past with setting up vault in this way.  On CI, the image is typically built with the architecture-specific tag, but locally it's not.  That means we end up pulling an older version of the image from the repository instead of using/testing the just-built image.  I recall specifically the python IT for vault encryption was updated very recently to account for this.

extensions/extensions-configuration-store-vault/pom.xml
csun-cpointe: yeah.. for both VaultUtil and DataEncryptionConfiguration

extensions/extensions-configuration-store-vault/src/test/java/com/boozallen/aissemble/configuration/dao/VaultPropertyDaoSteps.java
csun-cpointe: yeah.. this is where the `unused` function supposed to be used.. forgot to replace it.

extensions/extensions-configuration-store-vault/src/test/java/com/boozallen/aissemble/configuration/dao/VaultPropertyDaoSteps.java
csun-cpointe: updated

foundation/foundation-configuration-store/pom.xml
csun-cpointe: Moved the store to vault specific logic to the extensions-configuration-store-vault

ewilkins-csi (approved):

PR #25, opened 2024-04-30
As a data engineer, I want the default memory for Spark applications to be increased so it works out of the box for more than just toy examples

tianliang0038:
…plications to be increased so it works out of the box for more than just toy examples


JeffreyRoss:
Q:I don’t think we want to include the AIOPS tag in commit notes anymore?

tianliang0038:
> Q:I don’t think we want to include the AIOPS tag in commit notes anymore?

Thank you! I've removed it!

Issue #23, opened 2024-04-30
Update default memory for Spark


tianliang0038:
As a data engineer, I want the default memory for Spark applications to be increased so it works out of the box for more than just toy examples

tianliang0038:
**DOD**
* Investigate and address memory limit issues in spark pipelines/pods, known for frequent failures due to memory constraints.
** we should assess how additional memory setting will impact the minimum recommended settings so it's not overburdening the machine, and change the memory based on that 
* Update templates for the pipelines
 ** Pyspark data delivery pipeline
 ** Spark java data delivery pipeline
 * Validate the fix by adjusting memory settings and confirming resolution
 * Update baton migration table within the release notes (abandoned)
 * Create baton migration for updating spark application to the new values. (abandoned)
{code}
Feature: Migrate a spark application to use the new memory value
  Scenario: Migrate a spark application with new memory value in the base values
    Given a project that has a spark application
    And the base value contains the old memory value
    When the 1.7.0 spark application memory migration executes
    Then the memory is updated to new value

  Scenario: Skip spark application memory migration with non-default values in the base values
    Given a project that has a spark application
    And the base values.yaml does not have default memory values
    When the 1.7.0 spark application memory migration executes
    Then the spark application memory migration is skipped
{code}

*Test For 1.7.0-SNAPSHOT spark pipelines* (spark and pyspark)
 ** Pull the latest baseline project
 ** Run
{code:java}
mvn clean install -f foundation/foundation-mda
{code}
# Build a pipeline with the attached PysparkPipeline.json and SparkPipeline.json file to <YOUR-PIPELINE-NAME>-pipeline-models/src/main/resources/pipelines
{code}
mvn archetype:generate -B -DarchetypeGroupId=com.boozallen.aissemble \
                          -DarchetypeArtifactId=foundation-archetype \
                          -DarchetypeVersion=1.7.0-SNAPSHOT \
                          -DartifactId=aiops-3232 \
                          -DgroupId=org.test \
                          -DprojectName='Test AIOPS-3232' \
                          -DprojectGitUrl=test.org/aiops-3232\
&& cd aiops-3232 {code}
# Run the following:
{code}
mvn clean install{code}
# Follow the manual action and then run 
{code} mvn clean generate-sources{code}
[...]

PR #30, opened 2024-05-01
#23:As a data engineer, I want the default memory for Spark applicatio…

tianliang0038:
…ns to be increased so it works out of the box for more than just toy examples

change

### DIFF
--- a/foundation/foundation-mda/src/main/resources/templates/deployment/spark-operator/spark-application-base-values.yaml.vm
+++ b/foundation/foundation-mda/src/main/resources/templates/deployment/spark-operator/spark-application-base-values.yaml.vm
@@ -69,7 +69,7 @@ sparkApp:
     driver:
       cores: 1
       coreLimit: "1200m"
-      memory: "512m"
+      memory: "2048m"
       env:
         - name: KRAUSENING_BASE
           value: /opt/spark/krausening/base
@@ -90,7 +90,7 @@ sparkApp:
       #end
     executor:
       cores: 1
-      memory: "512m"
+      memory: "4096m"
       env:
         - name: KRAUSENING_BASE
           value: /opt/spark/krausening/base
--- a/DRAFT_RELEASE_NOTES.md
+++ b/DRAFT_RELEASE_NOTES.md
@@ -1,29 +1,28 @@
 # Major Additions
 
-* Python modules were renamed to reflect aiSSEMBLE. These include the following.
-
-| Old Python Module                          | New Python Module                                              |
-|--------------------------------------------|----------------------------------------------------------------|
-| foundation-core-python                     | aissemble-core-python                                          | 
-| foundation-model-training-api              | aissemble-foundation-model-training-api                        | 
-| foundation-versioning-service              | aissemble-foundation-versioning-service                        |
-| foundation-drift-detection-client          | aissemble-foundation-drift-detection-client                    |
-| foundation-encryption-policy-python        | aissemble-foundation-encryption-policy-python                  |
-| foundation-model-lineage                   | aissemble-foundation-model-lineage                             |
-| foundation-data-lineage-python             | aissemble-foundation-data-lineage-python                       |
-| foundation-messaging-python-client         | aissemble-foundation-messaging-python-client                   |
-| foundation-pdp-client-python               | aissemble-foundation-pdp-client-python                         |
-| foundation-transform-core-python           | aissemble-Foundation-transform-core-python                     |
-| extensions-model-training-api-sagemaker    | aissemble-extensions-model-training-api-sagemaker              |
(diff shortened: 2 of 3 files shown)


carter-cundiff:
Update commit message to being with `#23`

foundation/foundation-upgrade/src/test/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/SparkMemoryUpgradeMigrationSteps.java
```diff
@@ -0,0 +1,60 @@
...
+import org.apache.commons.io.FileUtils;
+import com.boozallen.aissemble.upgrade.migration.AbstractMigrationTest;
+
+import io.cucumber.java.en.Given;
+import io.cucumber.java.en.Then;
+import io.cucumber.java.en.When;
+
+public class SparkMemoryUpgradeMigrationSteps extends AbstractMigrationTest {
+
+    private SparkMemoryUpgradeMigration sparkmigration;
```
carter-cundiff, line 28: Can remove as it's no longer used

carter-cundiff (approved):

foundation/foundation-mda/src/main/resources/templates/deployment/spark-operator/spark-application-base-values.yaml.vm
```diff
@@ -69,7 +69,7 @@ sparkApp:
     driver:
       cores: 1
       coreLimit: "1200m"
```
d-ryan-ashcraft, line 71: S (Substantive): maybe I'm off, but my impression is that the driver doesn't need nearly as much memory in general as the nodes?  

When we work with memory settings, they quickly add up across a cluster.  So I think we want the defaults to be as small as we can reasonably make them, but no smaller.  How did we arrive at these values?

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/SparkMemoryUpgradeMigration.java
```diff
@@ -0,0 +1,84 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import com.boozallen.aissemble.upgrade.migration.AbstractAissembleMigration;
+import com.boozallen.aissemble.upgrade.pojo.AbstractYamlObject;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.List;
+
+public class SparkMemoryUpgradeMigration extends AbstractAissembleMigration {
```
ewilkins-csi, line 22: C: I don't think we want this migration. If a project has been running just fine with 512 MB, we don't need to bump them up. If they haven't been then they'll have already changed it to whatever value works for them.

foundation/foundation-mda/src/main/resources/templates/deployment/spark-operator/spark-application-base-values.yaml.vm
ewilkins-csi: I think Ryan is commenting on our memory changes and just happened to attach it to the coreLimit line.  I think it's a very fair point.  We still have seen issues with driver memory being too low at 512m, but It might make sense to halve the memory assignment here to 2g to match what we've seen as the bare minimum that works on all machines for what amounts to no-op pipelines.

FWIW, I think we tried to do a survey of what most projects were configuring here to arrive at the 4g, but I'm not sure if we were careful to distinguish between driver/executor, or if the projects we looked at cared to distinguish.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/SparkMemoryUpgradeMigration.java
tianliang0038: the logic here is updating the memory if it's default value (512m), if the user has set other values, then likely they've changed it to whatever value works for them, we'll skip memory update. Shall we remove this upgrade even if users haven't changed it and assume that 512m runs just fine?

foundation/foundation-mda/src/main/resources/templates/deployment/spark-operator/spark-application-base-values.yaml.vm
tianliang0038: We've set the default value for base-values.yaml to 4g as most downstream projects use 4g. To buffer this, we set 2g for dev-values.yaml for local. If it's still too much allocation, we can remove the 4g setting.

ewilkins-csi (approved):

Issue #15, opened 2024-04-30
Migrate downstream projects to use the updated Python module naming convention


jacksondelametter:
In aiSSEMBLE, the following Python modules were renamed to reflect aiSSEMBLE.

- foundation-core-python &rarr; aissemble-foundation-core-python
- foundation-model-training-api &rarr; aissemble-foundation-model-training-api
- foundation-versioinig-service &rarr; aissemble-foundation-versioning-service
- foundation-drift-detection-client &rarr; aissemble-drift-detection-client
- foundation-enryption-policy-python &rarr; aissemble-encryption-policy-python
- foundation-model-lineage &rarr; aissemble-foundation-model-lineage
- foundation-data-lineage-python &rarr; aissemble-foundation-data-lineage-python
- foundation-messaging-python-client &rarr; aissemble-foundation-message-python-client
- foundation-pdp-client-python &rarr; aissemble-pdp-client-python
- foundation-transform-core-python &rarr; aissemble-foundation-transfor-core-python
- extensions-model-training-api-sagemaker &rarr; aissemble-extensions-model-training-api-sagemaker
- extensions-data-delivery-spark-py &rarr; aissemble-extensions-data-delivery-spark-py
- extensions-encryption-vault-python &rarr; aissemble-extensions-encryption-vault-python
- extensions-transform-spark-python &rarr; aissemble-extensions-transform-spark-python
- test-data-delivery-pyspark-model &rarr; aissemble-test-data-delivery-pyspark-model
- test-data-delivery-pyspark-model-basic &rarr; aissemble-test-data-delivery-pyspark-model-basie
- machine-learning-inference &rarr; aissemble-machine-learning-inference
- machine-learning-training &rarr; aissemble-machine-learning-training
- machine-learning-training-base &rarr; aissemble-machine-learning-training-base
- machine-learning-sagemaker-training &rarr; aissemble-machine-learning-sagemaker-training

A migration feature needs to be added in order for current downstream projects to use the new Python modules when upgrading aiSSEMBLE.

akanshamishra617:
Hi there,

I'm interested in contributing to the development of a migration feature for this module renaming.

I've noticed several core Python modules within aiSSEMBLE have been renamed to include the "aiassemble-" prefix. This is a great step for clarity and organization.

To ensure a smooth upgrade path for existing projects, a migration feature seems essential. I'd be happy to help develop this feature.

Here are some initial thoughts:

[...]

PR #31, opened 2024-05-01
#15: Adding Poetry Python package migration in foundation-upgrade

jacksondelametter:


foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/PoetryMigration.java
```diff
@@ -0,0 +1,118 @@
+package com.boozallen.aissemble.upgrade.migration;
```
carter-cundiff, line 1: Can move this to the `v1_7_0` package/directory since it's only needed for this one upgrade. Same with the test files and spec.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/PoetryMigration.java
```diff
@@ -0,0 +1,118 @@
...
+        "extensions-encryption-vault-python", 
+        "extensions-data-delivery-spark-py", 
+        "foundation-data-lineage-python"
+    );
+
+    @Override
+    protected boolean shouldExecuteOnFile(File file) {
+        if(file == null || !file.exists()) {
+            logger.error("Unable to read file {} to check if migration should be executed", file.getAbsolutePath());
+        }
```
carter-cundiff, line 50: This can be removed as `baton` only passes valid file references.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/PoetryMigration.java
```diff
@@ -0,0 +1,118 @@
...
+            Set<String> dependencies = dependenciesOpt.get().valueMap().keySet();
+            shouldExecute = hasOldPackages(dependencies);
+        }
+        else {
+            logger.warn("Could not get dependencies for file {}", file.getAbsolutePath());
+        }
+        return shouldExecute;
+    }
+
+    private boolean hasOldPackages(Set<String> packages) {
```
carter-cundiff, line 65: A: Would be more readable to change `packages` to `currentPackages`.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/PoetryMigration.java
```diff
@@ -0,0 +1,118 @@
...
+            }
+        }
+        return hasOldPackages;
+    }
+
+    @Override
+    protected boolean performMigration(File file) {
+        if(file == null || !file.exists()) {
+            logger.error("Unable to read file {} for migration", file.getAbsolutePath());
+        }
```
carter-cundiff, line 80: Same as above, null check can be removed.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/PoetryMigration.java
```diff
@@ -0,0 +1,118 @@
...
+        boolean performedSuccessfully = false;
+        FileConfig pyproject = FileConfig.of(file);
+        pyproject.load();
+        Optional<Config> dependenciesOpt = pyproject.getOptional(DEPENDENCIES_KEY);
+        if(dependenciesOpt.isPresent()) {
+            try {
+                performedSuccessfully = migrateOldPythonPackages(dependenciesOpt.get(), file);
```
carter-cundiff, line 87: S: Instead of reading in the file and getting the dependencies again, you get add an attribute to the class `private Config config;` that you set in your `shouldExecuteOnFile` and can then access in this functions. Would remove a lot of the duplicate code.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/PoetryMigration.java
```diff
@@ -0,0 +1,118 @@
...
+            logger.warn("Could not get dependencies for file {}", file.getAbsolutePath());
+        }
+        return performedSuccessfully;
+    }
+
+    private boolean migrateOldPythonPackages(Config config, File pyproject) throws IOException {
+        boolean success = false;
+        for(String oldPackageName : OLD_PYTHON_PACKAGES) {
+            Optional<String> packageName = config.getOptional(oldPackageName);
+            if(packageName.isPresent()) {
```
carter-cundiff, line 102: Q: Can this be reduced to one line?
```
config.getOptional(oldPackageName).isPresent()
```

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/PoetryMigration.java
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * 
+ * Baton migration used to migrate aiSSEMBLE python packages to the new naming convention
+ */
+public class PoetryMigration extends AbstractAissembleMigration {
+
+    private static final Logger logger = LoggerFactory.getLogger(PoetryMigration.class);
+    protected static final String DEPENDENCIES_KEY = "tool.poetry.dependencies";
+    protected static final String PYTHON_PACKAGE_PREFIX = "aissemble-";
+    protected static final String UPGRADE_VERSION = "1.7.0.*";
```
carter-cundiff, line 34: S: We want to avoid hardcoding this version in if possible. If this release ends up being cut as `1.7.1` then this will be incorrect. Otherwise we have to remember to change this value after each failed release.

You could read in the version from the pom file (see `extensions/extensions-encryption/extensions-encryption-vault-java/pom.xml` for an example). Then based on that read in version, construct the `UPGRADE_VERSION`.

d-ryan-ashcraft (commented):
I'd venture that migrating a package name is probably something that we'll want a lot.  Might be worth creating a generic migration here that can be extended or externally configured to layer in the version it applies to as well as package names to migrate.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/PoetryMigration.java
jacksondelametter: Moved.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/PoetryMigration.java
jacksondelametter: Removed.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/PoetryMigration.java
jacksondelametter: Changed.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/PoetryMigration.java
jacksondelametter: Removed.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/PoetryMigration.java
jacksondelametter: I removed the Optional logic.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/PoetryMigration.java
jacksondelametter: Changed.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/PoetryMigration.java
jacksondelametter: This feature has been added.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/AbstractPoetryMigration.java
```diff
@@ -0,0 +1,95 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.boozallen.aissemble.upgrade.util.FileUtils;
+import com.boozallen.aissemble.upgrade.util.PythonPackageMigration;
+import com.electronwill.nightconfig.core.Config;
+import com.electronwill.nightconfig.core.file.FileConfig;
+
+/**
+ *
```
d-ryan-ashcraft, line 27: A: super tiny:
*  remove empty comment line 
* end the [first sentence of javadocs w/ a period](https://checkstyle.org/checks/javadoc/summaryjavadoc.html) (helps w/ IDE and Javadoc preview integration)

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/AbstractPoetryMigration.java
```diff
@@ -0,0 +1,95 @@
...
+                break;
+            }
+        }
+        return hasOldPackages;
+    }
+
+    @Override
+    protected boolean performMigration(File file) {
+        boolean performedSuccessfully = false;
+        FileConfig pyproject = FileConfig.of(file);
```
d-ryan-ashcraft, line 65: S: does @carter-cundiff's prior comment around just loading the file once still apply here?  I would think it would.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/AbstractPoetryMigration.java
```diff
@@ -0,0 +1,95 @@
...
+        pyproject.load();
+        Config dependencies = pyproject.get(DEPENDENCIES_KEY);
+        try {
+            performedSuccessfully = migrateOldPythonPackages(dependencies, file);
+        } catch (IOException e) {
+            logger.error(e.getMessage());
```
d-ryan-ashcraft, line 71: S: if we get and exception, we should fail hard IMO.  Wrap in a `BatonException` and re-throw.

d-ryan-ashcraft, line 71: S: if we get and exception, we should fail hard IMO.  Wrap in a `BatonException` and re-throw. 

```suggestion
            throw new BatonException("Could not migrate file: " + file.getPath(), e);
```

d-ryan-ashcraft (approved):
Looks good - a few minor comments inline.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/AbstractPoetryMigration.java
carter-cundiff: After thinking about it some more, I decided it was best not to set an attribute from a file in `shouldExecuteOnFile` and then use that attribute within `performMigration`. Didn't think it was safe to assume the same file would be passed between those two functions consecutively.

carter-cundiff: Okay seems my assumption was wrong. Looking closer at how baton executes, it looks like `shouldExecuteOnFile` and `performMigration` execute on the same file consecutively. So you should be safe to set a `Config config` object in your `shouldExecuteOnFile` and then reference that in your `performMigration`.

jacksondelametter: I'm not sure it would be a good idea to load it once in the shouldExecuteOnFile. Although at the moment it looks like Baton executes the performMigration immediately after the shouldExecuteOnFile, if this ever changed this would not work correctly and would not be obvious.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/AbstractPoetryMigration.java
jacksondelametter: Fixed.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/AbstractPoetryMigration.java
jacksondelametter: Throws a BatonException.

jacksondelametter: Throws a BatonException.

PR #33, opened 2024-05-01
 #22 Removed maven-clean-plugin version from pom generation

### LINKED ISSUE #22: Downstream projects *-deploy pom references invalid plugin version
When creating a new downstream project, its *-deploy module contains a reference to an old plugin version. This is now invalid and will cause failure when upgrade aissemble. It is also unnecessary because the aissemble-parent defines the version in the plugin management.
```
<plugin>
    <artifactId>maven-clean-plugin</artifactId>
    <version>${version.clean.plugin}</version>
</plugin>
``` 
We need to correct this and create a migration feature for current downstream projects to not use the old version property.

cwoods-cpointe:
Update the release notes with the breaking change

closes #22

### DIFF
--- a/foundation/foundation-mda/src/main/resources/templates/notifications/notification.deploy.clean.app.target.vm
+++ b/foundation/foundation-mda/src/main/resources/templates/notifications/notification.deploy.clean.app.target.vm
@@ -1,6 +1,5 @@
             <plugin>
                 <artifactId>maven-clean-plugin</artifactId>
-                <version>${version.clean.plugin}</version>
                 <configuration>
                     <filesets>
                         <fileset>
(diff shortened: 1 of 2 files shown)


foundation/foundation-mda/src/main/resources/templates/notifications/notification.deploy.clean.app.target.vm
```diff
@@ -1,6 +1,5 @@
             <plugin>
                 <artifactId>maven-clean-plugin</artifactId>
-                <version>${version.clean.plugin}</version>
```
csun-cpointe, line 3: Q: should this be `${version.maven.clean.plugin}` instead of removing it?

csun-cpointe (approved):

foundation/foundation-mda/src/main/resources/templates/notifications/notification.deploy.clean.app.target.vm
d-ryan-ashcraft: The version is inherited from the parent pom, so it isn't needed at all here.

d-ryan-ashcraft (approved):

JeffreyRoss (approved):

Issue #8, opened 2024-04-29
Require Docker image build to fail if any Python requirement fails to install


Cho-William:
We want to ensure that the docker image builds fails if any python dependency installation fails. In particular, when using `find` or `for-loops` in a dockerfile's instruction to install `requirements.txt` or `.tgz` or `.whl` files, a failed Python dependency installation will not cause the docker build to fail. This faulty behavior can be fixed by prefixing the relevant docker instruction with `set -e &&` and avoiding usage of the `find` command to ensure any non-zero exit code will cause the shell environment to exit. See below examples:

```dockerfile
# faulty behavior
RUN find /tmp/requirements -path '/tmp/requirements/*/*' -name requirements.txt -type f -exec python3 -m pip install --no-cache-dir -r '{}' ';'
RUN cd /modules/${trainingModule}; for x in *.whl; do pip install $x --no-cache-dir --no-deps; done

# fail fast behavior
RUN set -e && files=$(find /tmp/requirements -path '/tmp/requirements/*/*' -name requirements.txt -type f); for file in $files; do python3 -m pip install --no-cache-dir -r $file || exit 1; done;
RUN set -e && cd /modules/${trainingModule}; for x in *.whl; do pip install $x --no-cache-dir --no-deps; done
```

**Definition of Done**

- develop and test a docker modification that will cause the docker image build to fail if any python installation cannot be completed from the requirements.txt or wheels or tarballs 
- update the baseline so that all relevant docker files in other modules have this "fail fast" behavior
- identify all relevant dockerfile templates to also be updated
- write a baton migration to add to downstream docker files any needed changes to implement this behavior 
- update release notes to list the new baton migration

csun-cpointe:
Test passed with test steps:

**Test New Project Dockerfile Generation**

1. Pull the latest branch and build the following:
```
mvn clean install -pl :foundation-mda -Dmaven.build.cache.enabled=false
mvn clean install -pl :foundation-upgrade -Dmaven.build.cache.enabled=false
```

2. Create a downstream project from the archetype using the following command:
`mvn archetype:generate -DarchetypeGroupId=com.boozallen.aissemble -DarchetypeArtifactId=foundation-archetype -DarchetypeVersion=1.7.0-SNAPSHOT -DgroupId=com.example -DartifactId=example -DprojectGitUrl=url -DprojectName=example && cd example`

3. Add the attached PysparkPipeline.json to example-pipeline-models/src/main/resources/pipelines/

[...]

PR #34, opened 2024-05-02
#8 Require docker builds to fail if Python installation fails + migration

Cho-William:
- updated relevant source docker files in `extensions-docker` that we manage in the baseline to implement "fail fast" behavior with installing python dependencies
- updated templates accordingly
- added migration to update downstream dockerfiles

### DIFF
--- a/foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/DockerPipInstallMigration.java
+++ b/foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/DockerPipInstallMigration.java
@@ -1,0 +1,139 @@
+package com.boozallen.aissemble.upgrade.migration.v1_7_0;
+
+/*-
+ * #%L
+ * aiSSEMBLE::Foundation::Upgrade
+ * %%
+ * Copyright (C) 2021 Booz Allen
+ * %%
+ * This software package is licensed under the Booz Allen Public License. All Rights Reserved.
+ * #L%
+ */
+
+import com.boozallen.aissemble.upgrade.migration.AbstractAissembleMigration;
+import com.boozallen.aissemble.upgrade.util.FileUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class DockerPipInstallMigration extends AbstractAissembleMigration {
+
+    private static final Logger logger = LoggerFactory.getLogger(DockerPipInstallMigration.class);
+    // starts with a negative look-ahead to ensure we are not matching lines that already contain
+    // "RUN set -e &&"
+    private static final String forDoPattern = "^(?!.*RUN set -e &&)RUN .*for .+; do .+; done";
+    // () denote various capture groups to separate relevant portions of the regex match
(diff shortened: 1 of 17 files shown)


peter-mcclonski (approved):

csun-cpointe (approved):

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/DockerPipInstallMigration.java
```diff
@@ -0,0 +1,143 @@
...
+    // "RUN set -e &&"
+    private static final String forDoPattern = "^(?!.*RUN set -e &&)RUN .*for .+; do .+; done";
+    private static final String findExecPattern =
+            "(^RUN )(find )(.*)( -exec)(.*python.* install (?:-\\S*\\s)*)(.+)";
+    private static final String RUN = "RUN ";
+    private static final String RUN_WITH_SET_E = "RUN set -e && ";
+    private static final String FILES_FIND = "files=$(find ";
+    private static final String FOR_FILE_IN_FILES_DO = "); for file in $files; do";
+    private static final String FILE_OR_EXIT_1_DONE = "$file || exit 1; done;";
+    private static final String upgradableInstructionsPattern =
```
ewilkins-csi, line 44: A: Is there a reason some of these are named like constants with all caps and some aren't?

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/util/FileUtils.java
```diff
@@ -48,6 +49,43 @@ public static boolean replaceInFile(File file, String regex, String replacement)
         return replacedInFile;
     }
 
+    /**
+     *
```
ewilkins-csi, line 53: A: We should provide a short description here if we're going to add java-docs.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/DockerPipInstallMigration.java
```diff
@@ -0,0 +1,143 @@
...
+     * @param dockerFile the file to check
+     * @return boolean denoting if the dockerfile has at least one instruction to be migrated
+     */
+    @Override
+    protected boolean shouldExecuteOnFile(File dockerFile) {
+        boolean shouldExecute = false;
+
+        if (dockerFile != null && dockerFile.exists()) {
+            try {
+                ArrayList<String> upgradableDockerInstructions = getRegExMatches(
```
ewilkins-csi, line 57: A: Feels like overkill to find all of the matches just to figure out if at least one match exists.  Could just compile the pattern inline to a matcher and set `shouldExecute = matcher.find()`.  As a side note, we should probably not use a static method import here as it makes the code a bit harder to follow.  We tend to do it in tests because of the sheer number of times you have to call some static method from the Assert library, but here it's much less justified.

ewilkins-csi (approved):

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/DockerPipInstallMigration.java
Cho-William: so I have the regex patterns a camel case, and literal string replacements in the all caps. that was the logic, but I can unify if that's better

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/DockerPipInstallMigration.java
Cho-William: That makes sense. Updating now

PR #39, opened 2024-05-02
#9 fix the build issue 

csun-cpointe:
The surefire plugin might will run by default so pull the maven runs surefire configuration to the base level and only enable the test in the integration test profile


d-ryan-ashcraft:
Please make sure the commit message has the issue number in it.

carter-cundiff (approved):

cwoods-cpointe (approved):

Issue #2, opened 2024-04-26
Setup GitHub Action for building aissemble


d-ryan-ashcraft:
As the aissemble codebase is migrated into the github.com/aissemble repository, we should create a continuous integration (CI) build process that lives with it and is available at the same level of visibility.  

This process should migrate the internal process used in the past to GitHub Actions.

We should ensure that caching is enabled for faster builds, minimally:
   - Maven repository caching
   - Maven build caching
   - Docker caching

PR #42, opened 2024-05-03
#2 :green_heart: update OCI configuration to get helm chart publishing

d-ryan-ashcraft:


extensions/extensions-helm/pom.xml
```diff
@@ -105,18 +110,39 @@
...
-                        <skipInit>true</skipInit>
                         <appVersion>${project.version}</appVersion>
                         <chartDirectory>${project.basedir}</chartDirectory>
                         <namespace>default</namespace>
                         <templateGenerateName>true</templateGenerateName>
                         <skip>${skip.helm}</skip>
-                        <skipTemplate>false</skipTemplate>
                         <skipDependencyBuild>true</skipDependencyBuild>
+                        <!-- For OCI, do not init: -->
+                        <skipInit>false</skipInit>
```
ewilkins-csi, line 140: Q: There seems to be a mismatch between the comment and the code here. Which do we mean?

ewilkins-csi (approved):

extensions/extensions-helm/pom.xml
chang-annie: I think this should be set to true right?

aaron-gary (approved):

chang-annie (approved):
looks good! just need to address comment regarding `skipInit`

extensions/extensions-helm/pom.xml
d-ryan-ashcraft: Yep - my bad.  We DO need the init, but it needs to NOT have ghcri.io.  Will remove the comment.

extensions/extensions-helm/pom.xml
```diff
@@ -130,10 +156,7 @@
                         </uploadRepoSnapshot>
                         <lintStrict>true</lintStrict>
                         <helmExtraRepos>
-                            <helmRepo>
-                                <name>${aissemble.helm.repo.name}</name>
-                                <url>${aissemble.helm.repo}</url>
-                            </helmRepo>
+                            <!-- No ghcr.io repo here b/c it is oci:// -->
```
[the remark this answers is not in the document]
d-ryan-ashcraft: No - you cannot add an oci repo - instead, you fully quality the link.  This has to do w/ OCI being a direct link to a specific file (no yaml descriptor or search functionality).

Issue #48, opened 2024-05-08
Rename Data Lineage module with latest aiSSEMBLE landscape


habibimoiz:
**Description:** 
Rename Data Lineage module to conform to the latest aiSSEMBLE landscape.

**Definition of Done:** 
- Rename the Data Lineage package.
- Rename all the child directories, classes, and methods.
- Update Data Lineage aiSSEMBLE Docs documentation with renamed changes.  
- Correct all the imports inside of the velocity templates to ensure a successful build of the Data Lineage module. 

**Steps to Test:** 

**Test 1:** 
- Run the following 
- mvn clean install
- Ensure you receive a successful build. 

**Test 2:**
- Ensure Data Lineage downstream project is able to be build successfully.

PR #60, opened 2024-05-10
#48 - Rename Data Lineage module with latest aiSSEMBLE landscape PR

habibimoiz:
#48 - Rename Data Lineage module with latest aiSSEMBLE landscape
  - Renamed all the child directories, classes, and methods.
  - Updated Data Lineage aiSSEMBLE Docs documentation with renamed changes.
  - Corrected all the imports inside of the velocity templates to ensure a successful build of the Data Lineage module.
  - Checking in updated pyproject.toml files (latest dependencies from Habushu upgrade).


ewilkins-csi (changes requested):
C: These changes look good, but we are going to need a baton migration to update the package imports.

d-ryan-ashcraft:
C: Please make sure that you start the commit message with "#48" as well - the branch name will disappear, so this is what will be left for easier provenance tracking.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/util/DataLineageUtil.java
```diff
@@ -0,0 +1,74 @@
...
+            String content = Files.lines(path).map(line -> line.replaceAll(textToReplace, replacementText)).collect(Collectors.joining("\n"));
+
+            Files.write(path, content.getBytes());
+            System.out.println("Content has been replaced in " + file.getName());
+
+        } catch (IOException e) {
+            System.err.println(e.getMessage());
+        }
+    }
+    public static String extractAissembleVersion(String line) {
```
ewilkins-csi, line 46: S: This logic makes no sense to me, as imports don't have embedded versions.  It's also not obvious that this is even being used anywhere.  Given the operation we're doing with this migration, I am skeptical we even need a Util class for this. Why can't we use the existing `FileUtils` methods for search and replace and just embed the constants of this file into the migration itself?

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/util/DataLineageUtil.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+public class DataLineageUtil {
+
+    public static String VERSION_AISSEMBLE_POST_1_7_0 = "1.7.0";
+
+    public static final String VERSION_AISSEMBLE_POST_1_7_0_REGEX = "com\\.boozallen\\.aissemble\\.data\\.lineage";
+
+    public static final String VERSION_AISSEMBLE_PRE_1_7_0_REGEX = "com\\.boozallen\\.aiops\\.data\\.lineage";
```
ewilkins-csi, line 27: S: There's no reason for this to be regex since we're matching literal characters.  I'd just do literal string based searching instead.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/util/DataLineageUtil.java
```diff
@@ -0,0 +1,74 @@
...
+
+    public static final String EQUALS = "=";
+
+    public static void replaceTextInFile(File file, String textToReplace, String replacementText) {
+        Path path = file.toPath();
+
+        System.out.println("Text to Replace: " + textToReplace +"Replacement Text: " + replacementText);
```
ewilkins-csi, line 34: S: We shouldn't be printing to stdout directly.  Using a logger gives downstream consumers more control.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/DataLineageAiopsReferencesMigration.java
```diff
@@ -0,0 +1,85 @@
...
+            while ((line = dataLineageConfig.readLine()) != null) {
+                Pattern pattern = Pattern.compile(VERSION_AISSEMBLE_PRE_1_7_0_REGEX);
+                Matcher matcher = pattern.matcher(line);
+
+                if (matcher.find()) {
+                    shouldExecute = true;
+                }
+            }
+        } catch (Exception e) {
+            throw new BatonException(e);
```
ewilkins-csi, line 55: S: Provide an exception message.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/DataLineageAiopsReferencesMigration.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.regex.Pattern;
+
+import static com.boozallen.aissemble.upgrade.util.DataLineageUtil.*;
+import static com.boozallen.aissemble.upgrade.util.FileUtils.replaceInFile;
+
+public class DataLineageAiopsReferencesMigration extends AbstractAissembleMigration {
+
+    public static final Logger logger = LoggerFactory.getLogger(DataLineageAiopsReferencesMigration.class);
+
+    protected String getAissembleVersion() {
```
ewilkins-csi, line 31: A: unused?

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/DataLineageAiopsReferencesMigration.java
```diff
@@ -0,0 +1,85 @@
...
+        boolean shouldExecute = false;
+        try (BufferedReader dataLineageConfig = new BufferedReader((new FileReader(file)))) {
+
+            String line;
+            while ((line = dataLineageConfig.readLine()) != null) {
+                Pattern pattern = Pattern.compile(VERSION_AISSEMBLE_PRE_1_7_0_REGEX);
+                Matcher matcher = pattern.matcher(line);
+
+                if (matcher.find()) {
+                    shouldExecute = true;
```
ewilkins-csi, line 51: A: Short-circuit so we stop reading the file as soon as we identify it needs to be migrated. Either with a direct `return` here or (my preference) adding a `&& !shouldExecute` to the `while` condition.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/DataLineageAiopsReferencesMigration.java
```diff
@@ -0,0 +1,85 @@
...
+     * @return isMigrated - Whether the file was migrated successfully.
+     */
+    @Override
+    protected boolean performMigration(File file) {
+        boolean isMigrated = false;
+
+        try (BufferedReader dataLineageConfig = new BufferedReader((new FileReader(file)))) {
+            String line;
+            while ((line = dataLineageConfig.readLine()) != null) {
+                Pattern pattern = Pattern.compile(VERSION_AISSEMBLE_PRE_1_7_0_REGEX);
```
ewilkins-csi, line 72: I: Already noted I don't think we need a regex, but just for info; pattern compilation can be slow so often it's useful to compile a pattern once and re-use the pattern object.  Really would likely have negligible impact here since we don't _anticipate_ very large files, but if you were trying to apply a pattern hundreds of thousands of times you'd start to see the advantage more clearly.

foundation/foundation-upgrade/src/test/resources/specifications/v1_7_0/data-lineage-migration.feature
```diff
@@ -0,0 +1,11 @@
+Feature: Data Lineage upgrade migration
+
+  Scenario: Migrate the Data Lineage module files with the new package name when the VERSION_AISSEMBLE is less than the 1.7.0.
```
ewilkins-csi, line 3: A: This scenario is written oddly to me. There is no variable `VERSION_AISSEMBLE` that we're updating. Scenarios should focus on the functional goals.  I might rewrite these to be something more like:
```
Scenario: Update Data Lineage package name
  Given a Java file references the package "com.boozallen.aiops.data.lineage"
  When the 1.7.0 data lineage package migration executes
  Then the references are updated to "com.boozallen.aissemble.data.lineage"
```

foundation/foundation-upgrade/src/main/resources/migrations.json
```diff
@@ -91,6 +91,15 @@
           }
         ]
       },
+      {
+        "name": "data-lineage-package-import-migration",
+        "implementation": "com.boozallen.aissemble.upgrade.migration.v1_7_0.DataLineageAiopsReferencesMigration",
+        "fileSets": [
+          {
+            "includes": ["*-pipelines/*/src/*/java/*/cdi/*.java"]
```
ewilkins-csi, line 99: C: We need to migrate all Java files not just those in `cdi` packages of pipelines.  A project could easily be referencing this package through some custom lineage consumer.  Recommend changing this to `src/**/*.java`.

foundation/foundation-upgrade/src/main/resources/migrations.json
```diff
@@ -91,6 +91,15 @@
           }
         ]
       },
+      {
+        "name": "data-lineage-package-import-migration",
```
ewilkins-csi, line 95: Q/S: Shouldn't this be in the 1.7.0 group?

foundation/foundation-upgrade/src/main/resources/migrations.json
```diff
@@ -109,6 +118,15 @@
             ]
           }
         ]
+      },
+      {
+        "name": "data-lineage-package-import-migration",
+        "implementation": "com.boozallen.aissemble.upgrade.migration.v1_7_0.DataLineageAiopsReferencesMigration",
+        "fileSets": [
+          {
+            "includes": ["*-pipelines/*/src/*/java/com/boozallen/cdi/*.java"]
```
ewilkins-csi, line 127: I: In Maven/Ant fileglob syntax, `**` matches multiple directories.  I already recommended a complete change to our glob in the first migration instance, but you wouldn't have needed this if you had made that includes `*-pipelines/*/src/*/java/**/cdi/*.java`

foundation/foundation-upgrade/src/test/resources/specifications/v1_7_0/data-lineage-migration.feature
```diff
@@ -0,0 +1,11 @@
...
+    Given an existing downstream project with data lineage is less than VERSION_AISSEMBLE_one_seven_zero
+    When the data lineage version migration executes
+    Then the imports are updated in the data lineage modules with the new package name
+
+  Scenario: Do not migrate the Data Lineage module files with the new package name when the imports are already correct
```
ewilkins-csi, line 8: A: I'm not sure this is necessary.  At least if you rewrite the first scenario as recommended, I think it becomes clear that any file _not_ referencing the old package name won't be migrated.

foundation/foundation-upgrade/src/test/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/DataLineageMigrationSteps.java
```diff
@@ -0,0 +1,51 @@
...
+import static junit.framework.Assert.assertFalse;
+import static junit.framework.TestCase.assertTrue;
+
+public class DataLineageMigrationSteps extends AbstractMigrationTest {
+    private AbstractAissembleMigration migration;
+
+    @Given("an existing downstream project with data lineage is less than VERSION_AISSEMBLE_one_seven_zero")
+    public void anExistingDownstreamProjectWithDataLineageIsLessThanVERSION_AISSEMBLE_one_seven_zero() {
+        testFile = getTestFile("v1_7_0/DataLineageMigration/migration/CdiContainerFactory_1_6_0.java");
+        migration = new DataLineageAiopsReferencesMigration();
```
ewilkins-csi, line 28: A: We really should be constructing this in the `when` and making it a local variable. There's nothing useful we need it for, so creating a field that gets set in a multitude of `given`s just increases complexity without adding anything.

foundation/foundation-upgrade/src/test/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/DataLineageMigrationSteps.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @When("the data lineage version migration executes")
+    public void theDataLineageVersionMigrationExecutes() {
+        performMigration(migration);
+    }
+
+    @Then("the imports are updated in the data lineage modules with the new package name")
+    public void theImportsAreUpdatedInTheDataLineageModulesWithTheNewPackageName() {
+        assertTrue("Data Lineage modules are references an older package in imports.", shouldExecute);
```
ewilkins-csi, line 44: S: This does _not_ validate the core assertion of the `then`.  Firstly, I think it's fine to check that `shouldExecute` was true, but the more important flag is the one that says the file was actually migrated.  Secondly, we need to verify that the old package references were actually updated.  There are two approaches we tend to take and I think both are valid (we may need to just make a team norm): either create a "validation" file where you manually make the changes you'd expect the migration to make and validate that the migrated file matches the validation file, or you programmatically read in and check the migrated file with specific logic.

foundation/foundation-upgrade/src/test/resources/test-files/v1_7_0/DataLineageMigration/migration/CdiContainerFactory_1_6_0.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.commons.collections4.CollectionUtils;
+import org.jboss.weld.environment.se.WeldContainer;
+
+import com.boozallen.aiops.core.cdi.CdiContainer;
+import com.boozallen.aiops.core.cdi.CdiContext;
+import com.boozallen.aissemble.alerting.core.cdi.AlertingCdiContext;
+import com.boozallen.aiops.data.lineage.cdi.DataLineageCdiSelector;
```
ewilkins-csi, line 22: A: Not totally necessary, but I might include other ways you can reference the package name in addition to the imports.  So maybe a fully qualified type for a variable or return type, and/or a string reference.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/util/DataLineageUtil.java
habibimoiz: I removed this function as I was not using it.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/util/DataLineageUtil.java
habibimoiz: Updated the logic such that it does a literal string search (removed regex pattern matching).

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/util/DataLineageUtil.java
habibimoiz: Removed the function due to no usage.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/DataLineageAiopsReferencesMigration.java
habibimoiz: Added a exception message.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/DataLineageAiopsReferencesMigration.java
habibimoiz: Removed the function due to no usage.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/DataLineageAiopsReferencesMigration.java
habibimoiz: Made this change, thanks for the recommendation! :)

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/DataLineageAiopsReferencesMigration.java
habibimoiz: Created a new function in FileUtils called replaceInFileWORegex. This will do a simple search and replace NOT using pattern matching.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/util/DataLineageUtil.java
```diff
@@ -0,0 +1,19 @@
...
+ * #%L
+ * aiSSEMBLE::Foundation::Upgrade
+ * %%
+ * Copyright (C) 2021 Booz Allen
+ * %%
+ * This software package is licensed under the Booz Allen Public License. All Rights Reserved.
+ * #L%
+ */
+
+public class DataLineageUtil {
```
ewilkins-csi, line 13: A: Remove this class. Move constants to `DataLineageAiopsReferencesMigration`

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/DataLineageAiopsReferencesMigration.java
```diff
@@ -0,0 +1,75 @@
...
+     */
+    @Override
+    protected boolean performMigration(File file) {
+        boolean isMigrated = false;
+
+        try (BufferedReader dataLineageConfig = new BufferedReader((new FileReader(file)))) {
+            String line;
+            while((line = dataLineageConfig.readLine()) !=null && !isMigrated) {
+                if (line.contains(VERSION_AISSEMBLE_PRE_1_7_0)) {
+                    replaceLiteralInFile(file, VERSION_AISSEMBLE_PRE_1_7_0, VERSION_AISSEMBLE_POST_1_7_0);
```
ewilkins-csi, line 65: C: We're reading line by line in a file, and then reading the entire file to modify it within that line-by-line loop. This entire loop should just be replaced by this single call. Don' forget to remove the `BufferedReader` from the `try` definition as well.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/util/FileUtils.java
```diff
@@ -27,13 +27,33 @@
...
+    }
+
     /**
      *
      * @param file the File
      * @param regex a regex representing the text to replace, as a String
      * @param replacement the replacement text to substitute the regex
      * @return a boolean set to true if at least one replacement was performed in the file
      */
+
```
ewilkins-csi, line 56: A: accidentally added a blank line here.

foundation/foundation-upgrade/src/main/resources/migrations.json
ewilkins-csi: You don't know that this is the only location though.  It's very likely that a project using data lineage would also add their own custom code that leverages the data lineage package.  The breakdown of `shouldExecuteOnFile` vs `performMigration` should give you what you need to address any performance issues.

foundation/foundation-upgrade/src/test/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/DataLineageMigrationSteps.java
```diff
@@ -0,0 +1,62 @@
...
+import io.cucumber.java.en.When;
+
+import java.io.File;
+import java.nio.file.Files;
+
+import static junit.framework.TestCase.assertTrue;
+
+public class DataLineageMigrationSteps extends AbstractMigrationTest {
+    private AbstractAissembleMigration migration;
+    File validatedFile = getTestFile("/v1_7_0/DataLineageMigration/validation/CdiContainerFactory_Validation.java");
```
ewilkins-csi, line 26: A:
```suggestion
    private File validatedFile = getTestFile("/v1_7_0/DataLineageMigration/validation/CdiContainerFactory_Validation.java");
```
but why not make this a local variable?

foundation/foundation-upgrade/src/test/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/DataLineageMigrationSteps.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Function to check whether 2 given files are the same.
+     * @param original first file to compare
+     * @param migrated second file to compare
+     * @return shouldExecute - whether the migration is necessary.
+     */
+    private static Boolean validateMigration(File original, File migrated) {
```
ewilkins-csi, line 34: A: Consider using `FileUtils.contentEqualsIgnoreEOL`

foundation/foundation-upgrade/src/test/resources/specifications/v1_7_0/data-lineage-migration.feature
```diff
@@ -0,0 +1,6 @@
+Feature: Data Lineage upgrade migration
+
+  Scenario: Update Data Lineage package name
+    Given a Java file references the package com.boozallen.aiops.data.lineage
+    When the one.seven.zero data lineage package migration executes
```
ewilkins-csi, line 5: Q/A: Is this spelled out because the plugin that generates the glue code methods was treating the numbers as params? If so, would recommend just manually changing it to `the 1.7 data...` in the `@When` so this is a bit more readable.

foundation/foundation-upgrade/src/main/resources/migrations.json
```diff
@@ -66,11 +66,20 @@
             ]
           }
         ]
+      },
+      {
+        "name": "data-lineage-package-import-migration",
```
ewilkins-csi, line 71: S: Add to the automatic migrations table in `DRAFT_RELEASE_NOTES` so users know how to shut this off if they don't want it.

ewilkins-csi (approved):

Issue #59, opened 2024-05-10
`aissemble-archetype` test fails due to `helm template` invocation error


d-ryan-ashcraft:
## Description 
When running the `foundation-archetype` test script, the following failure is encountered:
```
[INFO] --- exec:3.1.0:exec (main-chart) @ example-data-delivery-pipeline ---
[ERROR] Command execution failed.
org.apache.commons.exec.ExecuteException: Process exited with an error: 1 (Exit value: 1)
    at org.apache.commons.exec.DefaultExecutor.executeInternal (DefaultExecutor.java:404)
    at org.apache.commons.exec.DefaultExecutor.execute (DefaultExecutor.java:166)
    at org.codehaus.mojo.exec.ExecMojo.executeCommandLine (ExecMojo.java:1000)
    at org.codehaus.mojo.exec.ExecMojo.executeCommandLine (ExecMojo.java:956)
    at org.codehaus.mojo.exec.ExecMojo.execute (ExecMojo.java:426)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:126)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2 (MojoExecutor.java:328)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute (MojoExecutor.java:316)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:212)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:174)
    at org.apache.maven.lifecycle.internal.MojoExecutor.access$000 (MojoExecutor.java:75)
    at org.apache.maven.lifecycle.internal.MojoExecutor$1.run (MojoExecutor.java:162)
    at org.apache.maven.buildcache.BuildCacheMojosExecutionStrategy.execute (BuildCacheMojosExecutionStrategy.java:145)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:159)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:105)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:73)
    at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:53)
[...]

PR #61, opened 2024-05-10
#59 :bug: fix helm template references in downstream builds so they w…

d-ryan-ashcraft:
…ill work with oci; :arrow_up: committing recently added pylint updates

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -121,6 +121,7 @@
         <dockerImageVersion>${project.version}</dockerImageVersion>
 
         <!-- Helm chart repository -->
+        <aissemble.helm.repo.protocol>oci</aissemble.helm.repo.protocol>
         <aissemble.helm.repo>ghcr.io/boozallen</aissemble.helm.repo>
         <aissemble.helm.repo.name>ghcr.io</aissemble.helm.repo.name>
 
--- a/foundation/foundation-mda/src/main/resources/templates/data-delivery-common/pipeline.chart.pom.configuration.vm
+++ b/foundation/foundation-mda/src/main/resources/templates/data-delivery-common/pipeline.chart.pom.configuration.vm
@@ -13,9 +13,7 @@
                 <executable>helm</executable>
                 <arguments>
                     <argument>template</argument>
-                    <argument>aissemble-spark-application-chart</argument>
-                    <argument>--repo</argument>
-                    <argument>${aissemble.helm.repo}</argument>
+                    <argument>${aissemble.helm.repo.protocol}://${aissemble.helm.repo}/aissemble-spark-application-chart</argument>
                     <argument>--version</argument>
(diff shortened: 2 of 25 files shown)


build-parent/pom.xml
```diff
@@ -121,6 +121,7 @@
         <dockerImageVersion>${project.version}</dockerImageVersion>
 
         <!-- Helm chart repository -->
```
d-ryan-ashcraft, line 123: I (Informational): adding in the protocol so a downstream project has the chance to update it to it something https if they are using something like Nexus inside a VPN or project-specific repository.

foundation/foundation-mda/src/main/resources/templates/data-delivery-common/pipeline.chart.pom.configuration.vm
```diff
@@ -13,9 +13,7 @@
                 <executable>helm</executable>
                 <arguments>
                     <argument>template</argument>
```
d-ryan-ashcraft, line 15: I: you cannot use a `repo` with OCI b/c there isn't an index or way to search what is in an OCI repository

test/test-mda-models/aissemble-test-data-delivery-pyspark-model-basic/pyproject.toml
```diff
@@ -36,6 +36,7 @@ behave = ">=1.2.6"
 nose = ">=1.3.7"
 testcontainers = ">=3.7.1"
 kappa-maki = ">=1.0.1"
```
d-ryan-ashcraft, line 38: I: these were added in to to commit these changes that are from a prior update, but we didn't get a clean build.

peter-mcclonski (approved):

foundation/foundation-mda/src/main/resources/templates/data-delivery-common/pipeline.chart.pom.configuration.vm
carter-cundiff: This can be handled in a later ticket. But have we thought about how downstream project will be incorporating these changes when they upgrade?

carter-cundiff:
Also need to update this template with the new helm template command:
`foundation/foundation-mda/src/main/resources/templates/notifications/notification.spark.worker.tilt.vm`

foundation/foundation-mda/src/main/resources/templates/data-delivery-common/pipeline.chart.pom.configuration.vm
d-ryan-ashcraft: Good question - Yes. This change will handle baseline charts.  They can use our public charts or re-wire this to their own mirror/proxy (e.g., Nexus, Artifactory).  Another ticket will allow specification of a totally different repo as the project-default chart repository.

Issue #62, opened 2024-05-13
As a lineage user, I want the schemas for the custom aiSSEMBLE facets publicly available so that my OpenLineage tooling understands their structure


carter-cundiff:
## Description
We have created and defined schemas for each of our [custom facets](https://github.com/boozallen/aissemble/blob/dev/foundation/foundation-lineage/aissemble-foundation-model-lineage/src/aissemble_model_lineage/facets.py). This ticket builds aims to make those schemas publicly available and versioned.

## Definition of Done

* Migrate over existing lineage schemas into the new codebase
* Ensure schema URLs defined in the custom facets work correctly with the new Antora Docs URL

### Verify schema URLs are correct in downstream projects
* Create a downstream project from the archetype using the following command:
```
mvn archetype:generate -DarchetypeGroupId=com.boozallen.aissemble -DarchetypeArtifactId=foundation-archetype -DarchetypeVersion=1.7.0-SNAPSHOT -DgroupId=com.example -DartifactId=example-ml -DprojectGitUrl=url -DprojectName=example-ml && cd example-ml
```

* Add the attached [MlPipelineTrainingLineage.json](https://github.com/boozallen/aissemble/files/15299264/MlPipelineTrainingLineage.json)
* Run the following once:
```
mvn clean install
```

* Run the following until all the manual actions are complete
```
mvn clean generate-sources
```

* Add the following to `example-ml-docker/example-ml-training-lineage-step-docker/src/main/resources/krausening/base/data-lineage.properties`:
```
data.lineage.MlPipelineTrainingLineage.TrainingLineageStep.namespace=test
data.lineage.MlPipelineTrainingLineageInput.namespace=testkubect
```

* Run the following once:
```
mvn clean install
```

* ```
  tilt up; tilt down
  ```
 
* Once all the pods are ready, run the following in a new terminal:
```
curl --location 'http://localhost:5001/training-jobs?pipeline_step=training-lineage-step' \--header 'Content-Type: application/json' \--data '{}'
```

* Wait for the pipeline to finish running (around 10 seconds). Will show as complete on the following command:
```
kubectl get jobs
```

* Using the job name from the previous command, view the output logs using the following command:
```
kubectl logs job.batch/<JOB NAME> | grep -e HardwareDetailsRunFacet.json -e HyperparameterRunFacet.json -e MLflowRunFacet.json -e SourceCodeDirectoryJobFacet.json -e PerformanceMetricRunFacet.json
```

[...]

PR #64, opened 2024-05-13
#62 Migrate custom open lineage schemas to new repo

carter-cundiff:


peter-mcclonski (approved):

docs/modules/ROOT/attachments/HardwareDetailsRunFacet.json
```diff
@@ -0,0 +1,46 @@
+{
```
d-ryan-ashcraft, line 1: S: I'm wondering if these should live in the codebase, then get copied here?  

Imagine a scenario where you don't have access to the internet - might need to be able to get at them locally or extract them to another, available hosting location as well?

If so, can definitely be another ticket.

carter-cundiff: We determined in the previous ticket that these schemas only serve as supplemental documentation of your custom facets. They are not technically needed for your lineage to function correctly in downstream projects, so your local project does not need direct access to these schemas.

d-ryan-ashcraft:
Ok - so we aren't using them to validate?  Seems like that might be useful
(with an option to turn it off).

docs/modules/ROOT/attachments/HardwareDetailsRunFacet.json
ewilkins-csi: For clarification, the big realization was that the schemas of OL are at the very least under-utilized.  From a cursory search the major tools that support OL don't interact with schemas at all (hence calling them documentation).  This may be due to the fact that OL doesn't really push you to make schemas for custom facets and implies that the most common approach is to omit schema info.  So until major tooling starts to support and leverage schema definitions of custom facets from libraries like ours, there isn't a ton of demand for edge cases like the one you're describing.

d-ryan-ashcraft: As I tried to note below (sorry - email response got misaligned), it just feels odd for us to provide a schema that we don't even use.  Feels like we should a least run the validation ourselves if we are going to provide it.

FWIW, I think most users would get more mileage out of the straight documentation of an example json file than reading a schema.   So this might be where the rub is for me.

carter-cundiff: Agreed that the current schema's aren't particularly useful in their current state. We do have documentation outlining all our custom facets and their inputs [here](https://boozallen.github.io/aissemble/aissemble/current/model-lineage.html#_facet). Could be worthwhile to actually use our schemas for actually validating our python implementation in another ticket.

Issue #50, opened 2024-05-08
As a user, I want to turn off manual action notifications that I don't intend to use so they don't confuse newcomers to my project.


ahartwellCpointe:
**DoD:**

-  Expose a unique id for every manual action that is displayed
-  Add the ability to blacklist manual actions by their id in the project's root pom to disable them
-  Update fermenter to ignore messages based on key values
-  Need a solution for how to make message keys available to users
-  don't want to edit the messages themselves
-- ordering isn't guaranteed, so they can't be added separately
-- could be done through a separate command
-  Stretch goal: update the message keys in aissemble to make them more sensible

Scenarios:
Scenario: display unique IDs
Given a notification key to suppress
When notifications are displayed
Then the notification indicated by the key is not output

ewilkins-csi:
Stretch goal covered by: https://github.com/boozallen/aissemble/issues/219

PR #69, opened 2024-05-14
#50 bumped to latest version

ahartwellCpointe:


build-parent/pom.xml
```diff
@@ -26,7 +26,7 @@
         <version.buildnumber.plugin>3.1.0</version.buildnumber.plugin>
         <version.cucumber.reporting.plugin>5.7.5</version.cucumber.reporting.plugin>
         <version.failsafe.plugin>${version.maven.surefire.plugin}</version.failsafe.plugin>
-        <version.fermenter>2.10.2</version.fermenter>
+        <version.fermenter>2.10.3-SNAPSHOT</version.fermenter>
```
peter-mcclonski, line 29: Q: Is this worth cutting a fermenter release?

ahartwellCpointe: This is the first feature added since the previous release, as far as I know there aren't any plans for an immediate release for just this addition. That may be a question for Ryan if you want to raise the possibility of cutting one

peter-mcclonski (approved):

build-parent/pom.xml
ewilkins-csi: I think going forward we probably want to either cut the release or keep snapshot updates on a feature branch and release before merging back to dev.  Unfortunately I didn't communicate this very clearly with @ahartwellCpointe offline and ended up implying we should merge this snapshot version into dev.  We're a bit off still from trying to release 1.7 so as long as we focus on getting the follow-on work for this in and cut the fermenter release for that PR we should be fine.

ewilkins-csi: Actually. Because Maven Central doesn't allow snapshot publishing anymore, we'll need to roll this back as there's no public jar matching this version

Issue #58, opened 2024-05-10
As a consumer of aiSSEMBLE, I want to accelerate my project builds by leveraging the build cache capability.


peter-mcclonski:
## Definition of Done 

- [x] For all new and existing aissemble-based projects, the file `$PROJECT_ROOT/.mvn/maven-build-cache-config.xml` shall be generated with appropriate content to enable the build cache.
- [x] For all new and existing aissemble-based projects, the file `$PROJECT_ROOT/.mvn/extensions.xml` shall be generated with appropriate content to enable the build cache.
- [x] A baton migration shall exist such that for any Habushu-packaged module, the following content shall be present in the pom, if and only if the tag does not currently exist within the pom:
-- `<build><directory>dist</directory></build>`
- [x] All existing templates for pom files specifying Habushu packaging shall updated to include the following content:
-- `<build><directory>dist</directory></build>`
- [x] Release notes shall be updated to describe this new functionality, with a link to the official documentation for the Maven build cache.
- [x] Release notes shall be updated with the new baton migration.

### New Projects

1. Generate a new aiSSEMBLE based project using the following command:
```
mvn archetype:generate -B -DarchetypeGroupId=com.boozallen.aissemble \
                          -DarchetypeArtifactId=foundation-archetype \
                          -DarchetypeVersion=1.7.0-SNAPSHOT \
                          -DartifactId=test-project-58\
                          -DgroupId=org.test \
                          -DprojectName='Test Cache' \
                          -DprojectGitUrl=test.org/test-project-58\
&& cd test-project-58
```
2. Execute `mvn clean install -pl :test-project-58-pipeline-models`
3. Ensure the command completes successfully.
4. Execute `mvn clean install -pl :test-project-58-pipeline-models | grep cache`
5. Ensure that you see the following output:
```
[INFO] Skipping plugin execution (cached): compiler:compile
[INFO] Skipping plugin execution (cached): build-helper:add-test-source
[INFO] Skipping plugin execution (cached): build-helper:add-test-resource
[INFO] Skipping plugin execution (cached): resources:testResources
[INFO] Skipping plugin execution (cached): compiler:testCompile
[INFO] Skipping plugin execution (cached): surefire:test
[INFO] Skipping plugin execution (cached): jar:jar
```
6. Delete the created project.

### Project Upgrades

1. Generate a new aiSSEMBLE based project using the following command:
```
mvn archetype:generate -B -DarchetypeGroupId=com.boozallen.aissemble \
[...]

PR #71, opened 2024-05-14
#58 Fix a bug when attempting to migrate pom files with no build sect…

peter-mcclonski:
…ion.


foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_7_0/EnableDistOutputFolderForHabushuBuildCacheMigration.java
```diff
@@ -33,23 +33,31 @@ private boolean isHabushuPackaged(Model model) {
...
-        String line = "\t\t<directory>dist</directory>\n";
         int indent = 2;
+        InputLocation inputLocation;
+        String line;
+        if (build != null) {
+            InputLocation referenceLocation = build.getLocation("");
+            inputLocation = new InputLocation(referenceLocation.getLineNumber() + 1, referenceLocation.getColumnNumber());
+            line = "\t\t<directory>dist</directory>\n";
+        } else {
+            InputLocation referenceLocation = model.getLocation("packaging");
```
ewilkins-csi, line 57: A: This feels like the wrong location given the test POM and our usual layout in our MDA templates.  Makes more sense to me to do it before `dependencies` at the very end before the `</project>` line.  The former feels a bit more natural, but the latter is less complex as dependencies could be missing.

ewilkins-csi (approved):

Issue #72, opened 2024-05-15
Remove unneeded dependencies from PySpark test modules.


ewilkins-csi:
We have a few POM dependencies that are added to _test/test-mda-models/aissemble-data-delivery-pyspark-model_ and _...-model-basic_ that aren't used at all and I think are left overs from the old generation of the test modules.  We could try to regenerate the POMs to update them with what would be generated today, but there are enough customizations to those POMs that it's easier just to update the dependencies directly.

PR #73, opened 2024-05-15
#72 resync test pyspark poms

ewilkins-csi:
Updating dependencies in the test-mda-models pyspark POMs to match what would be generated today by `foundation-mda`, so the generated code more is more directly tested as it would be in a downstream project.

### DIFF
--- a/DRAFT_RELEASE_NOTES.md
+++ b/DRAFT_RELEASE_NOTES.md
@@ -1,5 +1,15 @@
 # Major Additions
 
+## OpenLineage Namespace Conventions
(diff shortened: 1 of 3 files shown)


d-ryan-ashcraft (approved):

DRAFT_RELEASE_NOTES.md
```diff
@@ -1,5 +1,15 @@
...
+
+## Maven Build Cache
+The [Maven Build Cache](https://maven.apache.org/extensions/maven-build-cache-extension/) is now enabled by default for new projects.  Existing projects can reference [the generation template](https://github.com/boozallen/aissemble/tree/dev/foundation/foundation-archetype/src/main/resources/archetype-resources/.mvn) to enable this functionality in their own projects.
+
+## Kafka Docker Image
```
ewilkins-csi, line 9: I: Only real new content.  Others were just moved for readability.

DRAFT_RELEASE_NOTES.md
```diff
@@ -201,43 +188,54 @@ in this example - adjust for your project, as appropriate):
 
 ## Conditional Steps
 
+## Upgrade Steps for Projects Leveraging the aiSSEMBLE Kafka Image
```
ewilkins-csi, line 191: I: New content

PR #78, opened 2024-05-16
#47 create configuration store service

### LINKED ISSUE #47: Retrieve configurations through configuration service with its relevant environment context
[47Helper.zip](https://github.com/boozallen/aissemble/files/15355422/47Helper.zip)

**DOD**
- ensure that the Config Service can respond to a rest call to get a value
- create a docker image to house the service (also enables demo)
   - can build from the aissemble-quarkus docker image
   - implement startup script that does the following:
      - starts up the REST endpoint
      - calls on the config loader to load initial configurations

**BDD Scenario:**
```
Scenario: configuration service returns the configuration value
    Given configuration service starts
     When requests a configuration property
     Then configuration value is returned
```

**Technical Notes**
- do not need worry about the ConfigServiceClient
- do not need to worry about Helm Chart

**Test Steps:**

- **OTS Only:**

```
mvn clean install -pl :foundation-configuration-store  
mvn clean install -pl :aissemble-configuration-store-service 
```
- Run `mkdir test-issue-47 && cd test-issue-47`
- Unzip the attached 47Helper.zip file
- Place the docker-compose.yaml file in the `test-issue-47` directory
- Run `mkdir resources && cd resources`
- Place the configurations folder in the resources directory
- At the test-issue-47 root directory, run `docker-compose up`
- Verify the log that all yaml files are loading
- Verify http://localhost:8080/aissemble-properties/model-training-api/AWS_SECRET_ACCESS_KEY returns the property.

csun-cpointe:
Close #47


extensions/extensions-configuration-store-vault/pom.xml
```diff
@@ -87,5 +87,10 @@
             <artifactId>foundation-configuration-store</artifactId>
             <version>${version}</version>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-resteasy-reactive</artifactId>
+            <version>2.8.3.Final</version>
```
ewilkins-csi, line 93: S: Let's use the parent POM version. I think you can just omit it as it's defined in a `dependencyManagement`, or set it to `${version.quarkus}`

extensions/extensions-docker/aissemble-configuration-store-service/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>com.boozallen.aissemble</groupId>
+        <artifactId>extensions-docker</artifactId>
+        <version>1.7.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>aissemble-configuration-store-service</artifactId>
```
ewilkins-csi, line 11: A: To me it would make sense to drop `-service` since the service is the thing running on the image, not the image itself.

foundation/foundation-configuration-store/pom.xml
```diff
@@ -19,33 +19,145 @@
...
+                        <descriptor>${project.basedir}/src/main/resources/quarkus-app.xml</descriptor>
+                    </descriptors>
+                    <finalName>quarkus</finalName>
+                    <appendAssemblyId>true</appendAssemblyId>
+                </configuration>
+            </plugin>
+            <plugin>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <!-- this config will remove the io.vertx.core.VertxException: Thread blocked issue -->
```
ewilkins-csi, line 56: We've seen this in downstream project pipelines. (Even my test project for validating my last ticket had these.) Might look at how we could make it easy for projects to find/apply this solution.

foundation/foundation-configuration-store/pom.xml
```diff
@@ -19,33 +19,145 @@
...
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
     <dependencies>
         <dependency>
-            <groupId>io.cucumber</groupId>
-            <artifactId>cucumber-java</artifactId>
+            <groupId>org.yaml</groupId>
+            <artifactId>snakeyaml</artifactId>
+            <version>2.2</version> <!-- Check for the latest version -->
```
ewilkins-csi, line 97: A: Leftover todo? Also, we use snakeyaml in some other places. Might pull it up to build-parent

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/ConfigStoreInit.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * ConfigStoreInit loads configuration properties as application starts up
+ */
+@Startup
+@ApplicationScoped
+public class ConfigStoreInit {
+    private static final Logger logger = LoggerFactory.getLogger(ConfigStoreInit.class);
+
+    @PostConstruct
+    void init() {
```
ewilkins-csi, line 35: Q: Is this `package-local` on purpose? Can it not be `private` or `protected`?

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/ConfigStoreInit.java
```diff
@@ -0,0 +1,57 @@
...
+        String environmentUri = System.getenv("ENVIRONMENT_URI");
+        try {
+            ConfigLoader configLoader = CDI.current().select(ConfigLoader.class,new Any.Literal()).get();
+            Set<Property> properties = null;
+            if (environmentUri != null && !environmentUri.isEmpty()) {
+                properties = configLoader.loadConfigs(baseUri,environmentUri);
+            } else if (baseUri != null && !baseUri.isEmpty()){
+                properties = configLoader.loadConfigs(baseUri);
+            } else {
+                logger.warn("No properties are loaded. To load properties, you need to set the environment variable BASE_URI and ENVIRONMENT_URI");
```
ewilkins-csi, line 47: S: I might actually throw an exception here.  Unless I'm missing something, isn't the whole point of this service to load these in, so if it can't do that it's failing at its primary function?

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/dao/DefaultPropertyDao.java
```diff
@@ -0,0 +1,60 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.enterprise.context.ApplicationScoped;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * DefaultPropertyDao
+ * TODO: currently this is a dummy class which can be used to check whether the request is received,
```
ewilkins-csi, line 24: Q: Is this just so tests work? I thought `VaultPropertyDao` was our initial/default implementation.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/data/LocationUriInput.java
```diff
@@ -0,0 +1,39 @@
...
+ * Copyright (C) 2021 Booz Allen
+ * %%
+ * This software package is licensed under the Booz Allen Public License. All Rights Reserved.
+ * #L%
+ */
+
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class LocationUriInput {
```
ewilkins-csi, line 16: Q: What are we doing with this? Looks like init reads directly from system environment.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/service/ConfigService.java
```diff
@@ -0,0 +1,80 @@
...
+    @GET
+    @Path("/{groupName}/{propertyName}")
+    @Produces(MediaType.APPLICATION_JSON)
+    public String getProperty(@PathParam("groupName") String groupName,
+                              @PathParam("propertyName") String propertyName) {
+        Property property = null;
+        try {
+             property = this.configLoader.read(groupName, propertyName);
+        } catch (Exception e) {
+            logger.error("Error reading property: " + e);
```
ewilkins-csi, line 50: S: Would change to this to make sure the exception is printed properly:
```suggestion
            logger.error("Error reading property: ", e);
```
Also might consider printing group/property that failed to read.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/data/LocationUriInput.java
ewilkins-csi: Ah my question was answered. I didn't realize we were implementing a `/load` method.  I think we probably want to avoid taking the URIs as input.  I'd expect `/load` just to load with the preconfigured URL.  Allowing arbitrary URIs to be passed in feels like it introduces a lot of security issues that would have to be carefully managed without providing a whole lot of value.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/store/Property.java
```diff
@@ -49,9 +49,9 @@ public void setValue(String value) {
     public String toString() {
         return "Property{" +
                 "name='" + name +
-                ", value='" + value +
-                ", groupName='" + groupName +
-                '}';
+                "', value='" + value +
```
ewilkins-csi, line 52: S: I know it was already there and we kind of touched on this in a different PR, but I wonder if we should just remove `value` from toString all together just to really prevent in silly mistakes.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/service/ConfigService.java
ewilkins-csi: Actually, in the same vein as my other comment, it might be nice to just rethrow the exception if there is no security risk in that and/or return a 404.

foundation/foundation-configuration-store/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,27 @@
...
+
+<Configuration>
+	<Appenders>
+		<Console name="Console" target="SYSTEM_OUT">
+			<PatternLayout
+				pattern="%highlight{%d{HH:mm:ss.SSS} [%t] %-5level %logger{36}.%M():%L - %msg%n}{FATAL=red blink, ERROR=red, WARN=yellow bold, INFO=grey, DEBUG=green bold, TRACE=blue}" />
+		</Console>
+	</Appenders>
+	<Loggers>
+		<Logger name="com.boozallen.aiops" level="info" additivity="false ">
```
ewilkins-csi, line 20: A: The package root we're using here is `com.boozallen.aissemble` not aiops.  We also have an ongoing effort to move all current artifacts using the aiops package over to using aissemble instead.

foundation/foundation-configuration-store/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,27 @@
...
+			<AppenderRef ref="Console" />
+		</Logger>
+		<Root level="info">
```
ewilkins-csi, line 23: A: If we're setting the root logger to `info` and the package-specific logger to `info`, that seems redundant. Do we want to set this to `error` or remove the package-specific logger?

foundation/foundation-configuration-store/src/main/resources/quarkus-app.xml
```diff
@@ -0,0 +1,29 @@
...
+  #%L
+  AIOps Docker Baseline::Versioning::Service
+  %%
+  Copyright (C) 2021 Booz Allen
+  %%
+  This software package is licensed under the Booz Allen Public License. All Rights Reserved.
+  #L%
+  -->
+<assembly
+	xmlns="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.2"
```
ewilkins-csi, line 11: A: Consider replacing tabs with spaces

foundation/foundation-configuration-store/src/test/java/util/ConfigStoreInitTestHelper.java
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * ConfigStoreInit loads configuration properties as application starts up
+ */
+@Startup
+@ApplicationScoped
+public class ConfigStoreInitTestHelper {
+    private static final Logger logger = LoggerFactory.getLogger(ConfigStoreInitTestHelper.class);
+
+    @PostConstruct
+    void init() {
```
ewilkins-csi, line 44: Q: Can we extend the actual init class and re-use the logic there so we don't have a separate mirrored copy of the logic specific to the tests?

foundation/foundation-configuration-store/src/test/java/util/ConfigStoreInitTestHelper.java
```diff
@@ -0,0 +1,138 @@
...
+            if (properties != null) {
+                configLoader.write(properties);
+            }
+        } catch (Exception e) {
+            logger.error("Error loading properties:");
+            e.printStackTrace();
+        }
+    }
+
+    private void prepareTestEnv() {
```
ewilkins-csi, line 68: Actually if this is the only difference, can we not achieve it with a Cucumber `@Before` method?

foundation/foundation-configuration-store/src/test/resources/specifications/load-configurations.feature
```diff
@@ -16,4 +16,10 @@ Feature: Load configurations at specified URI based on environment context
   Scenario: The properties are not distinguishable
     Given URIs pointing to nondistinct configurations
     When the configurations are loaded
-    Then an exception is thrown stating configurations are not distinguishable
\ No newline at end of file
+    Then an exception is thrown stating configurations are not distinguishable
+
+  @config-service
+  Scenario: configuration service returns the configuration value
+      Given the configuration service starts
```
ewilkins-csi, line 23: A: Do we want to make this `...service has started` instead of `...service starts` to differentiate it from the when?

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/service/ConfigService.java
```diff
@@ -0,0 +1,80 @@
...
+            Set<Property> properties;
+            if (environmentUri == null || environmentUri.isEmpty()) {
+                properties = configLoader.loadConfigs(locationUriInput.getBaseUri());
+            } else {
+                properties = configLoader.loadConfigs(locationUriInput.getBaseUri(), locationUriInput.getEnvironmentUri());
+            }
+            configLoader.write(properties);
+            return PROPERTIES_LOADED;
+        } catch (Exception e) {
+            logger.error("Error loading properties from: " + uriInputs);
```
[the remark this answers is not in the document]
csun-cpointe: returning a 404 now as previous suggestion

foundation/foundation-configuration-store/src/test/java/util/ConfigStoreInitTestHelper.java
csun-cpointe: great point 👍

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/dao/DefaultPropertyDao.java
csun-cpointe: yeah.. this default property dao can be used for testing for now. Per conversation with Ryan, the vault can be the default option for now but we still need a more generic one since vault is a bit heavy for small project.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/service/ConfigService.java
csun-cpointe: throw exception would result a 500 and with an error id, return a 404 probably makes more sense

foundation/foundation-configuration-store/src/test/java/util/ConfigStoreInitTestHelper.java
csun-cpointe: yeah..unfortunately not, the `@Before` runs after the test server starts, and we load the properties when receiving the startup event.

extensions/extensions-docker/aissemble-configuration-store-service/pom.xml
csun-cpointe: I was thinking this is hosting the service but yeah, this can be a generic configuration store.

foundation/foundation-configuration-store/src/main/resources/log4j2.xml
csun-cpointe: great catch!

foundation/foundation-configuration-store/src/test/java/util/ConfigStoreInitTestHelper.java
```diff
@@ -0,0 +1,138 @@
...
+                properties = configLoader.loadConfigs(baseUri);
+            } else {
+                logger.warn("No properties are loaded. To load properties, you need to set the environment variable BASE_URI and ENVIRONMENT_URI");
+            }
+            if (properties != null) {
+                configLoader.write(properties);
+            }
+        } catch (Exception e) {
+            logger.error("Error loading properties:");
+            e.printStackTrace();
```
ewilkins-csi, line 64: A: More `printStackTrace` calls in this file that should be cleaned up (this one and one below)

foundation/foundation-configuration-store/src/test/java/util/TestPropertyDao.java
```diff
@@ -0,0 +1,47 @@
+package util;
```
ewilkins-csi, line 1: A: We should go ahead an put this and the Init Test Helper under `com.boozallen.aissemble.util`.  Not a huge deal since it's just test code that we aren't shipping, but just a bit cleaner organization.

ewilkins-csi (approved):

foundation/foundation-configuration-store/pom.xml
csun-cpointe: yeah.. did a quick look on this (add this resolution to the parent pom file but it seems still happening on other modules. There seems other issues involved for different modules.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/dao/DefaultPropertyDao.java
csun-cpointe: offline discussion and will keep this class but renamed to InMemoryPropertyDao

foundation/foundation-configuration-store/src/test/java/util/ConfigStoreInitTestHelper.java
ewilkins-csi: Spoke offline.  Because the Quarkus test framework is spinning this service up in a container, it's not super easy to access the local file system.  The juice really isn't worth the squeeze to keep trying to figure out how to do that, so we'll just stick with this technique.

Issue #66, opened 2024-05-14
As a downstream project, I want an aiSSEMBLE managed chart for Spark Infrastructure to facilitate easier adoption of future improvements.


carter-cundiff:
## Description
aiSSEMBLE is maintaining an effort to update its usage of helm charts by migrating from being generated into project instantiations to being maintained within a new extensions-helm module in the aiSSEMBLE monorepo. This ticket encompasses migrating and cleaning up the helm chart for Spark Infrastructure deployments and any interactions it has with Hive.

## Definition of Done

* Investigate publicly available charts that manage both the hive thrift server and spark-history
  * Bitnami has no spark history specific helm chart
  * Onyxia's spark-history and spark-thrift-server helm charts are both deprecated
  * The Lightbend helm chart has excellent support across cloud providers and storage solutions, but is no longer maintained and is deprecated.
* Spark History exists as a single optional chart
  * Add a basic initcontainer to dynamically retrieve spark history jar dependencies based on requirements specified in the spark-operator helm chart.
  * Load `spark.conf` from a configmap
  * Default spark-events to a PVC which can be configured by the user to work with any CSI driver as needed.
  * Ensure that arbitrary volumes and volume mounts are supported

<<<<<<<<<< TICKET BREAKPOINT >>>>>>>>>>
* Create an aggregator Spark chart which pulls both Hive Metastore Service and Hive Metastore DB as dependencies, as well as optionally Spark History, and optionally Thrift Server
  * Hive Service and Hive DB are linked as a single unified chart
  * Thrift Server exists as a single optional chart
    * Default spark-warehouse to a PVC which can be configured by the user to work with any CSI driver as needed.
    * Keep the current hive-site.xml configmap as-is.
      * ALTERNATIVE: Use a library to load hive-site.xml into the configmap using the helm Files handle.
    * Verify that only the mysql connector jar is needed.  If additional jars are conditionally needed (ie: delta), add the initcontainer package resolution system from spark-history.
    * Ensure that arbitrary volumes and volume mounts are supported.
* Update spark-application chart to default to mounting the PVC for spark-events and spark-warehouse.
  * Update existing baton migration to account for this
* Remove redundant and unused configuration properties from each chart.
* Disable automatic generation of s3local when generating the spark chart on v2.
* Migration instructions shall be written to move to the v2 chart

## Test Steps

1. Clone the latest version of the aiSSEMBLE on the `feature/spark-history` branch
[...]

PR #80, opened 2024-05-17
#66 spark charts v2

peter-mcclonski:


extensions/extensions-helm/aissemble-spark-history-chart/examples/values-s3.yaml
```diff
@@ -0,0 +1,19 @@
+eventVolume:
+  enabled: false
+
+dependencies:
+  packages:
+    - org.apache.hadoop:hadoop-aws:3.3.4
+
+deployment:
```
carter-cundiff, line 8: Would recommend mirroring the env methods supported on operator: https://github.com/kubeflow/spark-operator/blob/master/docs/user-guide.md#specifying-environment-variables

extensions/extensions-helm/aissemble-spark-history-chart/templates/local-pv.yaml
```diff
@@ -0,0 +1,20 @@
...
+  accessModes:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  capacity:
+    storage: {{ .Values.eventVolume.size }}
+  hostPath:
+    path: {{ .Values.eventVolume.volumePathOnNode }}
+    type: DirectoryOrCreate
+  persistentVolumeReclaimPolicy: Delete
+  storageClassName: spark-event-storage
```
carter-cundiff, line 19: Should this be conditional like it is in `events-pvc.yaml`?

extensions/extensions-helm/aissemble-spark-history-chart/examples/values-local.yaml
```diff
@@ -0,0 +1,10 @@
+# This file contains the default values for the aiSSEMBLE Spark Operator Helm chart.
+
+service:
+  type: LoadBalancer
+
+ingress:
+  enabled: false
+
+eventVolume:
+  enabled: true
```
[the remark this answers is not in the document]
peter-mcclonski: I can, but it's not strictly necessary since local is the default.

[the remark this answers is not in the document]
carter-cundiff: Would be more clear since it's the example

extensions/extensions-helm/aissemble-spark-history-chart/templates/local-pv.yaml
peter-mcclonski: No.  The name here is hardcoded so that if local storage is selected, it can be immediately plugged in without any additional configuration necessarily being required.  You'll notice that the condition on this template as a whole matches the condition for setting the `events-pvc-yaml` `storageClassName` to the value we see here.

extensions/extensions-helm/aissemble-spark-history-chart/examples/values-s3.yaml
carter-cundiff: Yea that makes sense. What about including these options though:
```
    envFrom:
      - configMapRef:
          name: env-config-map
      - secretRef:
          name: env-secret
```

carter-cundiff: Where you can reference a secret/config map as a whole

peter-mcclonski: I think that's a reasonable ask.  This PR is already going to be massive, though-- I'd rather tackle that as separate follow-ons.

carter-cundiff:
Thought we wanted to merge this into a feature branch?

peter-mcclonski:
> Thought we wanted to merge this into a feature branch?

You're right. I forgot to change the target. Will modify it now.

JeffreyRoss (approved):

extensions/extensions-helm/aissemble-spark-history-chart/values.yaml
```diff
@@ -0,0 +1,58 @@
+# This file contains the default values for the aiSSEMBLE Spark Operator Helm chart.
```
carter-cundiff, line 1: Update comment

extensions/extensions-helm/aissemble-spark-history-chart/values.yaml
```diff
@@ -0,0 +1,58 @@
...
+  metadata:
+    annotations: {}
+  ingressClassName: nginx
+  hosts:
+    - paths:
+        - path: /
+          pathType: Prefix
+          backend:
+            service:
+              name: spark-history-service
```
carter-cundiff, line 42: Since the name of the service will always be `{{ .Release.Name }}-service`, there's no point in being able to set it here. We know what the name will be in the ingress template right?

peter-mcclonski: Actually hear me out-- We accept a list of ingress hosts and paths (do we need to?), and hardcoding it to .Release.Name in the template would force _all_ ingress hosts and paths to point at the same service.  The question is, do we actually need to support multiple ingresses out of the box, or should we just support the expected one?

carter-cundiff: I think that makes sense to accept a list of ingress and point them all to the same service

extensions/extensions-helm/aissemble-spark-history-chart/templates/ingress.yaml
```diff
@@ -0,0 +1,26 @@
...
+  name: {{ .Release.Name }}-ingress
+  {{- with .Values.ingress.metadata.annotations }}
+  annotations:
+  {{- toYaml . | nindent 8 }}
+  {{- end }}
+spec:
+  ingressClassName: {{ .Values.ingress.ingressClassName }}
+  rules:
+    {{- range .Values.ingress.hosts }}
+    - http:
```
carter-cundiff, line 14: Don't you need the `host` attribute?

extensions/extensions-helm/aissemble-spark-history-chart/values.yaml
peter-mcclonski: I feel less confident that it does make sense, but I'm not sure.  Forcing it all to the same service seems like it would negate the benefit of allowing multiple ingress paths.

peter-mcclonski: I almost want to hardcode _just_ the default ingress, and say that any supplemental ingresses have to be made by hand.

carter-cundiff: When would you want an ingress to not point at your service? In any case, maybe we should just make the ingress service default to `{{ .Release.Name }}-service` with the option to be overrriden.

peter-mcclonski: I think that makes a lot of sense.

peter-mcclonski: Done-- Take a look!

extensions/extensions-helm/aissemble-spark-history-chart/templates/ingress.yaml
peter-mcclonski: I don't THINK so-- Why?

carter-cundiff: We have it in all our other ingress. Thought that was how you define the ingress URL?

extensions/extensions-helm/aissemble-spark-history-chart/values.yaml
```diff
@@ -0,0 +1,57 @@
...
+    annotations: {}
+  ingressClassName: nginx
+  hosts:
+    - paths:
+        - path: /
+          pathType: Prefix
+          backend:
+            service:
+              port:
+                number: 18080
```
carter-cundiff, line 43: Could also default this to `.Values.service.port.port`?

extensions/extensions-helm/aissemble-spark-history-chart/values.yaml
carter-cundiff: Looks good!

extensions/extensions-helm/aissemble-spark-history-chart/templates/ingress.yaml
peter-mcclonski: Ah, I see it now.  Added.

extensions/extensions-helm/aissemble-spark-history-chart/values.yaml
peter-mcclonski: I'm reluctant.  We're edging closer to producing duplicate ingresses.

carter-cundiff (approved):

Issue #85, opened 2024-05-21
requests 2.32.0 and 2.32.1 are breaking the build


d-ryan-ashcraft:
The 2.32.0 (and subsequent 2.32.1) releases of `requests` is breaking the aissemble build.  The [`requests` teams is working on a fix](https://github.com/psf/requests/issues/6707), but in the meantime we'll downgrade to 2.31.0 to allow users to continue to work without issue.  

We are subscribing to the fix and once it is complete add verified, we'll roll up to the latest version ('requests` fix targeted for later today, so we'll probably target the following day to roll that into aissemble as fast-followers of adopting it).

```
[ERROR] HOOK-ERROR in before_feature: DockerException: Error while fetching server API version: Not supported URL scheme http+docker
[ERROR] tests/features/encryption.feature  HOOK-ERROR in after_feature: AttributeError: 'Context' object has no attribute 'test_container'
```

## Definition of Done
- [x] Peg all `requests` dependencies to exactly version 2.31.0
- [x] Add an NB: #XX issue link to all locations where something was added to resolve

## Test Steps 
- Run the following build command and ensure the build passes fully:
```
mvn clean install
```

PR #86, opened 2024-05-21
#85 :arrow_down: downgrade requests temporarily to work around breaki…

d-ryan-ashcraft:
…ng change


foundation/aissemble-foundation-core-python/pyproject.toml
```diff
@@ -35,6 +35,9 @@ nose = ">=1.3.7"
 kappa-maki = ">=1.0.1"
 pylint = "^3.1.0"
```
d-ryan-ashcraft, line 37: I (Informational): added this to basically force just about every to stick to 2.31.0

peter-mcclonski (approved):

PR #112, opened 2024-05-28
#2 Set up GitHub Actions for building aissemble (GitHub Runner)

aaron-gary:


.github/workflows/build.yml
```diff
@@ -22,16 +22,12 @@ on:
...
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: aissemble
 
     steps:
       - uses: actions/checkout@v4
         with:
           ref: ${{ inputs.buildBranch }}
-      - name: Install Python # use direct install rather than pyenv for CI for large speed improvement
```
aaron-gary, line 31: Python is built into the Runner

.github/workflows/build.yml
```diff
@@ -62,19 +58,6 @@ jobs:
         with:
           java-version: '11'
           distribution: 'temurin'
-      - name: Install Docker
-        run: |
```
aaron-gary, line 66: Docker & Helm are built into the Runner

csun-cpointe (approved):
it seems the commit needs to be verified.

.github/workflows/build.yml
d-ryan-ashcraft: S (Substantive) Should we perhaps switch back to using pyenv in that case to also let us ensure we get the right version and can change it independently of the image?

d-ryan-ashcraft (approved):

.github/workflows/build.yml
ewilkins-csi: Looking at some of the previous builds on dev, it looks like installation of python, helm and docker takes a total of 33s.  Not sure if the community runners are optimizing those installs in some way, but if they aren't I'd rather take the 30s to allow the build script to drive the tooling entirely.

d-ryan-ashcraft: At least for pyenv, it will also cache the version of python - so slow the first time, then should be super fast after that.

I'm good with letting the build install these too.

aaron-gary: I agree that spending 30s to let GH install python is worth the wait, however, the github setup actions/installs fail on our EC2.  You would think they just work like they do on the community runners, but there seems to be an incompatibility with "Canonical, Ubuntu, 24.04 LTS, amd64 noble image build on 2024-04-23"

Error: The version '3.11.4' with architecture 'x64' was not found for Ubuntu 24.04.
  The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
 
And as far as Docker and helm, there were also some issues with the built in installs.  We may be able to address some of the issues by pre-installing a toolkit or something along those lines, but then we are customizing the EC2 anyways.

aaron-gary: As far as using Pyenv, I agree, we should use it.  Ironically it is being used (set to 3.11.4 globally), but the build command is set to ignore it.

aaron-gary:
> it seems the commit needs to be verified.

Good catch, I forgot to add my key to my profile.  It seems that adding it after the fact doesn't resolve the unverified flag for some reason.

Issue #108, opened 2024-05-24
Feature: As a project developer, I want to use the vault data access object that it's specified in the property file


csun-cpointe:
[108Helper.zip](https://github.com/boozallen/aissemble/files/15439814/108Helper.zip)

## Description
Currently, when package the configuration-store-vault with the configuration store, the vault property dao was discovered in the CDI but can't be injected properly. The vault property dao file should be injected properly so that we can use the data access object that is specified in the property file.

## DOD
Acceptance criteria required to complete the work
- [x] VaultPropertyDao class can be injected and used as it's specified in the application.properties

## Test Strategy/Script
 * **OTS Only**
```
mvn clean install -pl :foundation-configuration-store
mvn clean install -pl :extensions-configuration-store-vault
mvn clean install -pl :aissemble-configuration-store
```

 * Create a downstream project
```
mvn archetype:generate -U -DarchetypeGroupId=com.boozallen.aissemble \
                      -DarchetypeArtifactId=foundation-archetype \
                      -DarchetypeVersion=1.7.0-SNAPSHOT \
                      -DgroupId=com.boozallen.aissemble.test \
                      -DartifactId=test-issue-108 \
                      -DprojectGitUrl=git.example.com/test-issue-108 \
                      -DprojectName="Issue#108 Test"
```

 * Unzip the 108Helper.zip file.
 * Copy the SparkPipeline.json file to the `-pipeline-models/src/main/resources/pipelines` directory
 * Repeatedly execute `mvn clean install`, following the manual actions after each build until no more remain.
 * Unzip and copy the `test-issue-108-configuration-store-docker` folder to the `-docker` directory and include the module to the -docker/pom.xml file
```
    <modules>
       ...
        <module>test-issue-108-configuration-store-docker</module> 
        <!-- TODO: Add docker modules here -->
    </modules>
```

 * Run `mvn clean install`
 * Unzip and copy the configuration-store-service folder to be under the `-deploy/src/main/resources/app`
 * Include below content to the Tiltfile
```
yaml = helm(
   'test-issue-108-deploy/src/main/resources/apps/config-store-service',
   name='config-store-service',
   values=['test-issue-108-deploy/src/main/resources/apps/config-store-service/values.yaml',
       'test-issue-108-deploy/src/main/resources/apps/config-store-service/values-dev.yaml']
)
k8s_yaml(yaml)

# configuration-store
docker_build(
    ref='boozallen/test-issue-108-configuration-store-docker',
[...]

PR #113, opened 2024-05-28
#108 vault property dao class can be injected when use the extensions-configuration-store-vault module

csun-cpointe:
Ref: #108

### DIFF
--- a/extensions/extensions-configuration-store-vault/pom.xml
+++ b/extensions/extensions-configuration-store-vault/pom.xml
@@ -32,6 +32,37 @@
                     </systemPropertyVariables>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-maven-plugin</artifactId>
+                <extensions>true</extensions>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>build</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-assembly-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>single</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <descriptors>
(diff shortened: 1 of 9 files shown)


extensions/extensions-configuration-store-vault/pom.xml
```diff
@@ -86,11 +138,26 @@
             <groupId>com.boozallen.aissemble</groupId>
             <artifactId>foundation-configuration-store</artifactId>
             <version>${version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>io.quarkiverse.cucumber</groupId>
+                    <artifactId>quarkus-cucumber</artifactId>
```
ewilkins-csi, line 144: Q/S: Two thought occur to me: 1) Why are test dependencies being pulled here such that they have to be excluded, 2) why are these not the same version as is being used here (or are they being removed for some other reason)?

extensions/extensions-data-lineage/extensions-data-lineage-http-consumer-service/pom.xml
```diff
@@ -31,7 +31,7 @@
 
   <properties>
 <!--    Overriding cucumber version to align with quarkus-cucumber -->
-    <version.cucumber>7.4.1</version.cucumber>
+    <version.cucumber>${version.quarkus.cucumber.java}</version.cucumber>
```
ewilkins-csi, line 34: A: Can't we just remove this? We can set the version in dependency management of build-parent, or just inline the reference to `version.quarkus.cucumber.java`.

ewilkins-csi (approved):

extensions/extensions-configuration-store-vault/pom.xml
csun-cpointe: Yeah.. the foundation-configuration-store module needs these two packages for its tests (quarkus-cucumber test), and, for this module, we don't need to start up a webserver so we don't need the quarkus-cucumber test. Also, these two transitive dependencies are causing this module's cucumber testing not being detected because of version conflicts.

extensions/extensions-data-lineage/extensions-data-lineage-http-consumer-service/pom.xml
csun-cpointe: this is no inline reference in this module. This is an overwrite of the parent property definition.

PR #119, opened 2024-05-30
Add config to build workflow

aaron-gary:


.github/workflows/build.yml
```diff
@@ -59,7 +59,10 @@ jobs:
           java-version: '11'
           distribution: 'temurin'
       - name: Build aiSSEMBLE
-        run: ./mvnw -B clean install --file pom.xml -Pci
+        run: |
+        echo "$HOME/.pyenv/bin" >> $GITHUB_PATH
+        eval "$(pyenv init -)
```
csun-cpointe, line 64: Q: is this missing a closing `"`

csun-cpointe (approved):

.github/workflows/build.yml
aaron-gary: Yes, good catch

Issue #87, opened 2024-05-21
As a devops engineer, I want the configuration service to detect partial and complete loads so that my properties aren't reloaded when I refresh the deployment.


tianliang0038:
[87Helper.zip](https://github.com/user-attachments/files/15597010/87Helper.zip)

**Context:**
  we’ve already written the ConfigLoader tool of the config service which reads in the project’s properties and stores them in Vault. We now need to detect on start up if a) the properties were already loaded before and b) the properties were partially loaded but the process was interrupted.  
  We’d likely need some sort of indicator that we keep up-to-date — maybe a custom value in the Vault server itself that’s lists each file that was fully loaded and what the hash of the file contents were (so we know if we need to refresh a file that was previously loaded).

**DOD**
- Ensure secure handling of sensitive configurations
- Modify existing load scenarion
  - add user communication
  - recording of status
- Add new scenario for skip reloading
- Unit tests will be added to foundation-configuration-store
- Update documentation
  - reflect the configuration changes needed for the deployment in the antora docs
  - relevant baseline READMEs with any potential nuances with testing implementations

**BDD test scenarios**
```
  # Updating `Scenario: The ConfigLoader loads base and environment configurations successfully`
  # added communication to user of successful loading of properties
  # added recording of fully-loaded status
  Scenario: The Configuration Service loads configurations into an empty config store
     Given a base URI indicating a directory housing valid base configurations 
        And an environment-specific URI indicating a directory housing valid environment-specific configurations
     When the configuration service starts
       Then the configurations are loaded into the configuration store
         And the user is notified that the configurations were loaded successfully
         And the configuration service records the that the given configurations were loaded successfully

# New scenario
Scenario: the Configuration Service skips loading already fully loaded properties
     Given a base URI indicating a directory housing valid base configurations 
        And an environment-specific URI indicating a directory housing valid environment-specific configurations
        And the configuration store has been fully populated with the specified configurations
     When the configuration service starts
      Then the configuration service skips the loading process
        And notifies the user that the configurations were previously loaded 

# Potential Future Scenario Cases (atomic level: either we write everything successfully or we write nothing at all to the store)
  # partial load detected, clear existing properties, reload the proper URI (ex. URI was changed)
  # configuration store was tampered, clear existing properties, reload the proper URI
```

**Test Steps:**

1.  Create a downstream project

```
mvn archetype:generate -U -DarchetypeGroupId=com.boozallen.aissemble \
[...]

PR #120, opened 2024-05-30
#87 As a devops engineer, I want a combined manifest for the config s…

tianliang0038:
As a devops engineer, I want the configuration service to detect partial and complete loads so that my properties aren't reloaded when I refresh the deployment.

### DIFF
--- a/foundation/foundation-configuration-store/README.md
+++ b/foundation/foundation-configuration-store/README.md
@@ -22,3 +22,6 @@ properties:
 Because the `YamlProperty` and `YamlConfig` classes are required to have attributes that mirror the format of the configuration `YAML` files, there is a separate `Property` class that simply adds the `groupName` metadata to the existing information housed in `YamlProperty`. Furthermore, because an object that will house the deserialized `YAML` contents must have an empty constructor, it was not possible to have any extending relationship between a supposed `AbstractProperty` and `YamlProperty` / `Property`.
 
 See the [snakeyaml documentation](https://bitbucket.org/snakeyaml/snakeyaml/wiki/Documentation) for further details on deserializing yaml files.
+
+### Handling Load Status
+The `ConfigLoader` leverages `load-status` to detect whether properties were already fully loaded or only partially loaded due to an interrupted process. The code logic ensures that a `fully-loaded` status is set when properties are loaded successfully. This status is then used to skip loading if the properties were successfully loaded previously, thus avoiding unnecessary reloading during deployment refreshes. For details on this behavior, refer to the implementation code and comments in the `ConfigLoader` class.
--- a/foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/ConfigStoreInit.java
+++ b/foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/ConfigStoreInit.java
@@ -38,6 +38,10 @@ public class ConfigStoreInit {
         String environmentUri = System.getenv("ENVIRONMENT_URI");
         try {
             ConfigLoader configLoader = CDI.current().select(ConfigLoader.class,new Any.Literal()).get();
+            if (configLoader.isFullyLoaded()) {
+                logger.info("Properties are already fully loaded. Skipping reload.");
+                return;
+            }
             Set<Property> properties;
             if (environmentUri != null && !environmentUri.isEmpty()) {
                 properties = configLoader.loadConfigs(baseUri,environmentUri);
@@ -48,7 +52,7 @@ public class ConfigStoreInit {
             }
             configLoader.write(properties);
         } catch (Exception e) {
-            logger.error("Error loading properties:", e);
+            throw new RuntimeException("Error loading properties", e);
         }
     }
 }
--- a/foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/store/ConfigLoader.java
+++ b/foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/store/ConfigLoader.java
@@ -73,7 +73,6 @@ public class ConfigLoader {
     public Set<Property> loadConfigs(String baseURI, String environmentURI) {
         Set<Property> baseConfigs = loadURI(baseURI);
         Set<Property> environmentConfigs = loadURI(environmentURI);
-
         return reconcileConfigs(baseConfigs, environmentConfigs);
     }
 
@@ -167,8 +166,14 @@ public class ConfigLoader {
      * @param properties to be written to store
      */
     public void write(Set<Property> properties) {
-        propertyDao.write(properties);
-        logger.info("Successfully wrote all properties to the store.");
+        try {
+            propertyDao.write(properties);
+            logger.info("Successfully wrote all properties to the store.");
+            updateLoadStatus(true);
+        } catch (Exception e) {
+            logger.error("Error updating properties.", e);
+            updateLoadStatus(false);
+        }
     }
 
     /**
@@ -181,6 +186,24 @@ public class ConfigLoader {
         logger.info(String.format("Read property with groupName: %s, propertyName: %s from the store.", groupName, propertyName));
         return propertyDao.read(groupName, propertyName);
     }
+    
+    public boolean isFullyLoaded() {
+        try {
+            Property statusProperty = propertyDao.read("load-status", "fully-loaded");
+            return statusProperty != null && "true".equals(statusProperty.getValue());
+        } catch (Exception e) {
+            logger.warn("Properties are not loaded previously, continue", e);
+            return false;
+        }
+    }
+
+    private void updateLoadStatus(boolean status) {
+        Property statusProperty = new Property("load-status", "fully-loaded", String.valueOf(status));
+        try {
+            propertyDao.write(statusProperty);
+            logger.info("Successfully updated load status to: " + status);
+        } catch (Exception e) {
+            logger.error("Error updating load status.", e);
+        }
+    }
 }
-
-
--- a/foundation/foundation-configuration-store/src/test/java/com/boozallen/aissemble/configuration/store/LoadConfigurationsSteps.java
+++ b/foundation/foundation-configuration-store/src/test/java/com/boozallen/aissemble/configuration/store/LoadConfigurationsSteps.java
@@ -18,24 +18,25 @@ import io.cucumber.java.en.Then;
 import io.cucumber.java.en.And;
 
 import io.restassured.response.ValidatableResponse;
-import com.boozallen.aissemble.util.TestPropertyDao;
+import com.boozallen.aissemble.util.ConfigStoreInitTestHelper;
 
 import static io.restassured.RestAssured.given;
 import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.*;
+
 import java.util.Set;
 import java.util.HashSet;
 
-import static org.junit.Assert.*;
 
 public class LoadConfigurationsSteps {
 
     private final Property expectedProperty = new Property("messaging", "topic", "messaging-topic");
+    private final Property fullyLoadProperty = new Property("load-status", "fully-loaded", "true");
     private ConfigLoader configLoader;
     private String baseURI;
     private String environmentURI;
     private Exception foundError;
     private ValidatableResponse response;
-
 
     @Before("@config-loader")
     public void setup() {
@@ -48,12 +49,58 @@ public class LoadConfigurationsSteps {
         response = null;
     }
 
-    @Given("URIs pointing to valid base and environment configurations")
-    public void URIsPointingToValidBaseAndEnvironmentConfigurations() {
+    @Given("a base URI indicating a directory housing valid base configurations")
+    public void URIsPointingToValidBaseConfigurations() {
         baseURI = "src/test/resources/configurations/base";
-        environmentURI = "src/test/resources/configurations/example-env";
     }
 
+    @And("an environment-specific URI indicating a directory housing valid environment-specific configurations")
+    public void URIsPointingToValidEnvironmentConfigurations() {
+        environmentURI = "src/test/resources/configurations/example-env";
+    }
+    
+    @When("the configuration service starts")
+    public void theConfigurationServiceStarts() {
+        // the service is triggered automatically
+    }
+
+    @Then("the configurations are loaded into the configuration store")
+    public void theConfigurationsAreLoadedIntoConfigStore() {
+        String responseBody = getResponseBodyForProperty(expectedProperty.getGroupName(), expectedProperty.getName());
+        assertEquals(expectedProperty.toJsonString(), responseBody);
+    }
+
+    @And("the user is notified that the configurations were loaded successfully")
+    public void notifyOnSuccessConfigurationLoad() {
+        assertTrue("load successful", ConfigStoreInitTestHelper.verification.get("load successful").equals("true"));
+    }
+
+    @And("the configuration service records the that the given configurations were loaded successfully")
+    public void successStatusIsRecorded() {
+        String responseBody = getResponseBodyForProperty(fullyLoadProperty.getGroupName(), fullyLoadProperty.getName());
+        assertEquals(fullyLoadProperty.toJsonString(), responseBody);
+    }
+
+    @Given("the configuration store has been fully populated with the specified configurations")
+    public void theConfigurationsFullyLoaded(){
+        ConfigStoreInitTestHelper.loadProperties();
+    }
+
+    @Then("the configuration service skips the loading process")
+    public void theConfigurationServiceSkipLoading() {
+        assertTrue("skip reloading", ConfigStoreInitTestHelper.verification.get("skip reloading").equals("true"));
+    }
+
+    @And("notifies the user that the configurations were previously loaded")
+    public void notifyOnPreviousConfigurationFullLoad() {
+        assertTrue("skipping reloading notification sent", ConfigStoreInitTestHelper.skiploadingNotificationSent);
+    }
+
+    @Given("URIs pointing to misformatted configurations")
+    public void URIsPointingToMisformattedConfigurations() {
+        baseURI = "src/test/resources/configurations-misformatted/base";
+        environmentURI = "src/test/resources/configurations-misformatted/example-env";
+    }
     @When("the configurations are loaded")
     public void theConfigurationsAreLoaded() {
         try {
@@ -62,29 +109,6 @@ public class LoadConfigurationsSteps {
             foundError = error;
         }
     }
-
-    @Then("the ConfigLoader validates the URI and its contents")
-    public void theConfigLoaderValidatesTheURIAndItsContents() {
-        assertNull(foundError);
-    }
-
-    @And("consumes the base configurations")
-    public void consumesTheBaseConfigurations() {
-        // verification of consumption occurs after configurations are reconciled
-    }
-
-    @And("augments the base with the environment configurations")
-    public void augmentsTheBaseWithTheEnvironmentConfigurations() {
-        assertEquals(10, TestPropertyDao.loadedProperties.size());
-        assertPropertySetsEqual(createExpectedProperties(), new HashSet<>(TestPropertyDao.loadedProperties.values()));
-    }
-
-    @Given("URIs pointing to misformatted configurations")
-    public void URIsPointingToMisformattedConfigurations() {
-        baseURI = "src/test/resources/configurations-misformatted/base";
-        environmentURI = "src/test/resources/configurations-misformatted/example-env";
-    }
-
     @Then("an exception is thrown stating configurations are misformatted")
     public void anExceptionIsThrownStatingConfigurationsAreMisformatted() {
         String expectedMessage = "Could not parse yaml";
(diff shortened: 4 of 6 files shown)


foundation/foundation-configuration-store/src/test/java/com/boozallen/aissemble/configuration/store/LoadConfigurationsSteps.java
```diff
@@ -161,4 +198,53 @@ private void assertPropertySetsEqual(Set<Property> expected, Set<Property> resul
             assertTrue("Could not find " + expectedProperty, matchFound);
         }
     }
-}
+
+    private class ConfigStoreInitMock extends ConfigStoreInit {
```
csun-cpointe, line 202: Q: can we achieve this by leveraging the existing ConfigStoreInitTestHelper?

foundation/foundation-configuration-store/src/test/java/com/boozallen/aissemble/configuration/store/LoadConfigurationsSteps.java
```diff
@@ -48,43 +64,68 @@ public void cleanup() {
...
+    }
+
+    @And("an environment-specific URI indicating a directory housing valid environment-specific configurations")
+    public void URIsPointingToValidEnvironmentConfigurations() {
         environmentURI = "src/test/resources/configurations/example-env";
     }
+    
+    @When("the configuration service starts")
+    public void theConfigurationServiceStarts() {
+        configStoreInitMock.init();
```
csun-cpointe, line 79: S: we don't need to run this init() separately, it's already being initialized when the web server starts

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/store/ConfigLoader.java
```diff
@@ -73,7 +74,9 @@ public void setPropertyDao(PropertyDao propertyDao) {
     public Set<Property> loadConfigs(String baseURI, String environmentURI) {
         Set<Property> baseConfigs = loadURI(baseURI);
         Set<Property> environmentConfigs = loadURI(environmentURI);
-
+        boolean isBaseUriValid = baseURI != null && !baseURI.isEmpty();
+        boolean isEnvironmentUriValid = environmentURI != null && !environmentURI.isEmpty();
```
csun-cpointe, line 78: C: we probably don't need these since they are already being checked in the loadURI() function and if it's an empty url, the exception should've been thrown.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/store/ConfigLoader.java
```diff
@@ -167,8 +170,14 @@ private Set<Property> reconcileConfigs(Set<Property> baseConfigs, Set<Property>
...
     public void write(Set<Property> properties) {
-        propertyDao.write(properties);
-        logger.info("Successfully wrote all properties to the store.");
+        try {
+                propertyDao.write(properties);
+                logger.info("Successfully wrote all properties to the store.");
+                updateLoadStatus(fullyLoaded);
+            } catch (Exception e) {
+                logger.error("Error updating properties.", e);
+                updateLoadStatus(false);
```
csun-cpointe, line 179: Q: should we continue to throw out the exception once we capture the updateLoadStatus failed.

extensions/extensions-configuration-store-vault/src/main/java/com/boozallen/aissemble/configuration/dao/VaultPropertyDao.java
```diff
@@ -50,14 +50,18 @@ public boolean checkEmpty() {
...
             final Vault vault = getVault();
             // read from Vault
             final LogicalResponse readResponse = vault.logical()
                     .read(String.format("aissemble-properties/%s/%s", groupName, propertyName));
-
+            // Check for null responses
+            if (readResponse == null || readResponse.getData() == null || readResponse.getData().get(propertyName) == null) {
+                logger.warn("Property not found: {}/{}", groupName, propertyName);
+                return null;
+            }
```
csun-cpointe, line 64: Q: I thought exception will be thrown if server can't find the property?

foundation/foundation-configuration-store/src/test/java/com/boozallen/aissemble/configuration/store/LoadConfigurationsSteps.java
```diff
@@ -34,12 +43,19 @@ public class LoadConfigurationsSteps {
     private String baseURI;
     private String environmentURI;
     private Exception foundError;
-    private ValidatableResponse response;
+    private ValidatableResponse response; //change
```
csun-cpointe, line 46: A: forget to clean up?

foundation/foundation-configuration-store/src/test/java/com/boozallen/aissemble/configuration/store/LoadConfigurationsSteps.java
```diff
@@ -34,12 +43,19 @@ public class LoadConfigurationsSteps {
...
+    private Set<Property> properties;
+    @InjectMocks
+    private ConfigStoreInitMock configStoreInitMock;
 
     @Before("@config-loader")
     public void setup() {
+        testPropertyDao = new TestPropertyDao();
         configLoader = new ConfigLoader();
+        configLoader.setPropertyDao(testPropertyDao);
+        configStoreInitMock = new ConfigStoreInitMock();
```
csun-cpointe, line 58: A: we might be able to clean all these up since web server already loading all the properties

foundation/foundation-configuration-store/src/test/resources/specifications/load-configurations.feature
```diff
@@ -1,17 +1,22 @@
...
-    And augments the base with the environment configurations
+    Then the configurations are loaded into the configuration store
+    And the user is notified that the configurations were loaded successfully
+    And the configuration service records the that the given configurations were loaded successfully
 
-  Scenario: The configurations are misformatted
-    Given URIs pointing to misformatted configurations
-    When the configurations are loaded
-    Then an exception is thrown stating configurations are misformatted
+  # New scenario
```
ewilkins-csi, line 12: A: Should drop this comment

extensions/extensions-configuration-store-vault/src/main/java/com/boozallen/aissemble/configuration/dao/VaultPropertyDao.java
tianliang0038: Yep. However, at the config init process, if ConfigStoreInit can't find the property when reading "fully-loaded" status(since it wasn't loaded previously), we don't want it to directly fail, but rather execute write.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/ConfigStoreInit.java
```diff
@@ -48,7 +52,7 @@ protected void init() {
             }
             configLoader.write(properties);
         } catch (Exception e) {
-            logger.error("Error loading properties:", e);
+            throw new RuntimeException("Error loading propertiesL", e);
```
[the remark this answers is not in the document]
tianliang0038: fixed the typo

foundation/foundation-configuration-store/src/test/resources/specifications/load-configurations.feature
tianliang0038: done

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/store/ConfigLoader.java
tianliang0038: In updateLoadStatus, it will throw out exception if upload/write fails. So I think that should be sufficient?

foundation/foundation-configuration-store/src/test/java/com/boozallen/aissemble/configuration/store/LoadConfigurationsSteps.java
tianliang0038: thanks! changed

foundation/foundation-configuration-store/src/test/java/com/boozallen/aissemble/configuration/store/LoadConfigurationsSteps.java
tianliang0038: an error would show `"LoadConfigurationsSteps.java:[80,24] init() has protected access in com.boozallen.aissemble.util.ConfigStoreInitTestHelper"` as the void function is protected

csun-cpointe: You should be able to just use the `ConfigStoreInitTestHelper` class without creating a new test class to override the init() function of the `ConfigStoreInitTestHelper`.

tianliang0038: gotcha, changed it to using ConfigStoreInitTestHelper

foundation/foundation-configuration-store/src/test/java/com/boozallen/aissemble/configuration/store/LoadConfigurationsSteps.java
tianliang0038: hmm, we still need to call ConfigStoreInitTestHelper.init(), because the initialization isn't linked to a filestore (i.e. testPropertyDao in unit test), and the write function will be essentially void with default initialization

csun-cpointe: It should be since we specified to use the TestPropertyDao class in the application.properties file.

tianliang0038: gotcha! but how do we retrieve the same  TestPropertyDao in the test? When we want to validate anything, we need to access to TestPropertyDao object initiated in the ConfigStoreInit during initialization process. However, we don't have any getter function in ConfigStoreInit

foundation/foundation-configuration-store/README.md
```diff
@@ -21,4 +21,7 @@ properties:
...
 
-See the [snakeyaml documentation](https://bitbucket.org/snakeyaml/snakeyaml/wiki/Documentation) for further details on deserializing yaml files.
\ No newline at end of file
+See the [snakeyaml documentation](https://bitbucket.org/snakeyaml/snakeyaml/wiki/Documentation) for further details on deserializing yaml files.
+
+### Handling Load Status
+The `ConfigLoader` can leverages `load-status` to detect whether properties were already fully loaded or only partially loaded due to an interrupted process. The code logic ensures that a `fully-loaded` status is set when properties are loaded successfully. This status is then used to skip loading if the properties were successfully loaded previously, thus avoiding unnecessary reloading during deployment refreshes. For details on this behavior, refer to the implementation code and comments in the `ConfigLoader` class.
```
carter-cundiff, line 27: `The `ConfigLoader` can leverages `load-status`` -> `The `ConfigLoader` leverages `load-status``

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/store/ConfigLoader.java
```diff
@@ -181,6 +190,18 @@ public Property read(String groupName, String propertyName) {
...
         return propertyDao.read(groupName, propertyName);
     }
-}
-
+    public boolean isFullyLoaded() {
+        Property statusProperty = propertyDao.read("load-status", "fully-loaded");
+        return statusProperty != null && "true".equals(statusProperty.getValue());
+    }
 
+    private void updateLoadStatus(boolean fullyLoaded) {
```
carter-cundiff, line 198: Recommend changing this local argument to a different name than `fullyLoaded`, like `status`. Otherwise it confusing to know if `fullyLoaded` is referring to the argument or class attribute.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/store/ConfigLoader.java
```diff
@@ -167,8 +170,14 @@ private Set<Property> reconcileConfigs(Set<Property> baseConfigs, Set<Property>
      * @param properties to be written to store
      */
     public void write(Set<Property> properties) {
-        propertyDao.write(properties);
-        logger.info("Successfully wrote all properties to the store.");
+        try {
+                propertyDao.write(properties);
+                logger.info("Successfully wrote all properties to the store.");
+                updateLoadStatus(fullyLoaded);
```
carter-cundiff, line 176: If it successfully wrote, could we just do `updateLoadStatus(true);`?

tianliang0038: yes I've changed it to just updateLoadStatus(true)

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/store/ConfigLoader.java
tianliang0038: done

foundation/foundation-configuration-store/README.md
tianliang0038: fixed it

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/store/ConfigLoader.java
```diff
@@ -47,6 +47,7 @@
 public class ConfigLoader {
     private static final Logger logger = LoggerFactory.getLogger(ConfigLoader.class);
     private PropertyDao propertyDao;
+    public boolean fullyLoaded;
```
carter-cundiff, line 50: Think we don't actually need this attribute anymore?

foundation/foundation-configuration-store/pom.xml
```diff
@@ -159,5 +159,11 @@
             <artifactId>cucumber-reporting</artifactId>
             <version>${version.cucumber.reporting.plugin}</version>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>4.0.0</version>
+            <scope>test</scope>
+        </dependency>
```
csun-cpointe, line 167: A: this can also be cleaned up

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/store/ConfigLoader.java
tianliang0038: It's removed

foundation/foundation-configuration-store/src/test/java/com/boozallen/aissemble/configuration/store/LoadConfigurationsSteps.java
tianliang0038: removed those!

foundation/foundation-configuration-store/src/test/java/com/boozallen/aissemble/configuration/store/LoadConfigurationsSteps.java
tianliang0038: But directly making a request to server, we can access to properties uploaded. I've got those fixed and removed the init() call

csun-cpointe (approved):

foundation/foundation-configuration-store/pom.xml
tianliang0038: Cleaned those up!

carter-cundiff:
Add ticket number to commit message

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/service/ConfigService.java
```diff
@@ -38,14 +38,16 @@ public class ConfigService {
...
         try {
              property = this.configLoader.read(groupName, propertyName);
         } catch (Exception e) {
-            logger.error(String.format("Error reading property - groupName: %s, propertyName: %s", groupName, propertyName), e);
+            exception = e;
         }
         if (property != null) {
             return Response.ok(property.toJsonString()).build();
         }
+        logger.error(String.format("Error reading property - groupName: %s, propertyName: %s", groupName, propertyName), exception);
```
carter-cundiff, line 50: Any reason for move from the above catch statement?

csun-cpointe: I think it's because we don't throw exception for the configuration not found case anymore but we still wanna log it.

carter-cundiff: Could we still do that on line 44 like we had it originally?

tianliang0038: We made some changes to the `VaultPropertyDao's read()` function. Previously, if a property was not available in the server, the` read()` function would throw an exception. However, we need to check a potentially non-existent status property at server startup, so we modified the` read() `function to not throw an exception if the property is null. Due to this change, we also had to update the `getProperty() `function in the configuration service. Previously, `getProperty()` would throw an exception when reading a non-existent property, but now it won't. Therefore, we adjusted `getProperty()` to handle null values properly. If a property is null, it will not throw an exception, but it will still fail and display an appropriate message.

carter-cundiff: > Previously, `getProperty()` would throw an exception when reading a non-existent property, but now it won't.

Based on what used to be on line 44, it looks like the `getProperty()` function used to just `logger.error` isntead of throw an exception. So I don't understand why we're moving that same `logger.error` call to line 50 now. The functionality appears to be the same as before except now we have a `Exception exception = null` variable.

extensions/extensions-configuration-store-vault/src/main/java/com/boozallen/aissemble/configuration/dao/VaultPropertyDao.java
carter-cundiff: Could we retain the original logic of throwing an exception when a property is not found, and instead just utilize a try/catch block in the `isFullyLoaded()` function to handle that exception?

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/service/ConfigService.java
tianliang0038: gotcha! previously null case will through an exception and `logger.error` will be printed, not anymore with the change. We want to print the same msg    ```
logger.error(String.format("Error reading property - groupName: %s, propertyName: %s", groupName, propertyName));
```  in case of null on line 50. Thus we move it to line 50 so that as long as the property is not null, error msg will be printed out. Revert it back + adding a logger info on Line 50 seems cleaner, and I've reverted it back

Comment (carter-cundiff) [reply foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/service/ConfigService.java]: Think we could probably remove this logic altogether if we revert back to throwing the exception as part of `VaultPropertyDao's read()` and do the exception handling in the `isFullyLoaded()` function

Comment (tianliang0038) [reply foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/service/ConfigService.java]: Did the exception in `isFullyLoaded()` and changed it back

File: extensions/extensions-configuration-store-vault/src/main/java/com/boozallen/aissemble/configuration/dao/VaultPropertyDao.java (hunk 5 above)
Comment (tianliang0038) [reply extensions/extensions-configuration-store-vault/src/main/java/com/boozallen/aissemble/configuration/dao/VaultPropertyDao.java]: yep! I've reverted it back and use a try/catch in `isFullyLoaded`

File: extensions/extensions-configuration-store-vault/src/main/java/com/boozallen/aissemble/configuration/dao/VaultPropertyDao.java [hunk 16]
```diff
@@ -50,14 +50,13 @@ public boolean checkEmpty() {
...
      */
-    @Override
+    @Override 
     public Property read(String groupName, String propertyName) {
         try {
             final Vault vault = getVault();
             // read from Vault
             final LogicalResponse readResponse = vault.logical()
                     .read(String.format("aissemble-properties/%s/%s", groupName, propertyName));
-
```
carter-cundiff, line 60: Would clean up this file to remove the whitespace changes

tianliang0038: cleaned up the file

carter-cundiff (approved):

Issue #123, opened 2024-06-04
Feature: Release 1.7.0


ewilkins-csi:
## Description
We are ready to cut the release for 1.7.0.  Because we have not finalized out self-hosted GitHub runner, we can't run this release via GitHub actions and will continue to use the old infrastructure for release. See the [](url)

## DOD
- [ ] 1.7.0 artifacts have been published to public repositories
- [ ] 1.7.0 release notes have been published
- [ ] 1.7.0 documentation been published
- [ ] Create 1.8.0 milestone
- [ ] Move in-flight tickets to 1.8.0 milestone

## Test Strategy/Script
Verify a new project can be created, deployed, and executed from 1.7.0.

ewilkins-csi:
DOD with @csun-cpointe @carter-cundiff

ewilkins-csi:
OTS with @d-ryan-ashcraft

Cho-William:
final test passed
<img width="575" alt="1 7 0-build-success" src="https://github.com/boozallen/aissemble/assets/149192267/630f1526-ad88-45b5-be9e-f2002a7cb5a7">
<img width="1249" alt="1 7 0-deploy-success" src="https://github.com/boozallen/aissemble/assets/149192267/4a4900d5-0905-48b2-8e9d-07a7c69a692d">

PR #125, opened 2024-06-05
#123 Fix NPE when running in Tests module

ewilkins-csi:
~Now that downstream projects have the build cache enabled by default, our archetype test isn't always catching and testing new changes to the underlying dependencies. In this case, the baton migrations were cached and so were skipped, giving the impression that the new migration was functioning correctly when it just hadn't been tested. To fix this, the test script has been updated to disable the build cache for the final build of the archetype test.~

### DIFF
--- a/DRAFT_RELEASE_NOTES.md
+++ b/DRAFT_RELEASE_NOTES.md
@@ -7,10 +7,10 @@ Conventions for setting namespaces when leveraging `Data Lineage` has been updat
 The [Maven Build Cache](https://maven.apache.org/extensions/maven-build-cache-extension/) is now enabled by default for new projects.  Existing projects can reference [the generation template](https://github.com/boozallen/aissemble/tree/dev/foundation/foundation-archetype/src/main/resources/archetype-resources/.mvn) to enable this functionality in their own projects.
 
 ## Kafka Docker Image
-The baseline Kafka Docker image has moved from using the _wurstmeister/kafka_ image (which was outdated and is no longer available) to using [Bitnami's Kafka image](https://hub.docker.com/r/bitnami/kafka) as its base.  If you are using the v2 Kafka chart managed by aiSSEMBLE, it will now pull the baseline Kafka image instead of directly using the Bitnami image. If you are still on the older v1 chart, it is already using the baseline image and will be the Bitnami flavor in 1.7.0. Kafka Connect support is still included in the baseline image.
+The baseline Kafka Docker image has moved away from using the _wurstmeister/kafka_ image (which was outdated and is no longer available) to using [Bitnami's Kafka image](https://hub.docker.com/r/bitnami/kafka) as its base.  If you are using the v2 Kafka chart managed by aiSSEMBLE, it will now pull the baseline Kafka image instead of directly using the Bitnami image. If you are still on the older v1 chart, it is already using the baseline image and will be the Bitnami flavor in 1.7.0. Kafka Connect support is still included in the baseline image.
 
 ## Package Renaming
-* Python modules were renamed to reflect aiSSEMBLE. These include the following. 
+* Python modules were renamed to reflect aiSSEMBLE. These include the following.
 
 | Old Python Module                       | New Python Module                                 |
 |-----------------------------------------|---------------------------------------------------|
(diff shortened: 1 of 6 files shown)


foundation/foundation-archetype/test-project-archetype.sh
```diff
@@ -175,7 +175,7 @@ cp ../../../src/test/resources/test-project-archetype-script/maven-build-cache-c
 echo -e "\n# maven-suppress-warnings" >> Tiltfile
 
 echo -e "\nINFO: Running full build to generate project structure\n"
-mvn clean install || { echo -e '\n\n\t**** MAVEN BUILD FAILED ****\n\n' ; exit 1; }
+mvn clean install -Dmaven.build.cache.enabled=false || { echo -e '\n\n\t**** MAVEN BUILD FAILED ****\n\n' ; exit 1; }
```
ewilkins-csi, line 178: I: There are several other `mvn clean install` calls throughout the test.  I reasoned that only the first and last mattered, but that might be faulty logic. Happy to extend this to every `mvn clean install` call if that makes sense.

cwoods-cpointe (approved):

foundation/foundation-archetype/test-project-archetype.sh
carter-cundiff: My thought it why not extend it to every call. Could turn out later we were missing some bug in those middle ones that was getting cached. Not sure I fully understand the innerworkings of the cache, but no need to risk it.

carter-cundiff (approved):

foundation/foundation-archetype/test-project-archetype.sh
```diff
@@ -212,7 +212,7 @@ echo -e $fileContents > test-generator-deploy/pom.xml
 
 
 echo -e "\nINFO: Running full build to check that the build passes w/o any manual actions needed\n"
-mvn clean install || { echo -e '\n\n\t**** MAVEN BUILD FAILED ****\n\n' ; exit 1; } > maven-build.log
+mvn clean install -Dmaven.build.cache.enabled=false || { echo -e '\n\n\t**** MAVEN BUILD FAILED ****\n\n' ; exit 1; } > maven-build.log
```
ewilkins-csi, line 215: A: Updating these to use `skipCache` instead of `enabled`.

carter-cundiff:
Need to include baton migration in release notes table

foundation/foundation-archetype/test-project-archetype.sh
```diff
@@ -165,11 +165,6 @@ echo "{
       	]
       }" > test-generator-pipeline-models/src/main/resources/pipelines/ExampleMachineLearningPipeline.json
 
-# Create build cache files to make builds more efficient, when possible - testing the parts of the archetype impacted by change:
-# ExampleDataDelivery.json
-cp ../../../../../.mvn/extensions.xml .mvn/extensions.xml
-cp ../../../src/test/resources/test-project-archetype-script/maven-build-cache-config.xml .mvn/maven-build-cache-config.xml
```
ewilkins-csi, line 171: I: We now generate downstream projects with build cache config.

ewilkins-csi:
Quick recap of some off line conversations.  We really would prefer getting the cache to work _correctly_ over just skipping it.  However, going down that path turned over several issues that just aren't worth chasing down immediately but should be followed up on right after the release.

To recap some of the issues:
 - We're currently copying the aissemble build cache config into the archetype test, but now that we turn it on by default for downstream projects, we shouldn't be doing that anymore
 - We currently have src/generated and src/main/java added to the outputs for the aissemble cache config in order to get the test-mda-models to work as we wanted, but this is very inadvisable and not the solution we want to use
 - Currently, the build-cache extension does not interrogate plugin dependencies for potential changes which means baton/fermenter are sometimes skipped when they shouldn't be (See https://github.com/apache/maven-build-cache-extension/pull/146)
 - If either of the first two bullets are addressed, the archetype test will fail for subsequent builds if there is a cache hit, because fermenter is skipped and so the bulk of the source code is missing -- this seemed to be working at some point and we probably need to figure out how it was working before and what holes there are if any

Issue #127, opened 2024-06-05
Feature: As a devops engineer, I want an aissemble-managed helm chart for the Hive metastore service that uses a newer version of Hive, so I have access to the latest security fixes.


peter-mcclonski:
## Description
In order to improve usability and maintainability, we will be migrating to a v2 chart for the hive metastore service, keeping a similar usage pattern to that seen in #103.  This ticket will encompass #116 as well, to update the underlying Hive metastore version.

## Definition of Done
- [x] Update hive-metastore-service docker image to use Hive 4.0.0
- [x] Validate that the current v2 hive-metastore-service helm chart functions as expected
  - [x] If not, make necessary updates to ensure functionality.
  - [x] Refactor chart to live under extensions-helm-spark-infrastructure
- [x] Update generated values/Chart file in downstream projects using the v2 profile with sensible defaults

## Test Strategy/Script

1. Generate a new project using the following command:
```
mvn archetype:generate -B -DarchetypeGroupId=com.boozallen.aissemble \
                          -DarchetypeArtifactId=foundation-archetype \
                          -DarchetypeVersion=1.8.0-SNAPSHOT \
                          -DartifactId=test-project\
                          -DgroupId=org.test \
                          -DprojectName='Test' \
                          -DprojectGitUrl=test.org/test-project\
&& cd test-project
```
2. Add the following pipeline to `test-project-pipeline-models/src/main/resources/pipelines/`
```
{
  "name": "PysparkPersist",
  "package": "com.boozallen",
  "type": {
    "name": "data-flow",
    "implementation": "data-delivery-pyspark"
  },
  "steps": [
    {
      "name": "PersistData",
      "type": "synchronous",
      "persist": {
        "type": "hive"
      }
    }
  ]
}
```
3. Add the following record to `test-project-pipeline-models/src/main/resources/records/`
```
{
  "name": "CustomRecord",
  "package": "com.boozallen.aiops.mda.pattern.record",
  "description": "Example custom record for Pyspark Data Delivery Patterns",
  "fields": [
    {
      "name": "customField",
      "type": {
        "name": "customType",
        "package": "com.boozallen.aiops.mda.pattern.dictionary"
      }
    }
  ]
}
```
4. Add the following dictionary to `test-project-pipeline-models/src/main/resources/dictionaries/`
```
{
  "name": "PysparkDataDeliveryDictionary",
  "package": "com.boozallen.aiops.mda.pattern.dictionary",
  "dictionaryTypes": [
    {
[...]

Issue #116, opened 2024-05-30
Feature: As a data engineer, I want Spark backed by the latest version of Hive so I have the latest fixes and features


ewilkins-csi:
## Description
Upgrade to a newer version of Hive. We are currently on Hive 3.0 and the latest version is 4.0.0, which is a good starting target.  If there are too many breaking changes, or we decide we want to give 4.0 time to mature, 3.1.3 is the newest 3.x version.  There are now public DockerHub images published by Apache that package hive and _would_ be a great base to start from, however they are all still running Java 8 which is EOL.  We could consider aligning our image with the paradigms of Apache's public images so that when they do update the Java version we are poised to switch more easily

## DOD
- [ ] Hive metastore service Docker image is updated to run Hive 4.0.0 (or 3.1.3)
- [ ] Hive metastore service Helm chart is updated to maintain functionality with any image changes

--- If Helm chart changes are required, the following items must also be completed. ---
- [ ] Migration instructions are written for moving from the v1 Hive service chart to the v2 chart
- [ ] The v1 Helm chart Baton migration is updated to skip updating the image version for the v1 Hive chart
- [ ] The release notes are updated indicating the v2 Hive chart is available and v1 is deprecated and no longer compatible with the new docker image

## Test Strategy/Script
# Create a project with Data Access and a pipeline that reads and writes data to both Hive and Delta Lake
# Run the pipeline to create and read test data
# Query Data Access to ensure the test data can be retrieved
# Create a v1.6 project with a simple pipeline
# Follow the v1 -> v2 migration instructions for Hive Metastore service
# Ensure the project can deploy and the pipeline can run

PR #130, opened 2024-06-07
#127 #116 Hive Metastore Service v2 chart and Hive upgrade

peter-mcclonski:


extensions/extensions-docker/aissemble-hive/aissemble-hive-service/pom.xml
```diff
@@ -15,4 +15,57 @@
 
     <packaging>orphedomos</packaging>
 
+    <properties>
+        <dockerbuild.jars.directory>target/dockerbuild/jars/</dockerbuild.jars.directory>
+        <version.hive.metastore>4.0.0</version.hive.metastore>
+        <version.hadoop>3.4.0</version.hadoop>
```
peter-mcclonski, line 21: I: Taking the opportunity to bump the Hadoop version for this service.  Out of sync with Spark (which uses 3.3.4) as that version is no longer available on https://dlcdn.apache.org/hadoop/common/.

extensions/extensions-docker/aissemble-hive/aissemble-hive-service/pom.xml
```diff
@@ -15,4 +15,57 @@
 
     <packaging>orphedomos</packaging>
 
+    <properties>
+        <dockerbuild.jars.directory>target/dockerbuild/jars/</dockerbuild.jars.directory>
+        <version.hive.metastore>4.0.0</version.hive.metastore>
```
peter-mcclonski, line 20: I: Managing dependencies through Maven as a matter of good practice and centralization.

extensions/extensions-helm/extensions-helm-hive/pom.xml
```diff
@@ -15,7 +15,6 @@
 
 	<modules>
         <module>aissemble-hive-metastore-db-chart</module>
-        <module>aissemble-hive-metastore-service-chart</module>
```
peter-mcclonski, line 18: I: Refactored the metastore service to the `spark-infrastructure` aggregator module.  Next few files reflect this change.

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/README.md
```diff
@@ -0,0 +1,30 @@
...
+| service.spec.ports                              | The service spec ports                                                                                                                                  | No                | &emsp; - name: "thrift" <br/>&emsp;&emsp;port: 9083 <br/>&emsp;&emsp;targetPort: 9083                                                                                                                                                                                       |
+| configMap.metastoreServiceConfig.baseProperties | Default configuration for the metastore service                                                                                                         | No                | See values.yaml                                                                                                                                                                                                                                                             |
```
peter-mcclonski, line 28: I: The original version of this README enumerated every property that was set in the default.  This resulted in an illegible line in the README source.  Opted to replace it with a referral to the source material, but open to discussion as to whether this is appropriate.

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/templates/deployment.yaml
```diff
@@ -33,17 +33,28 @@ spec:
       nodeSelector:
         {{- toYaml . | nindent 12}}
       {{- end }}
-      {{- with .Values.deployment.initContainers }}
       initContainers:
-        {{- toYaml . | nindent 12 }}
-      {{- end }}
+        - name: "init-or-upgrade-schema"
```
peter-mcclonski, line 37: I: This initcontainer will check the version of the schema in the underlying database, and either create it or upgrade it, as needed.  This is future-proof for upgrades down the line.  Notably, it does require specification of the db driver.  aiSSEMBLE currently uses mysql and only mysql for this purpose, but the driver was still extracted to an intentionally undocumented property in the values file in case this ever changes.

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/tests/configmap_test.yaml
```diff
@@ -33,26 +33,12 @@ tests:
                   <value>jdbc:mysql://hive-metastore-db:3306/metastore?createDatabaseIfNotExist=true&amp;allowPublicKeyRetrieval=true&amp;useSSL=false</value>
                   <description>JDBC connect string for a JDBC metastore</description>
                 </property>
-                <property>
```
peter-mcclonski, line 36: I: Moved these properties to the downstream values.yaml

foundation/foundation-mda/src/main/resources/templates/deployment/spark-infrastructure/v2/spark.infrastructure.values.yaml.vm
```diff
@@ -12,11 +12,11 @@ aissemble-thrift-server-chart:
   deployment:
     envFromSecret:
       AWS_ACCESS_KEY_ID:
-        secretName: aws-s3-credentials
-        key: aws_access_key_id
+        secretName: remote-auth-config
```
peter-mcclonski, line 15: I: It was brought to my attention that we publish this secret as part of the localstack chart, so we use it for out of the box functionality and as a model of good practice.

foundation/foundation-upgrade/src/main/resources/migrations.json
```diff
@@ -160,7 +160,6 @@
               "**/jenkins/values.yaml",
               "**/kafka-cluster/values.yaml",
               "**/hive-metastore-db/values.yaml",
-              "**/hive-metastore-service/values.yaml",
```
peter-mcclonski, line 163: I: The v1 chart is deprecated after 1.7.0 due to significant breaking changes.

foundation/foundation-mda/src/main/resources/templates/deployment/hive-metastore-service/hive-metastore-service-values.yaml.vm
```diff
@@ -10,7 +10,7 @@ image:
   imagePullPolicy: IfNotPresent
   dockerRepo: ghcr.io/
   # Overrides the image tag whose default is the chart appVersion.
-  tag: "${versionTag}"
+  tag: "1.7.0"
```
peter-mcclonski, line 13: I: Locking the version as the v1 chart will be deprecated after 1.7.0

extensions/extensions-docker/aissemble-hive/aissemble-hive-service/pom.xml
peter-mcclonski: I would like to float the idea of using the `current` version rather than pegging a specific version, but I can see both pros and cons to this approach.  Namely, it becomes non-obvious what version we're actually on.

extensions/extensions-docker/aissemble-hive/aissemble-hive-service/src/main/resources/docker/Dockerfile
```diff
@@ -1,32 +1,29 @@
...
 
 WORKDIR /opt
 
-ENV HADOOP_VERSION=3.3.6
-ENV METASTORE_VERSION=3.0.0
-
-ENV HADOOP_HOME=/opt/hadoop-${HADOOP_VERSION}
-ENV HIVE_HOME=/opt/apache-hive-metastore-${METASTORE_VERSION}-bin
+ARG HADOOP_VERSION
+ARG METASTORE_VERSION
```
ewilkins-csi, line 11: Q: Isn't this ARG already defined above?

extensions/extensions-docker/aissemble-hive/aissemble-hive-service/src/main/resources/docker/Dockerfile
```diff
@@ -1,32 +1,29 @@
...
 LABEL org.opencontainers.image.source = "https://github.com/boozallen/aissemble"
 
 WORKDIR /opt
 
-ENV HADOOP_VERSION=3.3.6
-ENV METASTORE_VERSION=3.0.0
-
-ENV HADOOP_HOME=/opt/hadoop-${HADOOP_VERSION}
-ENV HIVE_HOME=/opt/apache-hive-metastore-${METASTORE_VERSION}-bin
+ARG HADOOP_VERSION
```
ewilkins-csi, line 10: A: Unused since we're just using the one that apache/hive was packaged with.

extensions/extensions-docker/aissemble-hive/aissemble-hive-service/src/main/resources/docker/Dockerfile
```diff
@@ -1,32 +1,29 @@
...
+    chown hive:hive -R ${HIVE_HOME} && \
+    ln -s ${JAVA_HOME} /opt/jre
+
+ADD ${JARS_DIR}/* ${HIVE_HOME}/lib/
 
 USER hive
-EXPOSE 9083
 
-ENTRYPOINT ["/opt/apache-hive-metastore-3.0.0-bin/bin/start-metastore"]
+ENTRYPOINT ["/opt/hive/bin/hive", "--skiphadoopversion", "--skiphbasecp", "--verbose", "--service", "metastore"]
```
ewilkins-csi, line 29: Q/A: Does it make sense to use the HIVE_HOME env, so that if a downstream builder installs their own hive, it "just works"?

peter-mcclonski: I want so badly to say yes, but unfortunately, when we deploy through helm, all the dockerfile ENV settings get blown away. I'm trying to let the dockerfile function in a vacuum, without duplicating environment variables across both the image and the helm chart.

extensions/extensions-docker/aissemble-hive/aissemble-hive-service/src/main/resources/docker/Dockerfile
```diff
@@ -1,32 +1,29 @@
-FROM openjdk:11-jre-slim
+ARG METASTORE_VERSION
+FROM apache/hive:${METASTORE_VERSION} as appsource
+
+FROM eclipse-temurin:17-jre-ubi9-minimal as final
```
ewilkins-csi, line 4: A: alpine is about half the size, and even the default `17-jre` (canonical) is about 20% smaller.

extensions/extensions-helm/extensions-helm-hive/pom.xml
ewilkins-csi: A: I'd rather keep hive separate, since it's a standalone data warehousing solution outside of it's use with Spark.

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/README.md
ewilkins-csi: I think this is fine, but you should be able to actually link values.yaml with the relative path.

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/values.yaml
```diff
@@ -2,8 +2,8 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 app:
-  # Sets labels for app.kubernetes.io/name Default is Chart.name (aissemble-hive-metastore-service-chart)
-  name: ''
+  # Sets labels for app.kubernetes.io/name. Default is Chart.name (aissemble-hive-metastore-service-chart)
+  name: 'hive-metastore-service'
```
ewilkins-csi, line 6: A: I think the previous behavior makes more sense in the context of a downstream project, where the chart will naturally inherit the `appName` set in the deploy POM.

extensions/extensions-docker/aissemble-hive/aissemble-hive-service/src/main/resources/docker/Dockerfile
peter-mcclonski: This is initially redeclared due to this: https://docs.docker.com/reference/dockerfile/#understand-how-arg-and-from-interact

With that said, upon further examination, the latter ARG is unused so it can be removed.

foundation/foundation-mda/src/main/resources/templates/deployment/hive-metastore-service/hive-metastore-service-values.yaml.vm
ewilkins-csi: We should go ahead and deprecate the profile for the v1 chart in this commit as well.

ewilkins-csi (approved):
S: Add hive-service v2 profile to containers.adoc and instruct users on migrating.

extensions/extensions-docker/aissemble-hive/aissemble-hive-service/src/main/resources/docker/Dockerfile
ewilkins-csi: Interesting to note nonetheless. I wasn't aware args worked that way.

peter-mcclonski:
> S: Add hive-service v2 profile to containers.adoc and instruct users on migrating.

I was under the impression that we would be waiting until the new spark-infrastructure chart was completed before making that change-- particularly since there are still so many manual steps required to keep the project functional when migrating.

ewilkins-csi:
> I was under the impression that we would be waiting until the new spark-infrastructure chart was completed before making that change-- particularly since there are still so many manual steps required to keep the project functional when migrating.

Ok. That works for me. As long as we don't lose track of it.

extensions/extensions-helm/extensions-helm-hive/pom.xml
peter-mcclonski: I can go either way on it.  I think it's an interesting case because while it absolutely can be used outside of spark, we're centralizing its usage to the spark infrastructure.  I agree that it's not immediately an intuitive sibling to SHS and STS, but given that the method of consumption through aissemble is exclusively in the context of spark (and, pointedly, through the newly consolidated spark-infrastructure chart) I feel a bit weird about treating the hive ecosystem (service + db) as a wholly separate entity in the baseline.  

As an idea, how would you feel about a directory structure where `extensions-spark-infrastructure` is parent to `extensions-hive` and `extensions-spark-services`-- The former housing `aissemble-hive-metastore-chart` and `aissemble-hive-mysql-db-chart`, while the latter houses the `STS` and `SHS` charts?

foundation/foundation-mda/src/main/resources/templates/deployment/hive-metastore-service/hive-metastore-service-values.yaml.vm
peter-mcclonski: I had intended to deprecate it alongside the v1 SHS and STS charts after completing the full chart (along with writing migration steps, etc).  I'm wary of hitting the deprecation button before the full chart is ready.  I imagine a case where for whatever reason we have to cut a 1.7.1 release and the deprecation is in before the replacement is ready.

extensions/extensions-docker/aissemble-hive/aissemble-hive-service/src/main/resources/docker/Dockerfile
peter-mcclonski: Alpine isn't published for `arm64`.  Good call on the `17-jre` option, though-- I accidentally looked at the size of the Windows images initially and thought they were larger.

peter-mcclonski: Another advantage of the `17-jre` image-- currently there are fewer vulnerabilities reported.

foundation/foundation-mda/src/main/resources/templates/deployment/hive-metastore-service/hive-metastore-service-values.yaml.vm
ewilkins-csi: We'd cut 1.7.1 from 1.7.0 not dev. It's a fair point either way though. I guess I was just thinking if we're refusing to update the image it'd be nice to tell the user why.  But also understand from a user perspective it might be frustrating to see a deprecation message and have no way to resolve it. Could we word it something like "due to ongoing effort to create a Hive v2 chart and subsequent incompatibilities, v1 cannot be upgraded beyond version 1.7" as a middle ground?

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/values.yaml
ewilkins-csi: Oh right. Hmmmm. Ok.

foundation/foundation-mda/src/main/resources/templates/deployment/hive-metastore-service/hive-metastore-service-values.yaml.vm
peter-mcclonski: Absolutely-- I can do that.

extensions/extensions-helm/extensions-helm-hive/pom.xml
ewilkins-csi: I guess I'm worried it will encourage unnecessary coupling to Spark.  At the end of the day, where it's housed is transparent to downstream projects, so it's not a huge deal either way.  It's just about what we think will create the lowest friction for baseline developers.  If we regret the decision, it's easy enough to move again.

peter-mcclonski: I can see it either way, but agree it's low lift to revert if we don't like it.  I'll keep it refactored for now, and will plan to refactor metastore-db in the next ticket.  Like you said, if we hate it, we can move it again.

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/values.yaml
peter-mcclonski: I do want to raise one interesting thing that this hints at: The hive charts and STS/SHS charts _feel_ different to interact with from a downstream user perspective.  This is fundamentally because they were built by different people at different times.  If we want to do a more wholesale unification of the styles and approaches to each, _we can do that_.  I wouldn't necessarily do it right now, but I think it's one of those oddities where we'll probably come back to it in the future.  Even basic things like how `metastore-config.xml` is built versus how `hive-site.xml` is loaded into STS.  One takes a raw XML string, while the other takes a list of objects that then gets transformed into an XML string.  This naming difference is sort-of one of those oddities-- Though the main motivator for keeping the name _specifically_ as-is was to minimize the breaking changes from previous versions.

Issue #128, opened 2024-06-06
Feature: Remove all occurrences of the mvn command and replace with the maven wrapper class. 


J-Clingerman:
## Description
As part of the efforts to fix up the config-tool, we need to remove all occurrences of the mvn command. Instead of using mvn, we will be using the maven wrapper class (./mvnw).

## DOD
Acceptance criteria required to realize the requested feature
- [ ] Using checkboxes helps to communicate progress
- [ ] Verify for Windows:
- Remove the install of maven in the script, and prompt to use the maven wrapper.

- [ ] Verify for Mac:

- Remove the install of maven in the script, and prompt to use the maven wrapper.

- [ ] Update scenarios in:

- aissemble-environment-config/config-tool/specifications/mac-installation.feature
- aissemble-environment-config/config-tool/specifications/windows-installation.feature

- [ ] Update all documentation where the 'mvn' command is referenced and replace with './mvnw'.

## Test Strategy/Script
SCENARIO 1 (Ensure all instances of the mvn command are converted to ./mvnw)

- In your IDE, do a search all for mvn keyword
- Confirm that there are no 'standalone' mvn commands
- - -i.e ./mvnw NOT mvn

 

SCENARIO 2 (Does the config-tool successfully run and not install MAVEN)

- Clone the [environment config repo](https://github.boozallencsn.com/aissemble/aissemble-environment-config) .

`git clone https://github.boozallencsn.com/aissemble/aissemble-environment-config.git
`
- Run the config-tool from aissemble root directory

`bash <Path_TO_CONFIG_TOOL>./aissemble-environment-config/config-tool/src/configure-env.sh
` 
- When you run the config-tool, you will notice there is no 'Installing maven' on the logs.

- Verify all your versions are correct to run aiSSEMBLE.
- [Build/Local configurations requirements.](https://boozallen.github.io/aissemble/aissemble/current/configurations.html)

- Run mvn wrapper version in the aiSSEMBLE directory.

 

`./mvnw --version
`

- Java Version should be 11
- Mvn version should be 3.9.6

SCENARIO 3 (Does aiSSEMBLE build successfully with mvn wrapper)

Confirm you can still successfully build aiSSEMBLE with mvn wrapper
 

`./mvnw clean install
`
## References/Additional Context
N/A

PR #135, opened 2024-06-11
 #128 Removed maven wrapper from the archetype.adoc and foundation-ar…

J-Clingerman:


d-ryan-ashcraft (commented):
S (Substantive): do we also need to generate in the maven wrapper artifacts so the they will be able to use it in subsequent builds?  Was expecting that in here too.

d-ryan-ashcraft:
> > S (Substantive): do we also need to generate in the maven wrapper artifacts so the they will be able to use it in subsequent builds? Was expecting that in here too.
> 
> My concern is that I do not want to hold-up the release Emily is working on. Since we didnt account for this in the ticket, we would have to start the process over and re-visit the DoD.

Also don't think we can leave it in a middle state.  I can help out with this.

foundation/foundation-archetype/test-project-archetype.sh
```diff
@@ -175,7 +175,7 @@ cp ../../../src/test/resources/test-project-archetype-script/maven-build-cache-c
 echo -e "\n# maven-suppress-warnings" >> Tiltfile
 
 echo -e "\nINFO: Running full build to generate project structure\n"
```
d-ryan-ashcraft, line 177: The instances below here need to be `./mvnw` since that is how the the instructions now work.

foundation/foundation-archetype/src/main/resources/archetype-resources/mvnw
d-ryan-ashcraft, line 1: I (Informational): This is just the default Maven wrapper shell script - no need to review, it's generated

foundation/foundation-archetype/src/main/resources/archetype-resources/mvnw.cmd
d-ryan-ashcraft, line 1: I: This is just the default Maven wrapper cmd script - no need to review, it's generated

foundation/foundation-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
```diff
@@ -66,11 +66,29 @@
...
+			<excludes>
+				<exclude>mvnw</exclude>
+				<exclude>mvnw.cmd</exclude>
+			</excludes>
 		</fileSet>
 		<fileSet>
 			<directory>.mvn</directory>
 			<includes/>
 		</fileSet>
+		<fileSet filtered="false" encoding="UTF-8">
```
d-ryan-ashcraft, line 78: I (Informational): filtering breaks the variables in the mvnw scripts, so turn it off when copying those files

ewilkins-csi (approved):

ewilkins-csi:
The second commit is ref'ing the wrong issue number.

PR #144, opened 2024-06-12
#123 add test.pypi.org as supplemental source

ewilkins-csi:


habibimoiz (approved):

Cho-William (approved):

foundation/foundation-mda/src/main/resources/templates/data-delivery-data-records/combined.pyproject.toml.vm
```diff
@@ -23,6 +23,11 @@ black = ">=22.1.0"
 behave = ">=1.2.6"
 nose = ">=1.3.7"
 
+[[tool.poetry.source]]
```
peter-mcclonski, line 26: Q: This makes total sense for devs working off snapshot versions, but is there any expectation that it can cause problems for users on release versions?  For instance, once 1.8.0 is released and 1.9.0-SNAPSHOT is pushed to test, can we say confidently that the 1.8.0 release will be selected over the 1.9.0-SNAPSHOT version?  I understand that the priority = supplemental is supposed to handle this, but just making sure.

Cho-William:
Noting that this fixes an issue with the CI's archetype tests, which were unable to pull 1.8.0.dev python artifacts, since snapshot artifacts are published to test.pypi.org, unlike release artifacts

foundation/foundation-mda/src/main/resources/templates/data-delivery-data-records/combined.pyproject.toml.vm
Cho-William: But wouldn't the `pyproject.toml` have  `1.8.0` listed for the dependency? Poetry would never grab the `1.9.0.dev` artifacts to begin with

peter-mcclonski: Ah, good point.  I'm still somewhat wary about putting this into projects on release versions (where it's unnecessary), but otherwise, LGTM

Cho-William: It's also my understanding that the `supplemental` priority only comes into play so that Poetry won't needlessly search test.pypi.org if it has already found artifacts from pypi for say, a non-aiSSEMBLE artifact like `jsonpickle`

Cho-William: I chatted with Emily, and adding the supplemental source (even if it's not necessary for a release version project) is actually mirroring what we do with Maven. I can see this actually being helpful if a user is on a release version, but wants to grab the latest snapshot version for a certain dependency... granted the recommendation would be to instead upgrade the entire project

Cho-William: Scratch that, looks like Emily pushed a new commit to only add the source if it is a snapshot project

peter-mcclonski (approved):

Issue #92, opened 2024-05-22
Fix generated RDBMS persistence support in PySpark data delivery pipelines


ekoniec1:
Given the following pipeline definition:
```
{
  "name":"IngestAndSaveToRdbmsPipeline",
  "package":"com.github.boozallen.aissemble",
  "type":{
    "name":"data-flow",
    "implementation":"data-delivery-pyspark"
  },
  "steps":[
    {
      "name":"IngestData",
      "type":"synchronous",
      "persist": {
        "type": "rdbms"
      }
    }
  ]
}
```

The following snippet is generated into the non-modifiable base pipeline step module `ingest_data_base.py`:
```
def save_dataset(self, db_url: str, dataset: DataFrame, table_name: str) -> None:
      ....
      rdbms_session = self.rdbms_connection_pool.get_connection(db_url)
      engine = rdbms_session.get_bind()
      dataset.createOrReplaceTempView(table_name)
      rdbms_session.close()
      ...
```
Using [DataFrame.createOrReplaceTempView()](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.createOrReplaceTempView.html) creates a memory-only, non-persistent view of the given `DataFrame`.  Instead, [DataFrame.write()](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.write.html) should be leveraged, taking into account the [associated guidance for leveraging Spark with JDBC datasources](https://spark.apache.org/docs/latest/sql-data-sources-jdbc.html).  

As an aside, connection pooling capabilities are exposed (but not used) via the injected `RDBMSConnectionPool` class, which in turn delegates to [SQLAlchemy's connection pooling](https://docs.sqlalchemy.org/en/20/core/pooling.html).  However, Spark doesn't provide explicit support for connection pooling, especially given its distributed execution model - developers *may* be able to extend [JdbcConnectionProvider](https://spark.apache.org/docs/latest/api/java/org/apache/spark/sql/jdbc/JdbcConnectionProvider.html) to provide connection pooling, but this may instead result in rapidly overtaxing the database.  Consider the guidance, configurations, and tuning described in the [following](https://www.oudeis.co/blog/2020/spark-jdbc-throttling-writes/) to help generate a reasonable set of defaults and expose extension points for easy customization within pipeline step implementations.
 
[...]

PR #146, opened 2024-06-13
Issue #92: Fix PySpark RDBMS data ingest steps writing to a non-persistent table

cpointe-ibllanos:
Related Issue: #92


foundation/foundation-mda/src/main/resources/templates/data-delivery-pyspark/persist.py.vm
```diff
@@ -19,20 +19,33 @@
...
     def save_dataset(self, dataset: ${step.persist.shortCollectionType}[${step.persist.shortRecordType}], ${saveParam}) -> None:
    #elseif ($step.isPersistTypeRdbms())
     #set ($db_url = "db_url: str")
     def save_dataset(self, ${db_url}, dataset: ${step.persist.shortRecordType}, ${saveParam}) -> None:
       ${step.capitalizedName}Base.logger.info('Saving %s To RDBMS...' % self.descriptive_label)
-      rdbms_session = self.rdbms_connection_pool.get_connection(db_url)
-      engine = rdbms_session.get_bind()
-      dataset.createOrReplaceTempView(table_name)
-      rdbms_session.close()
+      rdbms_session = self.get_connection(db_url)
```
ekoniec1, line 39: S: There shouldn't be any need to get/create a new Spark session and then stop it - I think this also may result in some adverse side effects?  Using `dataset.write.jdbc(..)` is sufficient for writing a `DataFrame` to the configured JDBC datastore.

foundation/foundation-mda/src/main/resources/templates/data-delivery-pyspark/persist.py.vm
```diff
@@ -19,20 +19,33 @@
...
+      jdbc_properties = {
```
ekoniec1, line 40: S: It would be ideal if JDBC properties and other [JDBC data source options](https://spark.apache.org/docs/latest/sql-data-sources-jdbc.html#data-source-option) were exposed in such a way that they could be easily extended and/or augmented within pipeline implementations.  Even just breaking out connection option creation into a separate method would work:

```
def _get_jdbc_data_source_connection_options(self) -> dict:
  return {
            "url": ...,
            ...
  }
```

cpointe-ibllanos: After pruning the use of SQLAlchemy connection pooling, the jbasic dbc connection properties can now just be directly configured through krausening properties (spark-rdbms.properties) for any relational database. Additionally, the class `get_data_source_connection` can be overwritten in the impl to provide additional configs unique to specific kinds of databases.

foundation/foundation-mda/src/main/resources/templates/data-delivery-pyspark/asynchronous.processor.impl.py.vm
```diff
@@ -42,6 +42,16 @@ class ${step.capitalizedName}(${step.capitalizedName}Base):
...
+    #if ($step.isPersistTypeRdbms())
+    def get_data_source_configs(self) -> dict:
+        """
+        Retrieves the base configurations for the JDBC Connection
+        :return: JDBC Configs as a dictionary
+        """
+        configs = super().get_data_source_configs()
+        # TODO: Apply additional config options as needed by your specific database!
+        return configs
+    #end
```
cpointe-ibllanos, line 54: I: Overwritten class to give the user the option to extend database config properties if needed. This is primarily for databases that require more than the default jdbc connection requirements.

foundation/foundation-mda/src/main/resources/templates/data-delivery-pyspark/persist.py.vm
```diff
@@ -8,32 +8,26 @@
...
-      For example, depending on the databse you choose:
-          PostgreSQL: 'postgresql://username:password@host:port/database'
-          MySQL: 'mysql://username:password@host:port/database'
-          SQLite: 'sqlite:///db.sqlite'
-      :return: A connection object from the RDBMS connection pool.
-      """
-      connection = self.rdbms_connection_pool.get_connection(db_url)
-      return connection
-    #end
-
```
cpointe-ibllanos, line 25: I: Pruning the use of SQLAlchemy - cannot be used with jdbc protocol to drive spark changes.

foundation/foundation-mda/src/main/resources/templates/deployment/spark-operator/spark-application-base-values.yaml.vm
```diff
@@ -56,9 +58,11 @@ sparkApp:
         #if (${enablePySparkSupport})
         - org.apache.sedona:sedona-python-adapter-3.0_2.12:${versionSedona}
         - org.datasyslab:geotools-wrapper:${versionGeotools}
-        - org.postgresql:postgresql:${versionPostgresql}
         #end
         #end
+        #if (${enablePySparkSupport})
+        - org.postgresql:postgresql:${versionPostgresql}
+        #end
```
cpointe-ibllanos, line 65: I: Provide the postgres jdbc driver by default to match default connection properties, even if Sedona isn't present. Regardless of this change, users will have to alter this driver dependency if they choose not to use postgres.

foundation/foundation-mda/src/main/resources/templates/data-delivery-pyspark/abstract.data.action.py.vm
```diff
@@ -51,8 +51,13 @@ class AbstractDataAction(ABC):
         builder = SparkSession.builder
         if properties.getProperty("execution.mode.legacy", "false") == "true":
             builder = builder\
-                .master("local[*]")\
+                .master("local[*]") \
+                #if ($step.isPersistTypeRdbms())
+                .config("spark.driver.host", "localhost")\
+                .config("spark.jars.packages", "org.postgresql:postgresql:{$versionPostgresql}")
```
cpointe-ibllanos, line 57: I: This actually is not needed. Removing.

foundation/foundation-mda/src/main/resources/templates/data-delivery-pyspark/abstract.data.action.py.vm
```diff
@@ -51,7 +51,7 @@ class AbstractDataAction(ABC):
         builder = SparkSession.builder
         if properties.getProperty("execution.mode.legacy", "false") == "true":
             builder = builder\
-                .master("local[*]")\
+                .master("local[*]") \
```
peter-mcclonski, line 54: A: Unrelated to this ticket but we really need to just remove this concept of legacy execution mode.

foundation/foundation-mda/src/main/resources/templates/data-delivery-pyspark/asynchronous.processor.base.py.vm
```diff
@@ -12,10 +12,11 @@ from aiops_core_filestore.file_store_factory import FileStoreFactory
 #if (!$step.provenance || $step.provenance.enabled)
 from aiops_core_metadata.hive_metadata_api_service import HiveMetadataAPIService
 #end
-#if ($step.isPersistTypePostgres() || $step.isPersistTypeRdbms())
+#if ($step.isPersistTypePostgres())
 from aiops_core_config import SparkPostgresConfig
```
peter-mcclonski, line 16: Q/S: Why are we still persisting the notion of Postgres as a distinct option separate from general RDBMS offerings?  My impression is that the SparkPostgresConfig has negligible functionality associated with it, and the RDBMS config should be able to fully replace it.

foundation/foundation-mda/src/test/java/com/boozallen/aiops/mda/metamodel/element/PipelineSteps.java
```diff
@@ -765,6 +797,12 @@ public void the_pipeline_is_created_with_the_specified_persist_save_mode() {
         assertEquals("Unexpected persist mode found", "test-mode", persist.getMode());
     }
 
+    @Then("the pipeline is created with the specified persist type {string}")
+    public void the_pipeline_is_created_with_the_specified_persist_save_type(String expectedType) {
+        String realType = pipeline.getSteps().get(0).getPersist().getType();
```
cwoods-cpointe, line 802: q: any worries of a null pointer getting the persist type?

foundation/foundation-mda/src/main/resources/templates/data-delivery-pyspark/synchronous.processor.base.py.vm
```diff
@@ -12,10 +12,11 @@ from aiops_core_filestore.file_store_factory import FileStoreFactory
 #if (!$step.provenance || $step.provenance.enabled)
 from aiops_core_metadata.hive_metadata_api_service import HiveMetadataAPIService
 #end
-#if ($step.isPersistTypePostgres() || $step.isPersistTypeRdbms())
+#if ($step.isPersistTypePostgres())
```
peter-mcclonski, line 15: Same question here

foundation/foundation-mda/src/main/resources/templates/data-delivery-pyspark/synchronous.processor.impl.py.vm
```diff
@@ -42,6 +42,17 @@ class ${step.capitalizedName}(${step.capitalizedName}Base):
         #end
         #end
 
+    #if ($step.isPersistTypeRdbms())
```
peter-mcclonski, line 45: S: Rather than copying this identical snippet between both processor impl templates, can we pull it out to its own file and include by reference with a #parse call?

foundation/foundation-mda/src/main/resources/templates/deployment/spark-operator/spark-application-base-values.yaml.vm
```diff
@@ -31,9 +31,11 @@ sparkApp:
         #if (${enablePySparkSupport})
         - https://repo1.maven.org/maven2/org/apache/sedona/sedona-python-adapter-3.0_2.12/${versionSedona}/sedona-python-adapter-3.0_2.12-${versionSedona}.jar
         - https://repo1.maven.org/maven2/org/datasyslab/geotools-wrapper/${versionGeotools}/geotools-wrapper-${versionGeotools}.jar
-        - https://repo1.maven.org/maven2/org/postgresql/postgresql/${versionPostgresql}/postgresql-${versionPostgresql}.jar
         #end
         #end
+        #if (${enablePySparkSupport})
```
peter-mcclonski, line 36: S: What do we gain by pulling this out to here?

foundation/foundation-mda/src/main/resources/templates/deployment/spark-operator/spark-application-base-values.yaml.vm
peter-mcclonski: Shouldn't we make inclusion conditional on whether Postgres is the persistence of choice?

foundation/foundation-mda/src/test/resources/specifications/pipeline-generation.feature
```diff
@@ -47,3 +47,10 @@ Feature: Generating pipeline steps
 		When the profile "data-delivery-pyspark-pipeline" is generated
 		Then a common step parent is created
 		And the common step parent is customizable
+
+	Scenario: PySpark Data Ingest steps in data-flow pipelines can write to a RDMS database
```
cwoods-cpointe, line 51: q: Noticed the file is marked with manual but the steps are pretty flushed out. Are this new test and steps meant to be ran?

foundation/foundation-mda/src/test/java/com/boozallen/aiops/mda/metamodel/element/PipelineSteps.java
```diff
@@ -594,6 +625,7 @@ public void aPipelineIsConfiguredForMessagingWithoutAChannelName() throws Except
     }
 
     @When("machine learning pipelines are validated")
+    @When("the pipelines are validated")
```
peter-mcclonski, line 628: Q: Should this method have both @When annotations on it?

foundation/foundation-mda/src/test/resources/specifications/pipeline-generation.feature
jacksondelametter, line 51: Should the test be for postgressql? From the code, it looks like there are distinctive differences between RDMS and Postgressql.

cpointe-ibllanos: Good catch - I originally had the tests under the data module feature file and didn't catch the manual tag on the switch. I moved it back and tests have been refactored.

foundation/foundation-mda/src/test/java/com/boozallen/aiops/mda/metamodel/element/PipelineSteps.java
cpointe-ibllanos: That is a risk should a future change in the tests cause the pipeline or its steps to get generated improperly. I can add a not-null assertion with an error message that provides the context for the failure should it happen.

foundation/foundation-mda/src/main/resources/templates/data-delivery-pyspark/asynchronous.processor.base.py.vm
cpointe-ibllanos: At the time of implementation I was being cautious not to pull the postgres configs that could still be being used in downstream projects under the impression it'd be best to save that for a baton migration follow-up. 

I think you're right though - since most of the changes ended up being to base files and it's unlikely the jdbc configs class will be referenced outside of them, there's probably no need for a migration and it'd be best to replace the postgres configs entirely.

foundation/foundation-mda/src/test/resources/specifications/pipeline-generation.feature
cpointe-ibllanos: As per the Peter's comment above, RDBMS as a type is intended to replace the postgres type as postgres is a RDBMS option. I was being cautious and holding out on completely pruning it in case they needed a migration in downstreams, but there are no breaking changes to any impl files so it'd be best to pull out the postgres type.

foundation/foundation-mda/src/test/java/com/boozallen/aiops/mda/metamodel/element/PipelineSteps.java
cpointe-ibllanos: Initially I was reusing the test step logic with different wording, but I ended up moving it to its own method.

foundation/foundation-mda/src/main/resources/templates/data-delivery-pyspark/asynchronous.processor.base.py.vm
peter-mcclonski: I would note that the `Postgres` persist type has been deprecated for a significant amount of time, and has not been present in our official documentation for the same amount of time.  In this case, I think it's safe to remove, especially since if it's entirely in base/generated code.  Strictly speaking, we don't have to do it _right now_, but the amount of changes in this PR that go out of their way to maintain explicit Postgres support as a distinct entity does concern me.

foundation/foundation-mda/src/main/resources/templates/deployment/spark-operator/spark-application-base-values.yaml.vm
cpointe-ibllanos: Yes. That said, since the switch to defining the jdbc connection details in a krausening file, I wonder if it would be best not to pre-emptively import drivers or have default rdbms configs that assume postgresql as the default. 

We do expect the users to be able to use the aissemble framework to implement their relational db of choice - dependency management included, correct? I'm trying to make the call on what is too little or too much scaffolding as providing the correct drivers dynamically for every possible rdbms option seems like its own effort. What do you think @peter-mcclonski ?

peter-mcclonski: Right now, the only RDBMS options present in aissemble are Postgres and, technically, MySQL-- though MySQL is used for only a single purpose (backing the Hive Metastore).  I would say that it's reasonable to support Postgres as the default implementation of the RDBMS option, but that it should be done in a way such that it's possible for a project to change from that default.

foundation/foundation-mda/src/main/resources/templates/deployment/spark-operator/spark-application-base-values.yaml.vm
cpointe-ibllanos: I was just trying to be consistent with the usePackageDeps conditional based on the changes below. That said, depending on the outcome of the other comment in this file below, it may not be warranted at all.

peter-mcclonski: It looks to me like we're saying that _if PySpark Support is enabled_, we always want to download the Postgres jar directly, whether we're using package deps or not.  It looks to me like for pyspark pipelines, we would have both the Jar and Package dependency simultaneously.

foundation/foundation-mda/src/main/resources/templates/deployment/spark-operator/spark-application-base-values.yaml.vm
peter-mcclonski: It still strikes me as bizarre that the condition for Postgres inclusion is based on whether we're running a pyspark pipeline, rather than whether we're running a pipeline with RDBMS storage

foundation/foundation-mda/src/main/resources/templates/data-delivery-pyspark/asynchronous.processor.base.py.vm
cpointe-ibllanos: @peter-mcclonski I was able to pull out all the postgres persist type code from the PySpark implementation. I noticed the Spark/Java side also makes an effort to preserve the postgres option, but I take it that's out of scope for this specific issue.

foundation/foundation-mda/src/main/resources/templates/deployment/spark-operator/spark-application-base-values.yaml.vm
cpointe-ibllanos: It seems like it's configured to pull one or the other.

```
      #if (!${usePackageDeps})
      jars:
      ...
      #else
      packages:
      ...
      #end

```

peter-mcclonski: Ah, I miscounted my #end calls.  Like I said separately, though, this still feels aggressively like the wrong condition.

foundation/foundation-mda/src/main/resources/templates/deployment/spark-operator/spark-application-base-values.yaml.vm
cpointe-ibllanos: Agreed - I added the RDBMS flag to the generator and changed it in the template.

foundation/foundation-mda/src/main/resources/templates/deployment/spark-operator/spark-application-base-values.yaml.vm
cpointe-ibllanos: Adressed, same as comment below

aaron-gary (approved):

Issue #118, opened 2024-05-30
Feature: As an aiSSEMBLE user, I want to specify rules for my property values, so that configuration can be refreshed when needed


carter-cundiff:
## Description
Follow on to implement #76. As part of Universal Configurations, we want to be able to specify rules for configuration properties within the vault server (like expiration date or use limits). This will allow the user to realize when property values are no longer valid and need to be updated.

## DOD
- [x] Create `propertyKey` class and refactor existing implemention to utilize it where applicable
- [x] Policies can be defined for properties outlining custom limitations (use limit, expiration date)
  - [x] Will be defined as JSON policies and include rules, regeneration method, and target properties
  - [x] Can define environment policies to override base policies
  - [x] Modify `Policy` class to accept multiple targets and deprecate previous single target methods
  - [x] Create `PropertyRegenerationPolicy` class for storing the policy in Java representation
  - [x] Create `PropertyRegenerationPolicyManager` for loading JSON policies into the `PropertyRegenerationPolicy` class
- [x] Policies will be read in by `ConfigLoader` where they will be reconciled and validated
  - [x] At most one policy per property
- [x] Update documentation to include information on utilizing policies

-------------Stretch Goal--------------------------

- [ ] Create `PropertyRegenerationRule` abstract class and basic rule classes that implement it:
  - [ ] Class to set a max number of uses before a property is regenerated
  - [ ] Class to set a expiration date before a property is regenerated
- [ ] Each rule within a policy will be initialized for each target
  - [ ] Metadata produced by each rule will be saved within the Vault property custom metadata
  - [ ] Create API call for [adding custom metadata](https://developer.hashicorp.com/vault/api-docs/secret/kv/kv-v2#create-update-metadata)
- [ ] Switch to updated fork of [Vault Java Driver](https://github.com/jopenlibs/vault-java-driver)
  - [ ] Open PR to merge metadata call into the fork

#### ConfigLoader:
Adding to `foundation/foundation-configuration-store/src/test/resources/specifications/load-configurations.feature`
```
  Scenario: The ConfigLoader loads base and environment policies successfully

  Scenario: The ConfigLoader creates rule metadata for each rule in each policy

  Scenario: The policy is misformatted

  Scenario: The policies are not distinguishable

  Scenario: A policy has a undefined target property

  Scenario: A property has multiple policies

  Scenario: A policy has a rule implementation class could not be found
```
#### RuleMetadataDAO:
[...]

PR #151, opened 2024-06-14
#118 Read in and validate policies for universal config properties

carter-cundiff:
- Updated config store to make use of `PropertyKey` object instead of `group name` and `property name` strings
- Updated java and python implementations of `Policy` to take a list of `targets` rather than a single `target` while maintaining backwards compatability


extensions/extensions-configuration-store-vault/src/test/java/com/boozallen/aissemble/configuration/dao/VaultPropertyDaoSteps.java
```diff
@@ -83,27 +82,10 @@ public void theConfigurationsAreLoaded() {
     @And("the properties can be read from the vault server")
     public void theConfigLoaderUsesVaultPropertyDaoToWriteToVault() {
         for (Property property : properties) {
-            assertEquals(propertyDao.read(property.getGroupName(), property.getName()), property);
+            assertEquals(propertyDao.read(property.getPropertyKey()), property);
         }
     }
 
-    public Set<Property> getProperties() {
```
carter-cundiff, line 90: I: Removed unused function

foundation/foundation-configuration-store/pom.xml
```diff
@@ -60,8 +60,8 @@
                     </systemPropertyVariables>
                     <!-- set environment variables for unit test -->
                     <environmentVariables>
-                        <BASE_URI>/tmp/configurations/base</BASE_URI>
-                        <ENVIRONMENT_URI>/tmp/configurations/example-env</ENVIRONMENT_URI>
+                        <BASE_PROPERTY_URI>${project.basedir}/src/test/resources/configurations/base</BASE_PROPERTY_URI>
+                        <ENVIRONMENT_PROPERTY_URI>${project.basedir}/src/test/resources/configurations/example-env</ENVIRONMENT_PROPERTY_URI>
```
carter-cundiff, line 64: I: Moved unit tests to use existing policies within `/src/test/resources/` rather than writing them to a `/tmp` directory outside of the repo

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/store/ConfigLoader.java
```diff
@@ -82,15 +89,15 @@ public Set<Property> loadConfigs(String baseURI, String environmentURI) {
...
+        return loadPropertyURI(baseURI);
     }
 
     /**
      * Inspects config file(s) at the given URI and gathers all properties.
      * @param URI Location housing configuration yamls.
      * @return Set of Property objects.
      */
-    private Set<Property> loadURI(String URI) {
+    private Set<Property> loadPropertyURI(String URI) {
```
Cho-William, line 100: Q: `loadPropertyURI` sounds to me like it implies that only one property is loaded. Maybe we could call this something like `loadConfigsURI` ?

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/store/PropertyKey.java
```diff
@@ -0,0 +1,67 @@
...
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (!(o instanceof PropertyKey)) {
+            return false;
+        }
+
+        // does not include value to enable check of erroneously indistinguisabe properties,
+        // where two or more properties have identical groupName and propertyName (but different values)
```
Cho-William, line 57: probably ported this over from the other class, but these comments are not relevant for this key class

foundation/aissemble-foundation-core-python/src/policy_manager/policy/json/policy_input.py
```diff
@@ -67,3 +70,23 @@ class PolicyInput(BaseModel):
     The rules for this policy.
     """
     rules: List[PolicyRuleInput] = []
+
+    """
+    This attribute is deprecated and should not be used. Target are now represented as
```
Cho-William, line 75:
```suggestion
    This attribute is deprecated and should not be used. Targets are now represented as
```

foundation/aissemble-foundation-core-python/src/policy_manager/policy/json/policy_input.py
```diff
@@ -67,3 +70,23 @@ class PolicyInput(BaseModel):
...
+    a List of Target's instead of a single Target attribute 'target'.
+    This attribute is replaced by `targets`.
```
Cho-William, line 77:
```suggestion
    This attribute is deprecated and should not be used. Targets are now represented as
    a List of Target objects instead of a single Target attribute 'target'.
    This attribute is replaced by `targets`.
```

foundation/aissemble-foundation-core-python/tests/features/steps/policy_test_util.py
```diff
@@ -106,10 +106,15 @@ def getRandomRules(number: int) -> List[PolicyRuleInput]:
...
-        return Target(retrieve_url=retrieveUrl, type=type)
+    def getRandomTargets(number: int) -> List[Target]:
+        targets = []
+
+        for index in range(number):
+            retrieveUrl = "http://" + PolicyTestUtil.getRandomString(5) + ".com"
+            type = PolicyTestUtil.getRandomString(5)
+            targets.append(Target(retrieve_url=retrieveUrl, type=type))
+
+        return targets
```
Cho-William, line 117: Q: (this isn't really about your PR) but not fully following why we originally had the random component / what this is accomplishing?

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/store/ConfigLoader.java
carter-cundiff: How do you feel about `loadPropertiesURI`. Was trying to avoid using config since it sort of ambiguous whether it would be loading properties or policies.

foundation/aissemble-foundation-core-python/tests/features/steps/policy_test_util.py
carter-cundiff: I guess the idea was to make each target unique. Doesn't really help or harm the testing as is.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/policy/PropertyRegenerationPolicyManager.java
```diff
@@ -0,0 +1,190 @@
...
+            Map<String, Policy> policies = super.getPolicies();
+            if(policies != null) {
+                for (Map.Entry<String, Policy> entry : policies.entrySet()) {
+                    Policy value = entry.getValue();
+                    if(value instanceof PropertyRegenerationPolicy) {
+                        this.propertyRegenerationPolicies.add((PropertyRegenerationPolicy) value);
+                    }
+                }
+            }
+        }
```
Cho-William, line 60: Q: so it seems like we lazily load the regeneration policies if the set is null, but would there ever be a case where the set of regen policies is not null, but it would still be worth checking to see if any regen policies need to be added? Or are we confident that the initial loading is sufficient

carter-cundiff: Given how the policy manager is used in the `configLoader` the set would always be null, so it makes sense to remove this null check.

foundation/aissemble-foundation-core-python/src/policy_manager/policy/json/policy_input.py
```diff
@@ -67,3 +70,23 @@ class PolicyInput(BaseModel):
...
+    """
+    target: Optional[Target]
+
+    def getAnyTargets(self) -> List[Target]:
```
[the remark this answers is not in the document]
carter-cundiff: This method is only used by the `abstract_policy_manager` for getting the `target` or `targets` from the JSON input. So the idea was it would check both attrubutes to support anyone still using the outdated `target` in their JSON policy. In the `Policy.py` it does property interception as that what the downstream user would be interacting with.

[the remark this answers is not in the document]
ewilkins-csi: This is what I mean: https://stackoverflow.com/a/17330273

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/service/ConfigService.java
```diff
@@ -32,14 +36,18 @@ public class ConfigService {
     @Inject
     public ConfigLoader configLoader;
 
+    @Inject
+    public ConfigStoreInit configStoreInit;
```
csun-cpointe, line 40: A: I would prefer that ConfigStoreInit be independent from this class since it fetches its own configLoader and are we going to handle force loading in the loadProperties() function

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/policy/PropertyRegenerationPolicyManager.java
```diff
@@ -0,0 +1,190 @@
...
+    public DefaultPolicy createPolicy(String policyIdentifier) {
+        return new DefaultPropertyRegenerationPolicy(policyIdentifier);
+    }
+
+    /**
+     * Method that allows subclasses to set any additional configurations while
+     * reading a policy.
+     */
+    @Override
+    public void setAdditionalConfigurations(Policy policy, PolicyInput input) {
```
Cho-William, line 94: A: might be worth calling out in the docstring params so laymen can know the difference between policy and input

Cho-William: A: suggesting to refactor the method name to `addPolicy` unless the current naming still makes more sense

carter-cundiff: This is an override from the abstract class, so can't change the method name

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/policy/PropertyRegenerationPolicyManager.java
```diff
@@ -0,0 +1,190 @@
...
+        // Make sure that the policy is the type we're expecting
+        if (policy instanceof DefaultPropertyRegenerationPolicy && input instanceof PropertyRegenerationPolicyInput) {
+            
+            // Set the additional configurations
+            DefaultPropertyRegenerationPolicy regenerationPolicy = (DefaultPropertyRegenerationPolicy) policy;
+            PropertyRegenerationPolicyInput regenerationPolicyInput = (PropertyRegenerationPolicyInput) input;
+
+            regenerationPolicy.setTargetPropertyKeys(convertTargetsToPropertyKeys(regenerationPolicyInput.getAnyTargets()));
+            regenerationPolicy.setPropertyRegenerationRules(convertInputRulesToPropertyRegenerationRules(regenerationPolicyInput.getRules()));
+            regenerationPolicy.setRegenerationStrategy(convertInputRuleToPropertyRegenerationStrategy(regenerationPolicyInput.getRegenerationMethod()));
```
Cho-William, line 105:
```suggestion
            regenerationPolicy.setPropertyRegenerationRules(convertInputRulesToPropertyRegenerationRules(regenerationPolicyInput.getRules()));
            regenerationPolicy.setRegenerationStrategy(convertInputRegenMethodToPropertyRegenerationStrategy(regenerationPolicyInput.getRegenerationMethod()));
```

A: would this refactoring make it a bit more consistent?

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/policy/PropertyRegenerationPolicyManager.java
```diff
@@ -0,0 +1,190 @@
...
+                throw new PropertyRegenerationPolicyException("Target's 'retrieve_url' must be in '/group_name/property_name` format " + 
+                                                                "for a Property Regeneration Policy");
+            }
+        }
+
+        return propertyKeys;
+    }
+
+    /**
+     * Method for validating and converting {@link PolicyRuleInput}'s to their respective {@link PropertyRegenerationRule}'s.
```
Cho-William, line 148:
```suggestion
     * Method for validating and converting {@link PolicyRuleInput}'s rules to their respective {@link PropertyRegenerationRule}'s.
```

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/policy/PropertyRegenerationPolicyManager.java
Cho-William: Herm, I see, but the regen method is still a `PolicyInputRule` type so maybe not

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/ConfigStoreInit.java
```diff
@@ -20,39 +23,82 @@
...
+import org.apache.commons.lang3.StringUtils;
+
 /**
  * ConfigStoreInit loads configuration properties as application starts up
  */
 @Startup
 @ApplicationScoped
 public class ConfigStoreInit {
     private static final Logger logger = LoggerFactory.getLogger(ConfigStoreInit.class);
+    private static Map<String, String> status = new HashMap<>();
```
csun-cpointe, line 40: Q: Can we just use String for the status since the status can only be "loading completed" or "Skip Load"? If we can, probably makes sense to create a enum for the status

foundation/foundation-configuration-store/src/test/java/com/boozallen/aissemble/util/ConfigStoreInitTestHelper.java
```diff
@@ -1,154 +0,0 @@
...
-@ApplicationScoped
-public class ConfigStoreInitTestHelper extends ConfigStoreInit {
-    private static final Logger logger = LoggerFactory.getLogger(ConfigStoreInitTestHelper.class);
-    public static Map<String, String> verification = new HashMap<>();
-    public static boolean skiploadingNotificationSent;
-    private static ConfigLoader configLoader;
-
-    @PostConstruct
-    protected void init() {
-        prepareTestEnv();
```
csun-cpointe, line 52: Q: do we still need this for the test?

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/service/ConfigService.java
carter-cundiff: If we make these classes independent, then we have to duplicate the logic that currently lives in `ConfigStoreInit.init()` and maintain it in two places, so was trying to avoid that. However if we envision the `ConfigService` being slightly different, which I think we do with the force reloading, then I think it makes sense to duplicate that logic for now with the understanding that it will eventually be different.

foundation/foundation-configuration-store/src/test/java/com/boozallen/aissemble/util/ConfigStoreInitTestHelper.java
carter-cundiff: The tests now use our config files that live in `src/test/resources` so this function was no longer needed for setup

foundation/aissemble-foundation-core-python/src/policy_manager/policy/json/policy_input.py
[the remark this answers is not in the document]
ewilkins-csi: Yeah I totally missed that this was the JSON mapping class.  You can ignore this.

foundation/aissemble-foundation-core-python/src/policy_manager/policy/policy.py
```diff
@@ -49,9 +50,49 @@ class ConfiguredRule(BaseModel):
...
+        a List of ConfiguredTarget's instead of a single ConfiguredTarget attribute `targetConfigurations`.
+        This attribute is replaced by `configuredTargets`.
+        """
+        ConfiguredRule.__logger.warn(
+            "Detected use of deprecated attribute 'targetConfigurations'. Existing "
+            + "usage should be moved to the new attribute 'configuredTargets'."
+        )
+        return self.configuredTargets[0] if len(self.configuredTargets) > 0 else None
+
+    def set_deprecated_targetConfigurations(self, new_value: ConfiguredTarget):
```
ewilkins-csi, line 71: Q/A: Do we need to support the setter? I'd have thought this was a read-only property that's determined by the definition JSON file.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/ConfigStoreInit.java
```diff
@@ -20,39 +23,82 @@
...
+        String basePropertyUri = System.getenv("BASE_PROPERTY_URI");
+        String environmentPropertyUri = System.getenv("ENVIRONMENT_PROPERTY_URI");
+        String basePolicyUri = System.getenv("BASE_POLICY_URI");
+        String environmentPolicyUri = System.getenv("ENVIRONMENT_POLICY_URI");
+
         try {
             ConfigLoader configLoader = CDI.current().select(ConfigLoader.class,new Any.Literal()).get();
             if (configLoader.isFullyLoaded()) {
                 logger.info("Properties are already fully loaded. Skipping reload.");
+                status.put("status","Skip Load");
```
ewilkins-csi, line 54: S: Consider using an `enum` instead of a magic string.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/service/ConfigService.java
ewilkins-csi: We should just drop the `/load` endpoint IMO. It wasn't in our original feature roadmap, and it seems more like a security concern than anything.

foundation/aissemble-foundation-core-python/src/policy_manager/abstract_policy_manager.py
```diff
@@ -186,19 +186,24 @@ def validateAndConfigureRule(
...
-            configuredTarget = None
-            if targetConfigurations:
-                configuredTarget = ConfiguredTarget(
-                    target_configurations=targetConfigurations
-                )
-                if target:
-                    configuredTarget.retrieve_url = target.retrieve_url
-                    configuredTarget.type = target.type
+            configuredTargets = []
+            if targets is not None and targets:
```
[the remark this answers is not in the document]
carter-cundiff: Updates

Cho-William (approved):

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/service/ConfigService.java
csun-cpointe: Yeah the function is not required but kept based on the idea that user can have a way for user to reload the property. If it becomes a overhead and that's not a function that we are interested to support reload function, which I think we can do it by restart the service. Let's drop it to keep thing simple.

carter-cundiff: Removed

foundation/aissemble-foundation-core-python/src/policy_manager/policy/policy.py
carter-cundiff: In theory we probably don't since I think these would only be set by the abstract policy manager which does it through the new property. However the java implementation has a now deprecated setter for it, so can't hurt to maintain parity here.

PR #155, opened 2024-06-17
#118 Store map of deprecated methods on class instead of Pydantic config

carter-cundiff:
Resolved pylint error:
```
[INFO] ************* Module policy_manager.policy.policy
[INFO] src/policy_manager/policy/policy.py:86:17: E1101: Class 'BaseConfig' has no 'property_set_methods' member (no-member)
[INFO] src/policy_manager/policy/policy.py:146:17: E1101: Class 'BaseConfig' has no 'property_set_methods' member (no-member)
```

### DIFF
--- a/foundation/aissemble-foundation-core-python/src/policy_manager/policy/policy.py
+++ b/foundation/aissemble-foundation-core-python/src/policy_manager/policy/policy.py
@@ -51,6 +51,7 @@ class ConfiguredRule(BaseModel):
     """
 
     __logger = LogManager.get_instance().get_logger("ConfiguredRule")
+    __deprecated_set_methods = {"target": "set_deprecated_targetConfigurations"}
     className: str
     configurations: Optional[Dict[str, Any]]
     configuredTargets: Optional[List[ConfiguredTarget]] = []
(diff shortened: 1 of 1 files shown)


foundation/aissemble-foundation-core-python/src/policy_manager/policy/policy.py
```diff
@@ -83,17 +84,12 @@ def set_deprecated_targetConfigurations(self, new_value: ConfiguredTarget):
     # Links ConfiguredRule 'targetConfigurations' attribute to 'set_deprecated_targetConfigurations()' method to support
     # people still assigning values to the old attribute.
     def __setattr__(self, key, val):
-        method = self.__config__.property_set_methods.get(key)
+        method = self.__deprecated_set_methods.get(key)
```
peter-mcclonski, line 87: Q: what's actually happening here? I'm not very familiar with pylint and it's not obvious to me what this change does, functionally.

carter-cundiff: We're deprecating the `target` attribute on a policy to instead be a list of targets. People still using the old attribute like this:
```
MyPolicy.target = MyTarget
```
will get a nice warning message. This just ties our custom function to the Python setter.

carter-cundiff: `property_set_methods` did not exist on the pydantic config class, so pylint through an error. We're moving the attribute to just live on our local class instead.

peter-mcclonski (approved):

Issue #166, opened 2024-06-25
Feature: Update messaging integration test infrastructure.


ewilkins-csi:
## Description
Several of the images/libraries we use to execute integration tests are a bit out of date.  We should be update them to leverage the latest enhancements and security fixes. The following can be updated:

- TestContainers
- RedPanda (lightweight Kafka impl)
- Artemis (lightweight ActiveMQ impl)

## DOD
Acceptance criteria required to realize the requested feature
- [x] Update TestContainers in POMs and pyproject.toml files
- [x] Update redpanda images
- [x] Update artemis images

## Test Strategy/Script
OTS:
 * Run `mvn clean install -Pintegration-test` and ensure build passes

Final Test:
 * Verify latest build in CI passed

ewilkins-csi:
**DoD**/**OTS** with @Cho-William

Cho-William:
Final test passed, see screenshot:
<img width="1272" alt="Screenshot 2024-06-25 at 12 35 14 PM" src="https://github.com/boozallen/aissemble/assets/149192267/b0f83b87-d1ed-4fec-8b65-64003ab27655">

PR #170, opened 2024-06-25
#166 configure quarkus to use latest redpanda image

ewilkins-csi:
This is a follow-up PR to specifically set the Kafka dev services images to the latest version of Red Panda in all Quarkus modules that enable the Kafka extension for Quarkus.

In the mid-term, we should look to update Quarkus, as I believe the newer versions of Quarkus would pull in the latest Red Panda image automatically.


extensions/extensions-data-lineage/extensions-data-lineage-http-consumer-service/pom.xml
```diff
@@ -1,213 +1,215 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
```
ewilkins-csi, line 2: I: This is a white-space only change

extensions/extensions-data-lineage/extensions-data-lineage-http-consumer-service/src/test/resources/application.properties
```diff
@@ -9,5 +9,6 @@
 ###
 quarkus.rest-client."com.boozallen.aissemble.datalineage.consumer.HttpProducerService".uri=http://127.0.0.1:15100/endpoint
 quarkus.amqp.devservices.image-name=quay.io/artemiscloud/activemq-artemis-broker:1.0.26
+quarkus.kafka.devservices.image-name=vectorized/redpanda:v24.1.7
```
carter-cundiff, line 12: Any reason not to use `vectorized/redpanda:latest`?

carter-cundiff (approved):

extensions/extensions-data-lineage/extensions-data-lineage-http-consumer-service/src/test/resources/application.properties
d-ryan-ashcraft: Generally, we want to avoid LATEST b/c it normally just means "most recently pushed" vice "newest release".  Also better for version control to know what we depend one (Python 'best practices' withstanding)

d-ryan-ashcraft (approved):

Issue #156, opened 2024-06-18
Feature: Implement universal config interface for simplified local testing


cwoods-cpointe:
[156-helper.zip](https://github.com/user-attachments/files/15976191/156-helper.zip)
## Description
The current universal config interface has some burden when testing locally. This issue should tackle making it faster and easier. Currently to test the config locally we are required to create a new docker image/container from the baselines config-store-service image. We then load the config directly onto the service using the new docker image. Ideally the config-store-service should be able to pull config from multiple different locations and not just from files on its local container path.

## DOD
- [x] Create a way for mounting local config file onto the cluster, removing the need to create new docker image
- [x] Update the config-store-service to pull from this location as opposed to files on its local path
- [x] if required, create a dedicated helm chart for the config-store-service
  - [x] Create unit tests
  - [x] Create mda generation
  - [x] Create manual action for tilt

## Test Strategy/Script

- Build aissemble
```
mvn clean install
```
- Create a downstream project with:
```
mvn archetype:generate -U -DarchetypeGroupId=com.boozallen.aissemble \
  -DarchetypeArtifactId=foundation-archetype \
  -DarchetypeVersion=1.8.0-SNAPSHOT \
  -DgroupId=com.boozallen.aissemble.test \
  -DartifactId=test-issue-156 \
  -DprojectGitUrl=git.example.com/test-issue-156 \
  -DprojectName="Issue#156 Test"
```
- Copy the src directory from the attached folder and put it into the root of the projects
- Add to the fermenter-mda plugin executions in test-issue-156-deploy/pom.xml
```
<execution>
    <id>configuration-store</id>
    <phase>generate-sources</phase>
    <goals>
        <goal>generate-sources</goal>
    </goals>
    <configuration>
        <basePackage>com.boozallen.aissemble.test</basePackage>
        <profile>configuration-store-deploy-v2</profile>
        <!-- The property variables below are passed to the Generation Context and utilized
             to customize the deployment artifacts. -->
        <propertyVariables>
            <appName>configuration-store</appName>
        </propertyVariables>
    </configuration>
</execution>
```
- Add the config-store-service to tilt with:
```
project_path = os.path.abspath('.')
# Update configuration_files_path to match the path of the store config files to be loaded into the configuration store
configuration_files_path = 'src/main/resources/configurations'
# For WSL users, the configuration files need to be in an accessible/shared path. Use this prefix to add the path to the root shared file store. Example: /mnt/c, /mnt/wsl/rancher-desktop
[...]

PR #172, opened 2024-06-25
 #156 Create helm chart for the configuration-store

cwoods-cpointe:
Created mda the new helm chart. Part of this stubs out mounting local files to the config-store for easier local config testing. Created the manual action for adding to tilt. It will be displayed in a future release when the configuration store is flushed out


extensions/extensions-helm/aissemble-configuration-store-chart/README.md
```diff
@@ -0,0 +1,42 @@
...
+| image.name                              | The image name                                                                                                                   | Yes               | boozallen/aissemble-configuration-store                                                                                                                                        | 
+| image.tag                               | The image tag. Will override the default chart AppVersion value                                                                  | No                | Chart.AppVersion                                                                                                                                                               |
+| image.imagePullPolicy                   | The image pull policy                                                                                                            | No                | Always                                                                                                                                                                         |
```
ewilkins-csi, line 26: A: 
```suggestion
| image.pullPolicy                   | The image pull policy                                                                                                            | No                | Always                                                                                                                                                                         |
```

extensions/extensions-helm/aissemble-configuration-store-chart/README.md
```diff
@@ -0,0 +1,42 @@
...
+| deployment.ports                        | The deployment ports                                                                                                             | No                | - name: http-1 <br/>&emsp;&emsp;containerPort: 8080 <br/>&emsp;&emsp;protocol: TCP                                                                                             |
+| service.spec.type                       | The service type                                                                                                                 | No                | None                                                                                                                                                                           | 
+| service.spec.ports                      | The service ports                                                                                                                | No                | - name: http <br/>&emsp;&emsp;port: 8083<br/>&emsp;&emsp;protocol: TCP<br/>&emsp;&emsp;targetPort: 8080                                                                        |
```
ewilkins-csi, line 38: Q/S: Can we simplify this so that we just use the same port for the deployment spec and the service spec? Could always add more specific overrides later if they're needed.  Also recommend dropping `spec` e.g. `service.ports`.

extensions/extensions-helm/aissemble-configuration-store-chart/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v2
+name: aissemble-configuration-store-chart
```
ewilkins-csi, line 2: Q/A: Do we think we should go ahead and add this to the `containers.adoc` page?  I know the feature isn't technically ready yet, but this feels like one of those things we'll overlook if we don't do it when we add the chart.

extensions/extensions-helm/aissemble-configuration-store-chart/README.md
cwoods-cpointe: Yeah thats not a problem.

extensions/extensions-helm/aissemble-configuration-store-chart/values.yaml
```diff
@@ -0,0 +1,35 @@
...
+  name: ''
+  # Sets the labels for app.kubernetes.io/version. Default is Chart.appVersion
+  version: ''
+
+hostname: configuration-store
+image:
+  name: boozallen/aissemble-configuration-store
+  imagePullPolicy: Always
+  dockerRepo: ghcr.io/
+  tag: "1.8.0-SNAPSHOT"
```
ewilkins-csi, line 13: C: We should avoid hard-coding this value.  We should be able to just leave it blank and default it to `Chart.AppVersion` in the templates. E.g. `{ Values.image.tag | default Chart.AppVersion }`. Otherwise this will have to be updated during each release and could be missed causing the version to be out of sync.

cwoods-cpointe: Yeah thats a great point!

foundation/foundation-mda/src/main/resources/templates/deployment/configuration-store/v2/configuration-store.values.yaml.vm
```diff
@@ -0,0 +1,30 @@
...
+
+  deployment:
+    env:
+      - name: BASE_PROPERTY_URI
+        value: /configurations/base
+      - name: ENVIRONMENT_PROPERTY_URI
+        value: /configurations/env
+
+  ingress:
+    enabled: true
```
ewilkins-csi, line 17: S: I'd consider removing this.  It's unlikely that any project will want to expose their config store over the internet. It's meant for in-cluster use.

extensions/extensions-helm/aissemble-configuration-store-chart/Chart.yaml
cwoods-cpointe: I was thinking we would create one for the release its going into so people dont try and use it thinking its ready. If you dont think thats an issue then I see no problem adding it

foundation/foundation-mda/src/main/resources/templates/notifications/notification.configuration.store.tilt.vm
```diff
@@ -0,0 +1,14 @@
+project_path = os.path.abspath('.')
+# Update configuration_files_path to match the path of the config files to be loaded into the configuration store. Example 'src/main/resources/config'
+configuration_files_path = ''
+# For WSL users, the configuration files need to be in an accessible path. Use this prefix to add the path to the root shared file store. Example: '/mnt/c' or '/mnt/wsl/rancher-desktop'
+configuration_files_path_prefix = ''
```
ewilkins-csi, line 5: A: Since we expect this to be used only for WSL, I might make the variable name indicate that. E.g. `wsl_path_prefix`

foundation/foundation-mda/src/main/resources/templates/deployment/configuration-store/v2/configuration-store.values.yaml.vm
cwoods-cpointe: to that point, you think i should remove ingress all together?

foundation/foundation-mda/src/main/resources/templates/notifications/notification.configuration.store.tilt.vm
```diff
@@ -0,0 +1,14 @@
+project_path = os.path.abspath('.')
+# Update configuration_files_path to match the path of the config files to be loaded into the configuration store. Example 'src/main/resources/config'
```
ewilkins-csi, line 2: Q/A: This would be relative to the directory from which `tilt up` is run, right? I might suggest a path that makes that more clear like `my-project-deploy/src/main/resources/config`.

foundation/foundation-mda/src/main/resources/templates/deployment/configuration-store/v2/configuration-store.values.yaml.vm
ewilkins-csi: I was thinking about that.  Maybe so.  If someone was deadset on it, they could add an ingress and wire it up themselves. But I think our general advice would be to avoid it, since it's introduces a lot of security concerns that you have to be careful about.

extensions/extensions-helm/aissemble-configuration-store-chart/Chart.yaml
ewilkins-csi: Yeah I could see the argument either way. Just worried we'll forget 😅

ewilkins-csi (approved):

extensions/extensions-helm/aissemble-configuration-store-chart/Chart.yaml
cwoods-cpointe: Ill go ahead and add it to the "Available V2 Profiles"

foundation/foundation-mda/src/main/resources/templates/deployment/configuration-store/v2/configuration-store.values.yaml.vm
cwoods-cpointe: I think thats valid. Ill delete it and we can always add it back in later if the need arises. I would rather add it when we need it than add it now to have it never used

Issue #171, opened 2024-06-25
Feature: As an aiSSEMBLE user, I want the MDA Docker module templates to leverage fabric8, so that I can manage Docker builds with fabric8 in downstream projects


Cho-William:
## Description
This issue is part of a series of issues targeting moving aiSSEMBLE from [orphedomos](https://github.com/TechnologyBrewery/orphedomos) to [fabric8](https://github.com/fabric8io/docker-maven-plugin). Issue #168 is targeting making this update in the baseline's python Docker modules, and Issue #154 is targeting one (separate) Docker module in the baseline, as well as the migration for downstream projects. 

This issue addresses updating the baseline's MDA generation such that downstream projects will leverage fabric8 instead of orphedomos.  

## DOD
- [x] Update relevant files in `foundation-mda` and `foundation-archetype` to leverage fabric8's [docker-maven-plugin](https://github.com/fabric8io/docker-maven-plugin) instead of orphedomos for the relevant downstream project's `pom.xml` files (see implementation details)
- [x] Ensure updated templates and generation are (reasonably) in line with the Issue [#154](https://github.com/boozallen/aissemble/issues/154) migrations
- [x] Ensure the docker image build runs as part of the default maven build (see Implementation Details and Affected Files)
- [x] Update the orphedomos sagemaker configurations manual actions to reference the correct fabric8 configurations
- [x] Update the release notes draft covering the change to fabric8
- [x] Update the antora docs for SageMaker configuration change
- [x] Adjust the archetype resources pom.xml to include `foundation-mda` and `foundation-upgrade` as (workaround) dependencies

## DOD Questions
- are we no longer going to force users to pass an extra argument to run the docker module builds
  - yes, although it seems there are redundant docker builds in the tilt file, the cache should have them available

## Implementation Details
- templates/archetype poms for downstream projects to change will include:
  -  the parent docker module `pom.xml` found in `<project-name>/<project-name>-docker`  
  - each individual docker module's `pom.xml` found in `<project-name>/<project-name>-docker/<project-name>-*`
  - the test docker `pom.xml` found in `<project-name>/<project-name>-tests/<project-name>-tests-docker`
- existing profile that disables docker builds by default will be removed for now, but will likely be reintroduced in the future
- testing changes for this issue will rely on the completion of the two related Issues mentioned above

## Affected pom.xml Files
- `foundation/foundation-archetype/src/main/resources/archetype-resources/__rootArtifactId__-docker`
[...]

PR #188, opened 2024-07-03
#171 updating docker templates to fabric8

Cho-William:
- updated archetype and foundation-mda templates to leverage fabric8 docker-maven-plugin in lieu of orphedomos
 - sagemaker antora documentation was updated to reflect updated configuration of ECR registry URL
 - machine learning pipeline and notifications to set the ECR registry URL has been updated
 - fixed null pointer exception bug with check for onnx conversion generation (MachineLearningPipelineGenerator)
 - release notes draft updated noting potential breaking changes regarding loss of extra orphedomos configs
 - update the orphedomos to fabric8 migration to include the default skip configuration

## IMPORTANT NOTE
Please note that this PR disables fabric8's docker builds as part of a default build for a downstream project:
- This configuration is consistent with the previous `orphedomos` configuration, although not leveraging this `skip` behavior would have been preferred
- Unfortunately, if a user were to generate a project and exclusively use `mvn clean install` rather than `mvn clean generate-sources` (which is how the archetype test script also does it) to create their project structure, the user may eventually encounter a broken build. Consider the following example:
  - The user begins generating a downstream project with only `mvn clean install` calls
  - The user's machine learning inference docker module is generated
  - The user builds again, and if the docker fabric8 build is not skipped, then that docker module will begin building
    - However, the generation for the machine learning inference pipeline step has not yet been generated at this point
  - Finally, the inference docker image expects a copy of ie the `sdist` `requirements.txt` from the inference pipeline step to be available in the docker build context/directory, but that ml pipeline inference step module has not yet been built yet
    - The reason why the inference pipeline step module has not yet been generated is because it is a sub-sub directory of the `-pipelines` module, and requires another `generate-sources` execution. Unfortunately, this source generation occurs after the aforementioned docker build.
    - Thus, previous archetype tests and downstream projects did not encounter this issue, as orphedomos docker builds were also disabled by default

Therefore, there are two options to remedy this issue for the archetype test script and downstream projects:
1. Simply keep the `skip` configuration as part of the default fabric8 docker build configuration
2. Update `foundation-mda`'s generation logic to automatically include sub-sub step modules as declared child modules in their respective pipeline parent poms. Namely this applies to machine learning pipelines, and any other pipelines that house child modules.

Unfortunately, option 2 was not able to be incorporated in the scope of this ticket, however it is in the interest of the aiSSEMBLE team to eventually implement option 2 and revert the default skip behavior


DRAFT_RELEASE_NOTES.md
```diff
@@ -100,11 +104,16 @@ extension-pkg-whitelist = "pydantic"
 ignore-patterns = '.*pb2[\S]*.py'
 ```
 
+### Upgrade Steps for Transitioning from Orphedomos to Fabric8
+If any extra configurations were added to the `orphedomos-maven-plugin` in addition to the generated defaults, executing the migration **will result in loss of these extra configurations**. To facilitate the upgrade, the following steps should be taken:
+1. Before executing the baton migration, it is recommended to ensure some form of version control is in place to preserve your existing `orphedomos-maven-plugin` configurations.
```
chang-annie, line 109: looks like these bullets should be 1., 2., 3., instead of all 1's

Cho-William: That is intentional. That way if we insert another point, we don't have to manually renumber. Markdown is smart enough to increment the numbers when the README is rendered

chang-annie: oo good to know!

chang-annie (approved):

foundation/foundation-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -84,6 +84,18 @@
             <version>${archetypeVersion}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
```
d-ryan-ashcraft, line 87: S: are we sure we need `foundation-mda`?  It's not really a maven plugin, but rather a dependency that is fed to fermenter.  I would have though that the normal reactor would pick it up for the cache without explicitly being set.  

If it is that the build cache isn't picking up dependencies listed within plugins, I might just note that so it's easier to know what it is here in the future.

d-ryan-ashcraft (approved):

DRAFT_RELEASE_NOTES.md
```diff
@@ -100,11 +104,16 @@ extension-pkg-whitelist = "pydantic"
 ignore-patterns = '.*pb2[\S]*.py'
 ```
 
+### Upgrade Steps for Transitioning from Orphedomos to Fabric8
```
ewilkins-csi, line 107: A: Maybe _"Upgrade Steps for Projects with Customized Orphedomos Configurations"_? Usually we try to convey what conditions would cause the instructions to be relevant to a project in these titles.

foundation/foundation-archetype/src/main/resources/archetype-resources/pom.xml
Cho-William: I believe it is indeed the latter case, because after adding these two dependencies, it resolved the inconsistent behavior with the archetype tests. I will add a comment to contextualize

Cho-William: Will also make sure the additions are not part of the plugins comment block

Issue #150, opened 2024-06-14
As a user, I can specify a single monorepo Habushu module and Habushu will stage the source code required to build that module in my build directory, so I can containerize my python services


Cho-William:
## Description
This ticket is the first of a series to implement a more resilient and consistent pattern for installing Python monorepo dependencies to Docker containers. The scope of this ticket will cover augmenting Habushu's existing functionality to utilize the Maven session to stage Python module source code files to the docker build context (without needing to specify any relative-paths for monorepo Python dependencies), and leveraging these changes with the aiSSEMBLE baseline's relevant docker images.

## DOD
- [x] Update the relevant docker module poms in the aiSSEMBLE baseline:
  - [x] Remove the Orphedomos `retrieve-wheels` execution
  - [x] Add the new Habushu execution of `containerize-dependencies`
- [x] Update the relevant docker module Dockerfiles in the aiSSEMBLE baseline:
  - [x] Modify build to work with the Habushu-injected logic
- [x] Ensure the modified docker modules in the baseline can build successfully and run

## Test Steps
1. _**OTS Only**_:  Build python Docker images in the baseline and ensure they build successfully
```
mvn clean install -am -pl :aissemble-model-training-api,:aissemble-model-training-api-sagemaker,:aissemble-versioning
```
2. Run `aissemble-model-training-api-sagemaker` and ensure it starts successfully (i.e. `uvicorn` starts up):
```
docker run --rm -it ghcr.io/boozallen/aissemble-model-training-api-sagemaker:1.8.0-SNAPSHOT
```
3. Create a new project:
```
mvn archetype:generate -DarchetypeGroupId=com.boozallen.aissemble \
                       -DarchetypeArtifactId=foundation-archetype \
                       -DarchetypeVersion=1.8.0-SNAPSHOT \
                       -DgroupId=org.test \
                       -DartifactId=test-150 \
                       -DprojectGitUrl=test.org/test-150.git \
                       -DprojectName="Test Issue 150" \
&& cd test-150
```
4. Add [TrainingPipeline.json](https://github.com/user-attachments/files/16132131/TrainingPipeline.json) to `test-150-pipeline-models/src/main/resources/pipelines`
5. Run `mvn clean install` and resolve manual actions
7. Repeat until no more actions are printed
8. Deploy the project and wait for everything to start: `tilt up`
9. Verify that the `versioning` service started up and is listening on port 80
10. Submit the training job to the model-training-api:
```
[...]

PR #192, opened 2024-07-08
#150 use Habushu's `containerize-dependencies` goal

ewilkins-csi:
The vast majority of the changes files are just updating the formatting of `ENV` and `LABEL` commands in Dockerfiles to address a build warning.  The primary files for this change were the POMs and the 3 FastAPI-based Dockerfiles:
 - aissemble-model-training-api
 - aissemble-model-training-api-sagemaker
 - aissemble-versioning
 
 Also fixed a few issues in `foundation-mda` found during OTS.

### DIFF
--- a/extensions/extensions-docker/aissemble-airflow/src/main/resources/docker/Dockerfile
+++ b/extensions/extensions-docker/aissemble-airflow/src/main/resources/docker/Dockerfile
@@ -1,7 +1,7 @@
 # Script for creating base Airflow Docker image
 FROM apache/airflow:2.7.0-python3.11
 
-LABEL org.opencontainers.image.source = "https://github.com/boozallen/aissemble"
+LABEL org.opencontainers.image.source="https://github.com/boozallen/aissemble"
 
 USER root
 
@@ -37,10 +37,10 @@ USER root
 RUN mkdir -p /tmp/mlruns /tmp/model
 RUN chown airflow:0 -R /tmp/mlruns /tmp/model
 
-ENV JAVA_HOME /usr/lib/jvm/default-java
+ENV JAVA_HOME=/usr/lib/jvm/default-java
 RUN export JAVA_HOME
 
-ENV DEBIAN_FRONTEND noninteractive
+ENV DEBIAN_FRONTEND=noninteractive
 ENV TERM linux
 
 # Airflow UID is 50000
--- a/extensions/extensions-docker/aissemble-model-training-api-containers/aissemble-model-training-api-sagemaker/src/main/resources/docker/Dockerfile
+++ b/extensions/extensions-docker/aissemble-model-training-api-containers/aissemble-model-training-api-sagemaker/src/main/resources/docker/Dockerfile
@@ -1,37 +1,43 @@
 # Script for creating base AIOps Model Training API image
-
 ARG DOCKER_BASELINE_REPO_ID
 ARG VERSION_AISSEMBLE
+
+#HABUSHU_BUILDER_STAGE - HABUSHU GENERATED CODE (DO NOT MODIFY)
+FROM python:3.11-slim AS habushu_builder
+# Poetry and supporting plugin installations
+RUN python -m ensurepip --upgrade && \
+    pip install poetry && \
+    poetry self add poetry-monorepo-dependency-plugin && \
+    poetry self add poetry-plugin-bundle
+
+WORKDIR /work-dir
+COPY --chown=1001 target/containerize-support ./containerize-support/
+RUN find . -type f -name pyproject.toml -exec sed -i 's|develop[[:blank:]]*=[[:blank:]]*true|develop = false|g' {} \;
+
+USER root
+WORKDIR /work-dir/containerize-support/extensions/aissemble-extensions-model-training-api-sagemaker
+ENV POETRY_CACHE_DIR="/.cache/pypoetry"
+
+# export target project's virtual environment to /opt/venv
+RUN --mount=type=cache,target=/.cache/pypoetry/ \
+    poetry lock && \
+    poetry bundle venv /opt/venv
+#HABUSHU_BUILDER_STAGE - HABUSHU GENERATED CODE (END)
+
 FROM ${DOCKER_BASELINE_REPO_ID}boozallen/aissemble-fastapi:${VERSION_AISSEMBLE} AS builder
 
-RUN python -m venv /opt/venv
+#HABUSHU_FINAL_STAGE - HABUSHU GENERATED CODE (DO NOT MODIFY)
+FROM python:3.11-slim AS final
+# Copy venv from builder and activate by adding to the path
+COPY --from=habushu_builder --chown=1001 /opt/venv /opt/venv/
 ENV PATH="/opt/venv/bin:$PATH"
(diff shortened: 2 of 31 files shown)


extensions/extensions-docker/aissemble-airflow/src/main/resources/docker/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # Script for creating base Airflow Docker image
 FROM apache/airflow:2.7.0-python3.11
 
-LABEL org.opencontainers.image.source = "https://github.com/boozallen/aissemble"
+LABEL org.opencontainers.image.source="https://github.com/boozallen/aissemble"
```
ewilkins-csi, line 4: I: This and the `ENV` format change are to address a deprecation warning that was output by the docker-maven-plugin

extensions/extensions-docker/aissemble-model-training-api-containers/aissemble-model-training-api-sagemaker/src/main/resources/docker/Dockerfile
```diff
@@ -1,37 +1,43 @@
...
 
-# Copy over the virtual environment (and all pip installed packages) from the initial builder stage
-COPY --from=builder /opt/venv /opt/venv
+LABEL org.opencontainers.image.source="https://github.com/boozallen/aissemble"
 
-COPY ./target/model_training_api_sagemaker.py /app/model_training_api_sagemaker.py
+# Set/move any required environmental variables and key scripts/configs needed by FastAPI from the base image
+# into the slimmed down final Python image
+ENV API_MAIN=model_training_api_sagemaker.model_training_api_sagemaker
+COPY --from=builder --chown=1001 /start.sh /start.sh
```
ewilkins-csi, line 41: I: As I was updating the image to work with the new Habushu feature (including running the APIs as a module instead of directly via a `.py` file) I noticed that we're largely just using the `aissemble-fastapi` image as a way to pass this start script.  At the end of the day, these images have the same base as `aissemble-fastapi`, so I think we could probably just use that image for the `dockerBase` configuration of Habushu?  However, it also sets up an example API server which complicated things a bit.  This is at least better than we found it.

extensions/extensions-docker/aissemble-versioning/src/main/resources/docker/Dockerfile
```diff
@@ -43,27 +57,15 @@ RUN echo "deb http://deb.debian.org/debian bullseye main" >> /etc/apt/sources.li
...
+ENV API_MAIN=model_versioning.versioning_api
 COPY --from=builder /start.sh /start.sh
-
-# Copy over the virtual environment (and all pip installed packages) from the initial builder stage
-COPY --from=builder /opt/venv /opt/venv
 COPY --from=builder /app /app
 
-COPY ./target/versioning_api.py /app/versioning_api.py
-
-# current workaround for arm-based machines to have cryptography installed properly for Krausening
```
ewilkins-csi, line 64: I: Latest version of cryptography fully supports ARM

foundation/foundation-mda/src/main/resources/profiles.json
```diff
@@ -350,9 +350,6 @@
 			},
 			{
 				"name": "dataAccessDockerPomFile"
-			},
-			{
-				"name": "vaultDockerPomFile"
```
ewilkins-csi, line 355: I: Disable vault inclusion by default, since UC isn't ready to be rolled out yet.

foundation/foundation-mda/src/main/resources/templates/deployment/model-training-api/model-training-api.values.yaml.vm
```diff
@@ -21,8 +21,8 @@ service:
 configMap:
   trainingApiConfig: |
     project_name=${projectName}
-    docker_registry=ghcr.io/
-    image_prefix=boozallen/
+    docker_registry=${dockerProjectRepositoryUrl}
+    image_prefix=
```
ewilkins-csi, line 25: I: This chart runs a project-specific image, and so needs to run with the project-specific settings.

carter-cundiff (approved):

extensions/extensions-docker/aissemble-versioning/src/main/resources/docker/Dockerfile
```diff
@@ -1,33 +1,47 @@
...
 ARG DOCKER_BASELINE_REPO_ID
 ARG VERSION_AISSEMBLE
-FROM ${DOCKER_BASELINE_REPO_ID}boozallen/aissemble-fastapi:${VERSION_AISSEMBLE} AS builder
 
-RUN python -m venv /opt/venv
-ENV PATH="/opt/venv/bin:$PATH"
+#HABUSHU_BUILDER_STAGE - HABUSHU GENERATED CODE (DO NOT MODIFY)
+FROM python:3.11-slim AS habushu_builder
+# Poetry and supporting plugin installations
+RUN python -m ensurepip --upgrade && \
```
d-ryan-ashcraft, line 8: S: not for adjudication here, but I wonder if we would benefit from just doing most of this in a docker image one and then using that as a builder?

ewilkins-csi: Yeah it's an interesting thought.  Because the builder image and the final image have to have the same base platform, Habushu allows you to set the base image and doesn't assume anything about that image other than `python` is resolvable on the `PATH`.  For aissemble, we could disable Habushu's dockerfile management but I think the two things you'd get (build speed and ease of maintenance) are covered by the docker layer cache and Habushu's code generation (respectively).  Could be missing a trick though.

extensions/extensions-docker/aissemble-model-training-api-containers/aissemble-model-training-api-sagemaker/src/main/resources/docker/Dockerfile
```diff
@@ -1,37 +1,43 @@
 # Script for creating base AIOps Model Training API image
-
 ARG DOCKER_BASELINE_REPO_ID
 ARG VERSION_AISSEMBLE
-FROM ${DOCKER_BASELINE_REPO_ID}boozallen/aissemble-fastapi:${VERSION_AISSEMBLE} AS builder
 
-RUN python -m venv /opt/venv
-ENV PATH="/opt/venv/bin:$PATH"
+#HABUSHU_BUILDER_STAGE - HABUSHU GENERATED CODE (DO NOT MODIFY)
```
csun-cpointe, line 5: Q: although this will be replaced each build, we always want to check in the updated Dockerfile and that's why this is included?

csun-cpointe (approved):

extensions/extensions-docker/aissemble-model-training-api-containers/aissemble-model-training-api-sagemaker/src/main/resources/docker/Dockerfile
ewilkins-csi: Yeah I think we treat it just like we do the `pyproject.toml` file changes that Habushu makes.  They are fairly stable and should only change when the underlying logic changes.

PR #196, opened 2024-07-09
#150 revert fastapi image to `python:3.11` and re-use default docker cache

ewilkins-csi:
Technically, the failures in CI were due to two issues:
 - the one line change to the `Dockerfile` of `aissemble-fastapi` to use `python:3.11`
 - a change to the builder image used by Habushu to `python:3.11`
   - In order to support using `3.11-slim` for the final image, a new version of Habushu was released an pulled in (2.16.1)

However, while diagnosing this I found some oddities that made this hard to track down:

1. `docker buildx prune` only prunes the default builder, the same seems to be true of `system prune`.
2. changing the config directory of docker via `--config` changes what builders are visible, and what cache space they are using
3. by default, the `docker-maven-plugin` creates a builder named `maven` with a config directory pointing inside the `target` of the project 

Because of this, the `aissemble-versioning` module seemed unaware of changes to the `aissemble-fastapi` base image. This both caused the issue found by CI to be missed locally, and caused difficulty with verifying that the change fixed the issue, as `aissemble-versioning` was always using the remote image.

To address this issue (and hopefully also improve build speed as [noted by fabric8](https://dmp.fabric8.io/#build-buildx)), I've set the `dockerStateDir` to re-use the default docker config location of `~/.docker`.  In the rare event that this is _not_ the default config location, we still get the benefit of sharing a cache across aissemble image builds.

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -28,7 +28,7 @@
         <version.failsafe.plugin>${version.maven.surefire.plugin}</version.failsafe.plugin>
         <version.fermenter>2.10.3</version.fermenter>
         <version.fermenter.legacy.tools>2.8.0</version.fermenter.legacy.tools>
-        <version.habushu.plugin>2.16.0</version.habushu.plugin>
+        <version.habushu.plugin>2.16.1</version.habushu.plugin>
         <version.python>3.11.4</version.python>
         <version.help.plugin>3.2.0</version.help.plugin>
         <version.krausening>19</version.krausening>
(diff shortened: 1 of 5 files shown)


build-parent/pom.xml
```diff
@@ -550,32 +550,29 @@
                     <artifactId>docker-maven-plugin</artifactId>
                     <version>${version.fabric8.docker.maven.plugin}</version>
                     <extensions>true</extensions>
-                    <executions>
-                        <execution>
-                            <id>default-build</id>
```
ewilkins-csi, line 555: I: If you hide whitespace changes this becomes more clear, but I essentially just moved the config out of the specific configuration to the main plugin config so that we don't have to duplicate the config for the `default-push` execution.

build-parent/pom.xml
```diff
@@ -550,32 +550,29 @@
...
-                        </execution>
-                    </executions>
+                    <configuration>
+                        <images>
+                            <image>
+                                <name>${docker.baseline.repo.id}/${dockerImageName}:${dockerImageVersion}</name>
+                                <build>
+                                    <buildx>
+                                        <!-- use default cache, passed to docker via config flag.-->
+                                        <dockerStateDir>~/.docker</dockerStateDir>
```
ewilkins-csi, line 560: I: Shares docker cache between all builds per Fabric8 docs

extensions/extensions-docker/pom.xml
```diff
@@ -124,7 +86,8 @@
                             <phase>prepare-package</phase>
                             <configuration>
                                 <dockerfile>${project.basedir}/src/main/resources/docker/Dockerfile</dockerfile>
-                                <dockerBase>python:3.11-slim</dockerBase>
+                                <dockerBuilderBase>python:3.11</dockerBuilderBase>
+                                <dockerFinalBase>python:3.11-slim</dockerFinalBase>
```
ewilkins-csi, line 90: I: Changes from the new Habushu release allowing us to use the full `3.11` image to build the venv (by having things like `gcc`) while moving the venv onto a slimmer image for the final output.

carter-cundiff (approved):

csun-cpointe (approved):

Issue #199, opened 2024-07-10
Feature: As a user, I want aiSSEMBLE 1.8.0 released, so I have access to the latest features. (Preparation)


csun-cpointe:
## Description
We are ready to cut the release for 1.8.0. We are still in process of getting GitHub actions ready for the release so will continue to use the old infrastructure for this release.

## DOD
Acceptance criteria required to realize the requested feature
Before release:
- [x] Apply the bug patch to fix the environment variable substitutions issues
   - [x] also check and fix other Dockerfile if it has the same issues
- [x] Fix issues blocking release
  - [x] Fabric8 Dockerfile substitution creating broken images (#201)
  - [x] Remove snapshot dependencies (#204)
  - [x] Setup mirror for Fabric8 (#208 - _TODO: Create post_)
  - [x] Fix PDP bugs from rename effort (#218)
  - [x] Apply best practices to Pipeline Invocation chart (#220)
  - [x] Shrink/stream-line Quarkus images (#225, #237)
  - [x] Flatten/stream-line FastAPI images (#225)
  - [x] Remove inadvertent Vault container creation outside of ITs (#239 )

Follow-on: #238

## Test Strategy/Script
1. Validate that the latest CI build has completed successfully
2. Run the policy decision point image: `docker run --pull=always --rm -p8080:8080 ghcr.io/boozallen/aissemble-policy-decision-point:1.8.0-SNAPSHOT`
3. Verify the service is running: `curl localhost:8080/api/healthcheck`
4. Submit an auth request and verify `NOT_APPLICABLE` is returned
```
curl -X POST -H "Content-Type: application/json" localhost:8080/api/pdp -d '{
[...]

PR #204, opened 2024-07-11
#199 fix baton version to use release v1.1.0

csun-cpointe:


cpointe-ibllanos (approved):
Confirmed the artifact is accessible in a local build. LGTM

Issue #190, opened 2024-07-05
Feature: Universal config: Read from encrypt-able sources


cwoods-cpointe:
[190-helper.zip](https://github.com/user-attachments/files/16182868/190-helper.zip)
## Description
Users of the universal config service will need to be able to encrypt config values for security reasons. For example, postgres username/password or S3 keys can not be stored as plain text. The current implementation of reading configs from yaml files does not allow for encrypting values.

Could update the existing yaml solution to read/parse encrypted values but Krausening already does this and we get some other benefits like: file watching, yaml parser maintenance, downstream projects are already familiar with Krausening, new features to Krausening can be shared to all users and not just the config-store-service.

## DOD
- [x] Update config-store-service to load Krausening property files
- [x] get the values from Krausening and store in config-store
- [x] Leverage Krausenings encryption
- [x] Update the helm chart and mda generated values accordingly
- [x] Update the charts documentation

## Implementation details
- [x] Group name will be derived differently
  - [x] This can come from file name
- [x] Ensure this encryption works with the RegenerationRule
  - [x] Shouldnt be an issue because the regeneration doesnt update the config files, only the vault value
- [x] krausening will need to have a KRAUSENING_PASSWORD environment var. How we handle this is still up in the air and we may not want to have another env var that the users have to set. Confirm if setting this env var is a no go

## Test Strategy/Script
- Create a downstream project with:
```
mvn archetype:generate -U -DarchetypeGroupId=com.boozallen.aissemble \
  -DarchetypeArtifactId=foundation-archetype \
  -DarchetypeVersion=1.8.0-SNAPSHOT \
  -DgroupId=com.boozallen.aissemble.test \
  -DartifactId=test-issue-config-store \
  -DprojectGitUrl=git.example.com/test-issue-config-store \
  -DprojectName="Issue Test"
```
- Copy the src directory from the attached folder and put it into the root of the projects
- Add to the fermenter-mda plugin executions in test-issue-config-store-deploy/pom.xml
```
<execution>
    <id>configuration-store</id>
    <phase>generate-sources</phase>
    <goals>
        <goal>generate-sources</goal>
    </goals>
    <configuration>
        <basePackage>com.boozallen.aissemble.test</basePackage>
        <profile>configuration-store-deploy-v2</profile>
        <!-- The property variables below are passed to the Generation Context and utilized
             to customize the deployment artifacts. -->
        <propertyVariables>
            <appName>configuration-store</appName>
        </propertyVariables>
[...]

PR #205, opened 2024-07-11
 #190 Read universal config properties using krausening

cwoods-cpointe:
Enables the use of Krausening encryption. This means the yaml parser is no longer needed as well as the duplication check and misformatting check because they are handled by krausening. The previously defined groupname in the yaml configs will be derived from the filename.


foundation/foundation-configuration-store/src/test/resources/specifications/load-configurations.feature
```diff
@@ -34,16 +34,6 @@ Feature: Load configurations at specified URI based on environment context
     And consumes the base properties
     And augments the base with the environment properties
 
-  Scenario: The properties are misformatted
```
cwoods-cpointe, line 37: i: These 2 scenarios are handled by Krausening

docs/modules/ROOT/pages/guides/guides-configuration-store.adoc
```diff
@@ -3,16 +3,17 @@
 == Overview
 The Configuration Store is a tool that enables the various configurations for a project to be centrally defined and
 managed, while also standardizing access to the configurations. The Configuration Store dynamically provides
-environment-specific configurations, based on the project's development lifecycle phase. Usage limitations and
-regeneration strategies can also be set, allowing configurations to automatically refreshed, thereby bolstering
-security of sensitive properties.
```
d-ryan-ashcraft, line 8: S (Substanative): I assume this is just the _first_ of the options, correct?  A Vault implementation, for instance, may well make sense for larger projects and/or test/prod use.  Should we segment this guide out a bit to account for that?

foundation/foundation-configuration-store/src/test/resources/configurations/example-env/aws-credentials.properties
```diff
@@ -0,0 +1,11 @@
+###
+# #%L
+# aiSSEMBLE::Foundation::Configuration::Store
+# %%
+# Copyright (C) 2021 Booz Allen
+# %%
+# This software package is licensed under the Booz Allen Public License. All Rights Reserved.
+# #L%
+###
```
d-ryan-ashcraft, line 9: S: probably worth a comment specifically noting that this is a test-only value for when it inevitably gets flagged by a secrets scanner.  Just helps folks maintaining or looking have a concise understanding (granted, it should be pretty contextual already since this is in the test folder...)

d-ryan-ashcraft (approved):
Looks good overall - couple of minor comments inline.

docs/modules/ROOT/pages/guides/guides-configuration-store.adoc
cwoods-cpointe: Sorry not sure if I follow. This overview section doesn't talk about implementations specifics like vault vs in memory.  I will add a paragraph bellow talking about the current options.

d-ryan-ashcraft: Yeah, that's what I'm getting at.  Since this in interface implementation, we probably want to radiate that concept into the docs.

Issue #194, opened 2024-07-09
Feature: Migrate aiops references in extensions-authzforce projects


habibimoiz:
## Description
As part of the OSS efforts in aiSSEMBLE, we need to remove all references of AIOPS from the `extensions-authzforce` module.

## DOD

- [x] Rename the extensions-authzforce package --> replace "aiops" with "aissemble".
   - `aissemble/extensions/extensions-security/extensions-authzforce`
- [x] Rename all the child directories, classes, and methods --> replace "`aiops`" with "`aissemble`".
- [x] Rename any references throughout aiSSEMBLE (i.e. adoc files) --> replace "`aiops`" with "`aissemble`".
- [x] Correct all the imports inside of the velocity templates to ensure a successful build of the Extensions Security module.
- [x] Ensure a complete build runs when running build with no cache.
   - Run: `mvn clean install -Dmaven.build.cache.skipCache=true`
- [x] Write a baton migration script to migrate older version to new changes from this ticket:
   -  Migrate any packages, classes, and or methods.
- [x] Run `archetype-test` profile against the extensions-authzforce module to ensure that a downstream project is successfully able to build.
- [x] Run `-Pintegration-test` profile against the `extensions-authzforce` module to ensure the `extensions-authzforce` test does not fail any integration tests in CI.
   - `Run: mvn clean install -Pintegration-test`

## Test Strategy/Script

- Test 1: Building aiSSEMBLE with changes applied to the extensions-authzforce module.
   -  Run without maven build cache enabled.
- Test 2: Visual check on the removal of "AIOPS" text.
- Test 3: Build a downstream project with that uses extensions-authzforce (Foundation Data Access uses authzforce).
   -  This will be done on 1.8.0-SNAPSHOT.
- Test 4: Verify the Baton migration is functioning properly:
   -  The project that will be generated will be on 1.7.0 and then upgraded to 1.8.0-SNAPSHOT.
   -  Provide a zipped version of the 1.7.0 to speed the testing.

## References/Additional Context
_n/a_

habibimoiz:
The original scope of this ticket was to update `extensions-authzforce` however, the remaining modules from `extensions-security` will also be included due to dependencies on one another.

PR #210, opened 2024-07-16
#194 :truck: update modules to conform to aissemble naming conventions

habibimoiz:
#194 :truck:  renaming authzforce modules from aiops to aissemble
 - refactored extensions-authzforce
 - refactored extensions-authzforce-attribute
 - extensions-policy-decision-point-service
 - updated velocity templates for generated files.


extensions/extensions-security/extensions-authzforce/src/main/java/com/boozallen/aissemble/security/authorization/policy/AissembleAttributeUtils.java
```diff
@@ -10,7 +10,6 @@
  * #L%
  */
 
-import com.boozallen.aiops.security.authorization.policy.AiopsAttribute;
```
aaron-gary, line 13: Q: Was this replaced by a corresponding AissembleAttribute, or is is no longer needed?

aaron-gary (approved):

extensions/extensions-security/extensions-authzforce/src/main/java/com/boozallen/aissemble/security/authorization/policy/AissembleAttributeUtils.java
habibimoiz: The import is no longer needed which is why it was removed.

ewilkins-csi (changes requested):
C: Are we not writing migrations for these?  If we were planning to do that in a follow-on, then we shouldn't merge this to dev

habibimoiz:
> C: Are we not writing migrations for these? If we were planning to do that in a follow-on, then we shouldn't merge this to dev

During my testing, what I found was all of the files that used authzforce were generated classes therefore not requiring any migrations. I tested with both Spark & PySpark implemented projects.

habibimoiz:
Closing pull request. Merging changes in a feature branch instead ([feature-authz-rename](https://github.com/boozallen/aissemble/tree/feature-authz-rename )) to accommodate for the migrations required for #194

PR #218, opened 2024-07-18
#199 fix PDP service so docker changes can be tested

ewilkins-csi:
- Deletes duplicative, unused `foundation-policy-decision-point` module
 - Repackages `extensions-pdp-service` to be more in-line with rest of the project
 - Undo aiops -> aissemble property file rename as the code hasn't been updated
 - Update JKS key alias to from aiops to aissemble
 - Add healthcheck URL to PDP service for easy testing


extensions/extensions-docker/aissemble-policy-decision-point/src/main/resources/docker/Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN mkdir -p /deployments/krausening/
 COPY target/dockerbuild/*.jar /deployments/
 
 COPY ./src/main/resources/truststore/aissemble-secure.jks /deployments/
-COPY ./src/main/resources/krausening/base/aissemble-security.properties /deployments/krausening/
+COPY ./src/main/resources/krausening/base/aiops-security.properties /deployments/krausening/
```
ewilkins-csi, line 15: I: This rename is the primary fix for PDP.  Because the code that reads this file hasn't been updated to look for the new name yet, it was failing to find this config.

foundation/foundation-security/pom.xml
```diff
@@ -15,7 +15,6 @@
 	<description>Contains the modules used for security</description>
 
 	<modules>
-		<module>foundation-policy-decision-point</module>
```
ewilkins-csi, line 18: I: Other than docs/javadocs, the classes/files in this module were identical to `extensions-policy-decision-point-service`. I believe these ended up as duplicates during the move from isolated repos to the monorepo. (Two separate repos had the same classes and got pulled into different places in the monorepo.)

carter-cundiff (approved):

aaron-gary (approved):

Issue #222, opened 2024-07-23
BUG: `aissemble-hive-service` docker container fails to build


d-ryan-ashcraft:
## Description
Beginning today, several team members are experiencing the following build problem with aissemble:
```
[INFO] DOCKER> Dockerfile:18
[INFO] DOCKER> --------------------
[INFO] DOCKER> 17 |  
[INFO] DOCKER> 18 | >>> RUN groupadd -r hive --gid=1000 && \
[INFO] DOCKER> 19 | >>>  useradd --home $HIVE_HOME -g hive --shell /usr/sbin/nologin --uid 1000 hive && \
[INFO] DOCKER> 20 | >>>  chown hive:hive -R $HIVE_HOME && \
[INFO] DOCKER> 21 | >>>  ln -s $JAVA_HOME /opt/jre
[INFO] DOCKER> 22 |  
[INFO] DOCKER> --------------------
[INFO] DOCKER> ERROR: failed to solve: process "/bin/sh -c groupadd -r hive --gid=1000 &&  useradd --home $HIVE_HOME -g hive --shell /usr/sbin/nologin --uid 1000 hive &&  chown hive:hive -R $HIVE_HOME &&  ln -s $JAVA_HOME /opt/jre" did not complete successfully: exit code: 4
[ERROR] DOCKER> Error status (1) when building
```

The base image contains the user and uid already.
## Steps to Reproduce
1. Run `mvn clean install -pl :aissemble-hive-server -amd`

## Test Steps
Run the above command and ensure it completes successfully

PR #223, opened 2024-07-23
#222 :wrench: update command to not fail on existing user or group

d-ryan-ashcraft:


extensions/extensions-docker/aissemble-hive/aissemble-hive-service/src/main/resources/docker/Dockerfile
```diff
@@ -1,7 +1,7 @@
 ARG METASTORE_VERSION
-FROM apache/hive:${METASTORE_VERSION} as appsource
```
d-ryan-ashcraft, line 2: I (Informational): cleaning up these case warnings while here...

extensions/extensions-docker/aissemble-hive/aissemble-hive-service/src/main/resources/docker/Dockerfile
```diff
@@ -15,8 +15,8 @@ ENV HIVE_HOME=/opt/hive
 COPY --from=appsource /opt/hadoop $HADOOP_HOME
 COPY --from=appsource /opt/hive $HIVE_HOME
```
d-ryan-ashcraft, line 17: I: adding force flags to not throw an error on the existing group or user.  These have existed for some time, so I suspect this is an upstream Linux flavor update that made the underlying commands more sensitive.

ewilkins-csi (approved):

Issue #161, opened 2024-06-21
Feature: Update to Pydantic 2.7.x


d-ryan-ashcraft:
## Description
aissemble is on the legacy version pydantic (`1.10.x`).  We should move up to `2.7.x` to gain new features and improvements that are not being added to the legacy codebase at this time.  

This is especially doable since there is an incremental approach we can use:

> Pydantic V1.10 vs. V2
> Pydantic V2 is a ground-up rewrite that offers many new features, performance improvements, and some breaking changes
> compared to Pydantic V1.
> 
> If you're using Pydantic V1 you may want to look at the [pydantic V1.10 Documentation](https://docs.pydantic.dev/) or, [1.10.X
> fixes git branch](https://github.com/pydantic/pydantic/tree/1.10.X-fixes). Pydantic V2 also ships with the latest version of
> Pydantic V1 built in so that you can incrementally upgrade your code base and projects: from pydantic import v1 as pydantic_v1.

## Definition of Done (DOD):
Update pydantic 1.10.x to 2.8.x or above. 
Fix any breaking changes due to upgrade of pydantic to new versions ([migration guide](https://docs.pydantic.dev/latest/migration/) is available to help with fixing version changes) 
Ensure complete aiSSEMBLE build is succeed. 
Run integration test.
Run archtype test.

## Testing Steps:

1. Make sure poetry correctly points to pydantic latest version

- Git into feature branch.
- cd into one of modules listed in the comment
- Ex)
`cd <your path to github>/aissemble/Aissemble-extension-model-training-api-sagemaker`
- run this command to verify pydantic is 2.8.x or above 
`poetry show`
2. Make sure aiSSEMBLE can be built without any issues.
- use `mvn clean instal`l to verify build success
- Run Integrated test of aiSSEMBLE using following command 
`mvn clean install -P integration-test`
- Run Archetype test of aiSSEMBLE using following command: 
`mvn clean install -P archetype-test`
3. build downstream project one with pyspark implementaion and ml pipeline and test 1.7 to 1.8 to verify there is no breaking change.
- build branch of aissemble 
`mvn clean install`
- create downstream project using 1.7.0 of aiSSEMBLE artifact id, giturl and project can be any name of your choosing for test. 
```
mvn archetype:generate -U -DarchetypeGroupId=com.boozallen.aissemble \
  -DarchetypeArtifactId=foundation-archetype \
  -DarchetypeVersion=1.7.0\
[...]

PR #224, opened 2024-07-23
#161 Update to pydantic 2.8.x

jaebchoi:
aiSSEMBLE is currently on legacy version of pydantic ( 1.10.x). In order for developer to make use of new features and improvements of pydantic, we would like to bump pydantic to 2.8.x or above.

**Github Issue:**
https://github.com/boozallen/aissemble/issues/161


foundation/foundation-drift-detection/aissemble-foundation-drift-detection-client-python/src/drift/detection/rest/client/drift_rest_client.py
```diff
@@ -68,10 +68,10 @@ def invoke_drift(self, policy_identifier, input, control):
         query_params = {"policyIdentifier": policy_identifier}
         data_headers = {"Content-type": "application/json"}
         # uncomment below for debugging purposes
-        # print('JSON being sent to drift invocation service: ' + drift_data_input.json())
+        # print('JSON being sent to drift invocation service: ' + drift_data_input.json(serialize_as_any=True))
         return requests.post(
             hostname + "/invoke-drift",
             params=query_params,
-            data=drift_data_input.json(),
+            data=drift_data_input.model_dump_json(serialize_as_any=True),
```
jaebchoi, line 75: I(Informational): This `serialize_as_any` setting `true` will dump base class and any child classes. 
This was defaulted to true in V1. 
In V2 it became defaulted to false in which made json to dump only base classes.
We need this to true since we need DriftVariables and DriftVariable class that is child classes of DriftData.

foundation/foundation-mda/src/main/resources/templates/inference/inference.api.grpc.base.py.vm
```diff
@@ -1,5 +1,5 @@
 from google.protobuf import json_format
-from pydantic.error_wrappers import ValidationError
+from pydantic import ValidationError
```
jaebchoi, line 2: I(informational): in V2, pydantic.error_wrappers packages are moved to pydantic.

foundation/foundation-drift-detection/aissemble-foundation-drift-detection-client-python/src/drift/detection/rest/client/drift_rest_client.py
```diff
@@ -68,10 +68,10 @@ def invoke_drift(self, policy_identifier, input, control):
         query_params = {"policyIdentifier": policy_identifier}
         data_headers = {"Content-type": "application/json"}
         # uncomment below for debugging purposes
-        # print('JSON being sent to drift invocation service: ' + drift_data_input.json())
+        # print('JSON being sent to drift invocation service: ' + drift_data_input.json(serialize_as_any=True))
```
habibimoiz, line 71: This might be the way its appearing in the GH UI, but in the off chance that its not, can you revert this such that it one line or adjust it so the indentations are in line with the rest of the code.

extensions/aissemble-extensions-model-training-api-sagemaker/pyproject.toml
```diff
@@ -12,7 +12,7 @@ packages = [
 python = ">=3.11.4, <3.12"
 fastapi = ">=0.95.0"
 uvicorn = {version = "^0.18.0", extras = ["standard"]}
-pydantic = ">=1.8.0,<2.0.0"
+pydantic = "^2.8.0"
 kubernetes = ">=26.1.0"
```
meliz19, line 16: In general, if we use `>=` we should see some better performance. When we use `^`, it limits the number of potential matches greatly. This is a good discussion Ryan posted a while back on the upper limit notation: https://iscinumpy.dev/post/bound-version-constraints/#tldr

foundation/foundation-drift-detection/aissemble-foundation-drift-detection-client-python/src/drift/detection/rest/client/drift_rest_client.py
ewilkins-csi: We should really just remove this or use an actual logger and output to debug and uncomment it.

foundation/foundation-drift-detection/aissemble-foundation-drift-detection-client-python/src/drift/detection/rest/client/drift_rest_client.py
ewilkins-csi: S: I'd prefer to use the [recommended method](https://docs.pydantic.dev/latest/concepts/serialization/#overriding-the-return-type-when-dumping-a-model) of specifying the actual schema if that's possible instead of relying on duck typing.  The change to the default was to make the behavior align better with production standards it seems.

foundation/foundation-drift-detection/aissemble-foundation-drift-detection-client-python/src/drift/detection/rest/client/drift_rest_client.py
jaebchoi: +1 on Emily. 
I feel like we can just remove it and have devs just add this in their testing branch if they are debugging as I don't see people put debugging log unless anyone have strong opinion to keep it.

ewilkins-csi:
Q/A: The following comments are in a ton of different pyproject.toml files and one template file. Is it still true now that we are on v2?
> `# Adding the pydantic package to pylint's whitelist as it raises linting errors. `
> `# See issue: https://github.com/pydantic/pydantic/issues/1961`

Q/S: Did we not need to migrate any of the pydantic usage in the inference templates at `foundation/foundation-mda/src/main/resources/templates/general-mlflow`?

jaebchoi:
> Q/A: The following comments are in a ton of different pyproject.toml files and one template file. Is it still true now that we are on v2?
> 
> > `# Adding the pydantic package to pylint's whitelist as it raises linting errors. `
> > `# See issue: https://github.com/pydantic/pydantic/issues/1961`
> 
> Q/S: Did we not need to migrate any of the pydantic usage in the inference templates at `foundation/foundation-mda/src/main/resources/templates/general-mlflow`?
> 
> A: Since there are breaking changes, I'd also add this to the draft release notes.

pylint whitelist issue is now resolved in Pydantic V2.
I have Verified with Pydantic V2 version that building mvn clean install without pylint whitelist is succeded. 

Also cross checked that service will get build failure if we omit whitelist under Pydantic V1 just in case.

For mentioned pydantic usage in inference, I inspected and syntax looks legit and only issue I found was package issue in which I have resolved foundation/foundation-mda/src/main/resources/templates/inference/inference.api.grpc.base.py.vm

Also Added Draft Release note please let me know if this needs changed!

PR #233, opened 2024-07-25
#161 Update to pydantic 2.8.x

jaebchoi:
aiSSEMBLE is currently on legacy version of pydantic ( 1.10.x). In order for developer to make use of new features and improvements of pydantic, we would like to bump pydantic to 2.8.x or above.

Github Issue:
https://github.com/boozallen/aissemble/issues/161


foundation/foundation-drift-detection/aissemble-foundation-drift-detection-client-python/src/drift/detection/rest/client/drift_rest_client.py
```diff
@@ -67,11 +67,9 @@ def invoke_drift(self, policy_identifier, input, control):
         drift_data_input = DriftDataInput(input=input, control=control)
         query_params = {"policyIdentifier": policy_identifier}
         data_headers = {"Content-type": "application/json"}
-        # uncomment below for debugging purposes
-        # print('JSON being sent to drift invocation service: ' + drift_data_input.json())
         return requests.post(
             hostname + "/invoke-drift",
             params=query_params,
-            data=drift_data_input.json(),
+            data=drift_data_input.model_dump_json(serialize_as_any=True),
```
jaebchoi, line 73: @ewilkins-csi Hi! I have been investigating on specifying schema but seems like might not be doable given that one of class have recursive call of object. I also tried consolidating all three classes into one and use model_serializer but no luck there too =(
Looks like they made it to default to prevent any security risk of exposing any PIIs or sensitive information like password. But given that our subclasses doesn't seem to contain any of that, I would say we can keep duck typing. see [here](https://docs.pydantic.dev/latest/concepts/serialization/#serializing-with-duck-typing)
If not, I am open for any implementation ideas!

jaebchoi:
Q/A: The following comments are in a ton of different pyproject.toml files and one template file. Is it still true now that we are on v2?

# Adding the pydantic package to pylint's whitelist as it raises linting errors. 
# See issue: https://github.com/pydantic/pydantic/issues/1961

Q/S: Did we not need to migrate any of the pydantic usage in the inference templates at foundation/foundation-mda/src/main/resources/templates/general-mlflow?

A: Since there are breaking changes, I'd also add this to the draft release notes.

pylint whitelist issue is now resolved in Pydantic V2.
I have Verified with Pydantic V2 version that building mvn clean install without pylint whitelist is succeded.

Also cross checked that service will get build failure if we omit whitelist under Pydantic V1 just in case.

For mentioned pydantic usage in inference template, I inspected and syntax looks legit and only issue I found was package issue in which I have resolved foundation/foundation-mda/src/main/resources/templates/inference/inference.api.grpc.base.py.vm

Also Added Draft Release note please let me know if this needs changed!

foundation/foundation-drift-detection/aissemble-foundation-drift-detection-client-python/src/drift/detection/rest/client/drift_rest_client.py
ewilkins-csi: Ok. Thanks for giving it a shot! Since this is just a client and not intended to really be extended, I think it's fair to say our usage is generally safe from the security concern.

ewilkins-csi (approved):
Your thoroughness in resolving feedback is very much appreciated. Great work!

PR #237, opened 2024-07-25
#199 fix Krausening/Quarkus config through helm

ewilkins-csi:
Prior to transitioning the Quarkus-based Docker images off of the `aissemble-quarkus` image, we were setting JAVA_OPTIONS at startup via the `run-env.sh` script to include the Krausinening configuration through environment variables.  When it was moved, the script was dropped and the ENV vars were set directly during the image build. However, this makes the config unresponsive to the `env` config for Helm deployments which removes the ability to point Krausening to a different location at deploy time. For now, we'll just revert to using `aissemble-quarkus` and the `run-env.sh` script there.  Further work on Universal Config and a more "serverless" style of deployment will likely end up illuminating a better answer in the future.

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -1007,9 +1007,6 @@ To suppress enforce-helm-version rule, you must add following plugin to the root
 
         <profile>
             <id>ci</id>
-            <properties>
-                <docker.repo.id>ghcr.io</docker.repo.id>
-            </properties>
             <build>
                 <pluginManagement>
                     <plugins>
@@ -1057,7 +1054,6 @@ To suppress enforce-helm-version rule, you must add following plugin to the root
                 </pluginManagement>
             </build>
         </profile>
-
         <profile>
             <id>integration-test</id>
             <build>
--- a/extensions/extensions-docker/aissemble-quarkus/src/main/resources/docker/Dockerfile
+++ b/extensions/extensions-docker/aissemble-quarkus/src/main/resources/docker/Dockerfile
@@ -1,36 +1,14 @@
 # Script for creating base Quarkus application Docker image
-
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.3 
+FROM registry.access.redhat.com/ubi9/openjdk-11-runtime:1.20
 
 LABEL org.opencontainers.image.source="https://github.com/boozallen/aissemble"
 
-ARG JAVA_PACKAGE=java-11-openjdk-headless
-ARG RUN_JAVA_VERSION=1.3.8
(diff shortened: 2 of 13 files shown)


build-parent/pom.xml
```diff
@@ -1007,9 +1007,6 @@ To suppress enforce-helm-version rule, you must add following plugin to the root
 
         <profile>
             <id>ci</id>
-            <properties>
-                <docker.repo.id>ghcr.io</docker.repo.id>
```
ewilkins-csi, line 1011: I: Setting this here will inadvertently cause downstream projects to build against `ghcr.io` when they run the `ci` profile. Instead we should and already have configured this value in `extensions/extensions-docker/pom.xml`.

extensions/extensions-docker/aissemble-quarkus/src/main/resources/docker/Dockerfile
```diff
@@ -1,36 +1,14 @@
...
-    && chown 1001 /deployments/run-java.sh \
-    && chmod 540 /deployments/run-java.sh \
-    && chown -R 1001:root /deployments \
-    && echo "securerandom.source=file:/dev/urandom" >> /etc/alternatives/jre/conf/security/java.security
-
-# Configure the JAVA_OPTIONS
-ENV JAVA_OPTIONS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager"
-
-EXPOSE 8080
-USER 1001
```
ewilkins-csi, line 31: I: The above work was already determined to be part of the openjdk UBI, or unnecessary.

extensions/extensions-docker/aissemble-quarkus/src/main/resources/scripts/run-env.sh
```diff
@@ -18,21 +18,21 @@ then
     then
       # Default to value that was originally in Dockerfile to maintain backwards compatability.
       echo "Defaulting KRAUSENING_BASE to /deployments/krausening/"
-      export JAVA_OPTIONS="$JAVA_OPTIONS -DKRAUSENING_BASE=/deployments/krausening/"
+      export JAVA_OPTS_APPEND="$JAVA_OPTS_APPEND -DKRAUSENING_BASE=/deployments/krausening/"
```
ewilkins-csi, line 21: I: The `run-java.sh` script in the base image uses this variable to set the java options instead of `JAVA_OPTIONS`. This allows it to have some more advanced features like a debug flag that automatically includes the right options at startup.

cwoods-cpointe (approved):

PR #239, opened 2024-07-26
#199 only start Vault during ITs

ewilkins-csi:
There is one unit test in `aissemble-extensions-encryption-vault-python` and 3 ITs. Because of the way the feature files and `environent.py` setup was structured, the Vault container was spun up even if only the unit test is being executed and the ITs are excluded.  This causes problems during the release as the Vault container is not available during the preparation phase, and generally slows down builds.

### DIFF
--- a/extensions/extensions-encryption/aissemble-extensions-encryption-vault-python/tests/features/encryption.feature
+++ b/extensions/extensions-encryption/aissemble-extensions-encryption-vault-python/tests/features/encryption.feature
@@ -1,32 +1,8 @@
 @properties
 Feature: Encryption
 
-  @integration
-  Scenario: Data can be encrypted and decrypted using Vault
-    Given a plain text string
-    When the string is encrypted using Vault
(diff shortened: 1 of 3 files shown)


ewilkins-csi:
I: Consider ignoring whitespace

extensions/extensions-encryption/aissemble-extensions-encryption-vault-python/tests/features/encryption.feature
```diff
@@ -1,32 +1,8 @@
 @properties
 Feature: Encryption
 
-  @integration
-  Scenario: Data can be encrypted and decrypted using Vault
```
ewilkins-csi, line 5: I: Move ITs to a separate feature file, so we can use feature tags to determine whether a Vault container is required.

colinpalmer-pro (approved):

carter-cundiff (approved):

Issue #209, opened 2024-07-15
Feature: As a user of universal configurations, I want to deploy a mutating admissions webhook so that I can ingest my configurations


carter-cundiff:
## Description
A [Kubernetes web hook](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#what-are-admission-webhooks:~:text=monitor%20admission%20webhooks.-,What%20are%20admission%20webhooks%3F,-Admission%20webhooks%20are) is a service that runs in a Kubernetes cluster and monitors the cluster according to a set of predefined rules. These rules are defined within the webhook yaml and can specify which events to catch such as when a pod is created. When these events occur, the web hook is invoked and an AdmissionReview object is passed to our mutate method defined as a quarkus endpoint. The AdmissionReview object contains an AdmissionRequest which holds the YAML data for the pod that was created. This data can be intercepted and updated as needed. The new data is then packaged into the AdmissionResponse which is updated in the Kubernetes object. This can be used to inject environment variables or update other parts of the YAML.

**Questions**:
- Do we want to include the resource modifications as part of this ticket, or just create the webhook?
  - A: Will be handled in follow on ticket

## DOD
- [x] Create a new endpoint within `foundation-configuration-store` to act as the [MutatingAdmissionWebhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook)
  - [x] Investigate priority of multiple webhooks for the same resource - order is not guaranteed or configurable 
  - [x] Endpoint will simply log it was called, follow on ticket will handle injection logic
- [x] Update `aissemble-configuration-store-chart` helm chart to utilize the new webhook
  - [x] Webhook should be called on all resources for CREATE and UPDATE only 
  - [x] Follow on will implement resource selection logic

### BDD Scenarios:
```
Scenario: The configuration service can process Kubernetes resources
```
## Test Strategy/Script
- **OTS Only**:
  - Run the following
    ```
    mvn clean install -pl :foundation-mda,:foundation-configuration-store,:aissemble-quarkus,:aissemble-configuration-store,:aissemble-quarkus-chart,:aissemble-configuration-store-chart -Dmaven.build.cache.skipCache=true
    ```

- Create a downstream project from the archetype using the following command:
```
mvn archetype:generate -U -DarchetypeGroupId=com.boozallen.aissemble \
  -DarchetypeArtifactId=foundation-archetype \
[...]

PR #244, opened 2024-07-29
#209 Create mutating admissions webhook for Universal Configuration Store

carter-cundiff:
- Migrated config store helm chart to use quarkus helm chart as it's parent.
- Updated quarkus helm chart dependents to set a unique config map name for their quarkus configs to avoid naming collisions in downstream deployments.


extensions/extensions-helm/aissemble-quarkus-chart/values.template.yaml
```diff
@@ -20,7 +20,7 @@ deployment:
   volumes:
     - name: quarkus-application-properties
       configMap:
-        name: supplemental-quarkus-config
+        name: "{{ .Values.configMap.name }}"
```
carter-cundiff, line 23: I: Set's the config map name to the value of `.Values.configMap.name` which can then be easily overwritten in downstream helm charts.

extensions/extensions-docker/aissemble-configuration-store/src/main/resources/docker/Dockerfile
```diff
@@ -6,3 +6,17 @@ COPY --chown=default target/quarkus-app/lib/ $JAVA_APP_DIR/lib/
...
+
+# Custom scripts to generate certs for the mutating webhook and create the webhook itself
+COPY ./src/main/resources/scripts/genCerts.sh /usr/bin/
+COPY ./src/main/resources/scripts/genWebhook.sh /usr/bin/
+
+USER root
+
+RUN microdnf install -y openssl
+
+RUN chmod +x /usr/bin/genCerts.sh \
```
ewilkins-csi, line 18: C: These probably need to be generated at startup, so keys aren't shared.

extensions/extensions-docker/aissemble-configuration-store/src/main/resources/docker/Dockerfile
```diff
@@ -6,3 +6,17 @@ COPY --chown=default target/quarkus-app/lib/ $JAVA_APP_DIR/lib/
...
 COPY --chown=default target/quarkus-app/app/ $JAVA_APP_DIR/app/
 COPY --chown=default target/quarkus-app/quarkus/ $JAVA_APP_DIR/quarkus/
+
+# Custom scripts to generate certs for the mutating webhook and create the webhook itself
+COPY ./src/main/resources/scripts/genCerts.sh /usr/bin/
+COPY ./src/main/resources/scripts/genWebhook.sh /usr/bin/
+
+USER root
+
+RUN microdnf install -y openssl
```
ewilkins-csi, line 16: A: I'd move this to the top since it's expensive and not likely to change.

extensions/extensions-docker/aissemble-configuration-store/src/main/resources/docker/Dockerfile
```diff
@@ -6,3 +6,17 @@ COPY --chown=default target/quarkus-app/lib/ $JAVA_APP_DIR/lib/
 COPY --chown=default target/quarkus-app/*.jar $JAVA_APP_DIR/
 COPY --chown=default target/quarkus-app/app/ $JAVA_APP_DIR/app/
 COPY --chown=default target/quarkus-app/quarkus/ $JAVA_APP_DIR/quarkus/
+
+# Custom scripts to generate certs for the mutating webhook and create the webhook itself
+COPY ./src/main/resources/scripts/genCerts.sh /usr/bin/
+COPY ./src/main/resources/scripts/genWebhook.sh /usr/bin/
```
ewilkins-csi, line 12: A: Recommend using `--chown`/`--chmod` instead of a separate `RUN` statement.

extensions/extensions-docker/aissemble-configuration-store/src/main/resources/docker/Dockerfile
ewilkins-csi: I just realized this is only changing the perms not actually running the script.

extensions/extensions-helm/aissemble-configuration-store-chart/templates/webhook-cleanup-job.yaml
```diff
@@ -0,0 +1,46 @@
+{{ if .Values.webhook.enable }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ index .Values "aissemble-quarkus-chart" "app" "name" }}-webhook-cleanup
+  annotations:
+    "helm.sh/hook": pre-delete, pre-upgrade
```
ewilkins-csi, line 7: Q/S: Does this regenerate the cert on upgrade, and if so, is that necessary? I'd think we'd want the cert to stay the same. Is the idea that they webhook values might have changed and affect cert generation?

extensions/extensions-helm/aissemble-configuration-store-chart/templates/webhook-cleanup-job.yaml
```diff
@@ -0,0 +1,46 @@
...
+spec:
+  template:
+    metadata:
+      name: {{ index .Values "aissemble-quarkus-chart" "app" "name" }}-webhook-cleanup
+    spec:
+      serviceAccountName: {{ .Values.webhook.serviceAccount.name }}
+      restartPolicy: OnFailure
+      containers:
+      - name: clean-webhook
+        image: {{ index .Values "aissemble-quarkus-chart" "deployment" "image" "repo" }}{{ index .Values "aissemble-quarkus-chart" "deployment" "image" "name" }}:{{ index .Values "aissemble-quarkus-chart" "deployment" "image" "tag" | default .Chart.AppVersion }}
```
ewilkins-csi, line 18: S: I'd rather use a slimmer image, especially if we're writing the shell script in the template and not executing something on the image. (Could argue that our current generation scripts might work better as a separate, targeted image.)

ewilkins-csi, line 18: A: This file might look a bit tidier by using `with` and/or local variables.  Like these image lines becomes a bit more readable if you wrap it in `{{- with .Values "aissemble-quarkus" "deployment" "image" }}`

extensions/extensions-helm/aissemble-configuration-store-chart/templates/webhook-cleanup-job.yaml
carter-cundiff: I just copied that hook from the spark-operator clean up job. However thinking it through now, I don't think it makes sense for it to run on `pre-upgrade` since we don't need to the certs to change.

extensions/extensions-helm/aissemble-configuration-store-chart/templates/webhook-cleanup-job.yaml
```diff
@@ -0,0 +1,46 @@
...
+          https://kubernetes.default.svc/api/v1/namespaces/{{ .Release.Namespace }}/secrets/{{ .Values.webhook.certSecret.name }}
+        {{- end }}
+          && \
+          curl -ik \
+          -X DELETE \
+          -H \"Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)\" \
+          -H \"Accept: application/json\" \
+          -H \"Content-Type: application/json\" \
+          --data \"{\\\"kind\\\":\\\"DeleteOptions\\\",\\\"apiVersion\\\":\\\"batch/v1\\\",\\\"propagationPolicy\\\":\\\"Foreground\\\"}\" \
+          https://kubernetes.default.svc/apis/batch/v1/namespaces/{{ .Release.Namespace }}/jobs/{{ index .Values "aissemble-quarkus-chart" "app" "name" }}-webhook-init"
```
ewilkins-csi, line 45: Q: Why only delete the init job at delete/upgrade? Couldn't we delete it right after it runs?

extensions/extensions-helm/aissemble-configuration-store-chart/templates/webhook-init-job.yaml
```diff
@@ -0,0 +1,45 @@
...
+            "--secret-name", "{{ .Values.webhook.certSecret.name }}"
+          ]
+      {{- end }}
+      containers:
+      - name: create-webhook
+        image: {{ index .Values "aissemble-quarkus-chart" "deployment" "image" "repo" }}{{ index .Values "aissemble-quarkus-chart" "deployment" "image" "name" }}:{{ index .Values "aissemble-quarkus-chart" "deployment" "image" "tag" | default .Chart.AppVersion }}
+        imagePullPolicy: {{ index .Values "aissemble-quarkus-chart" "deployment" "image" "imagePullPolicy"}}
+        command: ["/usr/bin/genWebhook.sh"]
+        args: [
+            "--ca-bundle", "$(CA_BUNDLE)",
```
ewilkins-csi, line 34: Q: Why create a subshell here? Couldn't we just do `$CA_BUNDLE`?

extensions/extensions-helm/aissemble-configuration-store-chart/values.template.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+  configMap:
+    name: configuration-store-quarkus-config
+    supplementalQuarkusConfig:
+      - quarkus.http.ssl.certificate.file=/etc/webhook/cert/tls.crt
+      - quarkus.http.ssl.certificate.key-file=/etc/webhook/cert/tls.key
+    
+webhook:
+  enable: true
+  name: aissemble-configuration-store.webhook.boozallen # must be at least three segments separated by dots
```
ewilkins-csi, line 54: Q/A: Would `boozallen.aissemble-configuraiton-store.webhook` be more approopriate?

extensions/extensions-helm/aissemble-configuration-store-chart/templates/webhook-serviceaccount.yaml
```diff
@@ -0,0 +1,50 @@
...
+    "helm.sh/hook": pre-install, pre-upgrade
+    "helm.sh/hook-delete-policy": hook-failed, before-hook-creation
+    "helm.sh/hook-weight": "-10"
+
+---
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: config-store-webhook-clusterrole
```
ewilkins-csi, line 16: A: Should we not use the app/release name?

extensions/extensions-helm/aissemble-configuration-store-chart/templates/webhook-init-job.yaml
```diff
@@ -0,0 +1,45 @@
+{{ if .Values.webhook.enable }}
+apiVersion: batch/v1
+kind: Job
```
ewilkins-csi, line 3: Q: These are done as jobs instead of init-containers so that the main deployment can be less privileged?

extensions/extensions-helm/aissemble-configuration-store-chart/templates/webhook-serviceaccount.yaml
```diff
@@ -0,0 +1,50 @@
...
+  annotations:
+    "helm.sh/hook": pre-install, pre-upgrade
+    "helm.sh/hook-delete-policy": hook-failed, before-hook-creation
+    "helm.sh/hook-weight": "-10"
+rules:
+- apiGroups: ["batch"]
+  resources: ["jobs"]
+  verbs: ["delete", "deletecollection"]
+- apiGroups: [""]
+  resources: ["secrets"]
```
ewilkins-csi, line 26: Q/S: Don't we only need secrets access if we're generating the certs?

extensions/extensions-helm/aissemble-data-access-chart/values.yaml
```diff
@@ -9,13 +9,11 @@ aissemble-quarkus-chart:
         protocol: TCP
         targetPort: 8080
   deployment:
-    ports:
-      - name: http-1
-        containerPort: 8081
```
ewilkins-csi, line 14: Q: Unused?

extensions/extensions-helm/aissemble-quarkus-chart/values.template.yaml
ewilkins-csi: So the `tpl` call in the template processes this bit as a template and so expands this to the actual value instead of using the string literally?

extensions/extensions-helm/extensions-helm-pipeline-invocation/aissemble-pipeline-invocation-app-chart/values.yaml
```diff
@@ -24,7 +24,7 @@ aissemble-quarkus-chart:
           name: spark-application-values-files
 
   configMap:
-    name: supplemental-quarkus-config
+    name: pipeline-invocation-quarkus-config
```
ewilkins-csi, line 27: Q: Presumably we don't have to worry about migrations because any down stream project would have had to fix this already or they only have one Quarkus deployment?

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -0,0 +1,63 @@
...
+import javax.ws.rs.Produces;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.core.MediaType;
+
+/**
+ * Provides a quarkus endpoint for our MutatatingWebhook that is created as part of the universal configuration
+ * helm chart. This endpoint is used to modify the incoming kubernetes resource with values from the config store.
+ */
+@Path("")
```
ewilkins-csi, line 33: A: Since this is bundled with the REST API, I'd consider putting this under a `/webhook` path to avoid collisions.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -0,0 +1,63 @@
...
+@ApplicationScoped
+public class ConfigMutatingWebhook {
+    private static final Logger logger = LoggerFactory.getLogger(ConfigMutatingWebhook.class);
+    
+    @POST
+    @Path("/inject")
```
ewilkins-csi, line 39: A: I might lean toward the name `/process` or `/expand` here.

foundation/foundation-configuration-store/src/test/resources/specifications/load-configurations.feature
```diff
@@ -69,3 +69,9 @@ Feature: Load configurations at specified URI based on environment context
     Given there exists a krausening_password and encrypted properties
     When the configuration service starts
     Then the loaded properties contains the decrypted value
+
+  @config-webhook
+  Scenario: The configuration service can process Kubernetes resources
```
ewilkins-csi, line 74: A: I'd consider moving this to it's own feature file since we know we'll have multiple scenarios related to the webhook.

extensions/extensions-docker/aissemble-configuration-store/src/main/resources/scripts/genWebhook.sh
```diff
@@ -0,0 +1,177 @@
...
+
+echo "Creating a MutatingWebhookConfiguration"
+
+TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
+
+TMP_DIR="/tmp/config-store-webhook/"
+mkdir -p ${TMP_DIR}
+
+# Base64 encode and then remove the any line breaks to avoid issues in the curl command
+ca_bundle=$(echo "$CA_BUNDLE"| base64 -w 0| tr -d '\n')
```
ewilkins-csi, line 101: Q: Wouldn't the CA bundle already be b64 encoded?

extensions/extensions-docker/aissemble-configuration-store/src/main/resources/scripts/genWebhook.sh
```diff
@@ -0,0 +1,177 @@
...
+
+case "$STATUS" in
+  201)
+    printf "\nSuccess - mutating webhook created.\n"
+  ;;
+  409)
+    printf "\nSuccess - mutating webhook already exists.\n"
+    ;;
+    *)
+    printf "\nFailed creating mutating webhook.\n"
```
ewilkins-csi, line 170: A: Might be helpful to include the status in the output

extensions/extensions-helm/aissemble-configuration-store-chart/README.md
```diff
@@ -17,22 +17,35 @@ helm install configuration-store ghcr.io/boozallen/aissemble-configuration-store
...
+| Property | Description | Required Override | Default |
+|----------|-------------|-------------------|---------|
+| webhook.enable | Enable custom webhook | No | true |
+| webhook.name | Name of the custom webhook (must be at least three segments separated by dots) | No | aissemble-configuration-store.webhook.boozallen |
+| webhook.serviceAccount.create | Enable the creation of a service account for managing the webhook | No | true |
+| webhook.serviceAccount.name | Name of the service account used to managing the webhook | Yes* | aissemble-configuration-store-webhook-sa |
+| webhook.certSecret.create | Enable the creation of a secret with necessary certificates for using the webhook | No | true |
+| webhook.certSecret.name | Name of the secret the certificates are stored in | Yes* | aissemble-configuration-store-webhook-certs |
+
+\* When creation of the service account/certificate secret is disabled, the name must be overridden with an existing resource on the cluster to be there replacement.
```
ewilkins-csi, line 51: A: We should specify what keys we expect to exist in the associated secret.

ewilkins-csi (approved):

extensions/extensions-helm/aissemble-configuration-store-chart/templates/webhook-init-job.yaml
carter-cundiff: Copied the approach from spark-operator. I supposed you could do them as init containers on the main deployment too, however, this does have the added benefit of using a separate service account to control the privileges.

extensions/extensions-helm/aissemble-configuration-store-chart/templates/webhook-init-job.yaml
carter-cundiff: Not sure why, but when I do just `$CA_BUNDLE` only the last line of the variable gets passed in. Surrounding it with () ensure the full value gets passed in.

extensions/extensions-docker/aissemble-configuration-store/src/main/resources/scripts/genWebhook.sh
carter-cundiff: Interestingly when it gets passed in from the env it get un-encoded. So we have to re-encode again here.

extensions/extensions-helm/aissemble-data-access-chart/values.yaml
carter-cundiff: Correct. The quarkus chart sets the deployment ports using the `service.ports`, so this attribute it never actually used. Good catch though, the env below should have the name updated to `http`.

extensions/extensions-helm/aissemble-quarkus-chart/values.template.yaml
carter-cundiff: Correct.

extensions/extensions-helm/extensions-helm-pipeline-invocation/aissemble-pipeline-invocation-app-chart/values.yaml
carter-cundiff: Correct

extensions/extensions-helm/aissemble-configuration-store-chart/templates/webhook-cleanup-job.yaml
carter-cundiff: Update the hook to delete the init job on success. Left this curl delete here though so it still deletes the job on helm uninstall if the init job failed.

extensions/extensions-helm/aissemble-configuration-store-chart/templates/webhook-serviceaccount.yaml
carter-cundiff: Updated to make secret permissions conditional.

extensions/extensions-helm/aissemble-configuration-store-chart/templates/webhook-cleanup-job.yaml
carter-cundiff: Updated to use `curlimages/curl:latest`

carter-cundiff: Refactored cleanup and init to use helm variables

extensions/extensions-helm/aissemble-configuration-store-chart/values.template.yaml
carter-cundiff: Updated to `boozallen.aissemble-configuraiton-store.webhook`

foundation/foundation-configuration-store/src/test/resources/specifications/load-configurations.feature
carter-cundiff: Moved to it's own feature file.

PR #250, opened 2024-07-31
Fix Python service containers

ewilkins-csi:
While there are several confounding factors, the root issue was a release by the `bundle` plugin team on Friday. Release 1.4.0 appears to have a bug so that if a project has an existing virtual environment, the target location is _ignored_.  This means we weren't creating the venv bundle at `/opt/venv` as expected.

We weren't seeing this locally because we use in-project virtual environments that are not copied to CI.  For reasons outside the scope of this PR, CI was using a poetry.toml file that did _not_ use an in-project env thus encountering the bug.  To provide better consistency between CI and local (and between Maven/Habushu and native Poetry) we will commit `poetry.toml`.

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -575,8 +575,8 @@
                                         <build>
                                             <buildx>
                                                 <builderName>maven</builderName>
-                                                <cacheTo>type=local,dest=~/.docker/cache .</cacheTo>
-                                                <cacheFrom>type=local,src=~/.docker/cache .</cacheFrom>
+                                                <cacheTo>type=local,dest=${user.home}/.docker/cache</cacheTo>
+                                                <cacheFrom>type=local,src=${user.home}/.docker/cache</cacheFrom>
                                             </buildx>
                                         </build>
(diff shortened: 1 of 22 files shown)


build-parent/pom.xml
```diff
@@ -575,8 +575,8 @@
                                         <build>
                                             <buildx>
                                                 <builderName>maven</builderName>
-                                                <cacheTo>type=local,dest=~/.docker/cache .</cacheTo>
-                                                <cacheFrom>type=local,src=~/.docker/cache .</cacheFrom>
+                                                <cacheTo>type=local,dest=${user.home}/.docker/cache</cacheTo>
```
ewilkins-csi, line 578: I: Currently on CI, the syntax above was creating a _literal_ directory named `~` instead of dereferencing to the home directory of the user. `$HOME` had a similar issue, so using the Java system property instead.  That way the expansion happens before the path is even passed to the fabric8 plugin.

carter-cundiff (approved):

Issue #251, opened 2024-07-31
Feature: Switch to official Spark Operator release


ewilkins-csi:
## Description
There was a bug in Spark Operator that we had to manually fix with a custom fork/compilation of Spark Operator.  That change bug fix was finally [merged in](https://github.com/kubeflow/spark-operator/pull/1979) on May 17.  We should upgrade to the version with our fix and remove the custom image if possible.  It might also make sense to jump to the latest version of the Spark Operator chart while we're at it.

## DOD
- [x] aiSSEMBLE no longer clones and compiles the [Tech Brewery fork](https://github.com/TechnologyBrewery/spark-on-k8s-operator) of Spark Operator
- [x] Dynamic allocation of executors still works as expected

## Test Script
1.  Generate a new project
```
mvn archetype:generate '-DarchetypeGroupId=com.boozallen.aissemble' \
                       '-DarchetypeArtifactId=foundation-archetype' \
                       '-DarchetypeVersion=1.8.0-SNAPSHOT' \
                       '-DartifactId=test-251' \
                       '-DgroupId=org.test' \
                       '-Dpackage=org.test' \
                       '-DprojectGitUrl=test.org/test-251.git' \
                       '-DprojectName=Test Project' -B && cd test-251
```
2. Add the attached [**PysparkPipeline.json**](https://github.com/user-attachments/files/16457532/PysparkPipeline.json) to _test-251-pipeline-models/src/main/resources/pipelines_
3. Build the project and follow all manual actions, repeating until no more actions are printed
```
mvn clean install
```
4. Do a full build of the project and complete any new manual actions
```
mvn clean install -Dmaven.build.cache.skipCache
```
5.  **OTS Only** Switch the spark-operator chart to use the locally modified chart
    1. Add `upgrade-v2-chart-files-aissemble-version-migration` to the disabled migrations in the root _pom.xml_
    2. Update _test-251-deploy/src/main/resources/apps/spark-operator/Chart.yaml_ to point to the local chart version 1.0.0
    3. Update _test-251-deploy/src/main/resources/apps/spark-operator/values.yaml_ to set `aissemble-spark-operator-chart.spark-operator.image.pullPolicy` to **Never**
    4. Rebuild the deploy module: `mvn clean install -pl :test-251-deploy`
6. Deploy the project
```
tilt up
```
7. Execute the _pyspark-pipeline_ resource in Tilt
8. Verify the pipeline completes successfully with no exceptions from the `CoarseGrainedExecutor` class

[...]

PR #253, opened 2024-08-01
#251 remove custom spark operator compilation

ewilkins-csi:
- remove the custom compilation of Spark Operator
 - update the Spark Operator chart dependency to the latest version
 - modify the Spark Operator chart to work with the stock kubeflow image
   - remove custom RBAC
   - change ivy cache mount location

Note that the only thing preventing the complete removal of our custom Spark Operator image is the fact that kubeflow's Spark Operator is on Spark 3.5 while we're on Spark 3.4.  The upgrade is actually pretty easy without many breaking changes at all, however it started to require some dependency untangling with different versions being pulled in between Spark 3.5 and Quarkus 2.8.  Since we know we'll be upgrading Quarkus soon, I'm leaving the Spark 3.5 upgrade to follow that.  This should cut a significant amount of time off of the image build though, as the custom Go compilation was taking at least half of the build time.


build-parent/pom.xml
```diff
@@ -31,7 +31,6 @@
         <version.assembly.plugin>3.6.0</version.assembly.plugin>
         <version.buildhelper.plugin>3.2.0</version.buildhelper.plugin>
         <version.buildnumber.plugin>3.1.0</version.buildnumber.plugin>
-        <version.cucumber.reporting.plugin>5.7.5</version.cucumber.reporting.plugin>
```
ewilkins-csi, line 34: I: Just consolidating the two properties into the common parent pom.

extensions/extensions-docker/aissemble-spark-operator/src/main/resources/docker/Dockerfile
```diff
@@ -16,40 +16,20 @@
...
-RUN apt-get update --allow-releaseinfo-change \
-    && apt-get update \
-    && apt-get install -y openssl curl tini \
-    && rm -rf /var/lib/apt/lists/* \
-    && chmod +x /usr/bin/entrypoint.sh \
-    && chmod +x /usr/bin/gencerts.sh
+COPY --from=builder --chmod=755 /usr/bin/spark-operator /usr/bin/
+COPY --from=builder --chmod=755 /usr/bin/entrypoint.sh /usr/bin/
+RUN apt-get update \
+    && apt-get install -y tini \
```
ewilkins-csi, line 31: I: Spark Operator 1.6+ no longer uses the gencerts script or the webhook init/cleanup k8s jobs.  So we only need `tini` as it's used in `entrypoint.sh` to launch the service.

extensions/extensions-helm/aissemble-spark-operator-chart/Chart.yaml
```diff
@@ -10,7 +10,7 @@ sources:
   - https://github.com/boozallen/aissemble
 dependencies:
   - name: spark-operator
-    version: 1.1.27
+    version: 1.4.6
```
ewilkins-csi, line 13: I: Latest version as of today

extensions/extensions-helm/aissemble-spark-operator-chart/README.md
```diff
@@ -45,7 +45,6 @@ aissemble-spark-operator-chart:
 | volumes                    | Volumes for the pod                | No                | `spark-logging=/tmp/spark-logging`          |
 | volumeMounts               | Volume Mounts for the pod          | No                | `spark-logging=/tmp/spark-logging`          |
 | fullnameOverride           | String to override release name    | No                | spark-operator                              |
-| rbac.createClusterRole     | See `Migrated Properties`          | No                | false                                       |
```
ewilkins-csi, line 48: I: Custom RBAC doesn't seem to be needed anymore, and using ours caused failures with the latest Spark Operator version

extensions/extensions-helm/aissemble-spark-operator-chart/values.template.yaml
```diff
@@ -48,7 +43,7 @@ spark-operator:
     - name: spark-logging
       mountPath: "/tmp/spark-events"
     - name: ivy-cache
-      mountPath: "/opt/spark/.ivy2"
+      mountPath: "/home/spark/.ivy2"
```
ewilkins-csi, line 46: I: The base Spark Operator image uses the default home location for the `spark` user. Technically we could have updated the home in the image build with `RUN usermod`, but this sets us up to remove the aissemble-spark-operator image entirely after the upgrade to Spark 3.5.

extensions/extensions-helm/aissemble-spark-operator-chart/README.md
carter-cundiff: A: There is more README below this about our custom RBAC that should be cleaned up/ removed

extensions/extensions-helm/aissemble-spark-operator-chart/values.template.yaml
```diff
@@ -30,11 +30,6 @@ spark-operator:
     fsGroup: 1000
     fsGroupChangePolicy: "OnRootMismatch"
 
-  rbac:
-    # -- Create and use RBAC `ClusterRole` resources
-    # -- Set to false in order to enable overriding with our own RBAC template
-    createClusterRole: false
```
carter-cundiff, line 36: Q: Assuming this defaults to `true` now?

pom.xml
```diff
@@ -67,6 +67,7 @@
         <version.pegasus>27.7.18</version.pegasus>
         <version.smile>2.6.0</version.smile>
         <version.weld>3.1.4.Final</version.weld>
+        <version.cucumber.reporting.plugin>5.7.5</version.cucumber.reporting.plugin>
```
carter-cundiff, line 70: Q: Why here instead of the `build-parent`?

carter-cundiff (approved):

extensions/extensions-helm/aissemble-spark-operator-chart/README.md
ewilkins-csi: The first time I read that I thought it was giving v1 to v2 migration details and that the info was still accurate, except the new RBAC is provided by the base chart instead of our chart.  But on second read you're totally correct. Will remove.

extensions/extensions-helm/aissemble-spark-operator-chart/values.template.yaml
ewilkins-csi: Indeed

pom.xml
ewilkins-csi: We use the value in our enforcer extension module, which extends from the root pom instead of build-parent. (That's required to avoid circular dependency, as build-parent uses the enforcer extension itself). Prior to this change, we had the version specified both in the enforcer pom and in build-parent.  This consolidates them into one place.

Issue #255, opened 2024-08-01
Feature: Ensure Dockerfile filtering does not interfere with ARG/ENV values.


ewilkins-csi:
## Description
Fabric8's docker-maven-plugin provides [filtering](https://dmp.fabric8.io/#build-filtering) functionality that is enabled by default. We [addressed this](https://github.com/boozallen/aissemble/pull/201) for `ENV` substitutions through out the code base, but didn't cover `ARG` substitutions.  Unfortunately, we can't take the same approach of dropping the braces (`{}`) because some of the substitutions are embeded in strings (e.g. `python-${PYTHON_VERSION}-dist-utils`).  To avoid this entirely, we should take the recommendation from Fabric8 to set the filtering delimiter to `@`.

## DOD
- [x] Quarkus-based images are updated to use `@` delimiter
- [x] All docker images build successfully with the updated delimiter

## Test Strategy/Script
1. Build the PDP image with a clashing property: 
```sh
mvn clean install -pl :aissemble-policy-decision-point -am -DJAVA_APP_DIR=/fake/path
```
2. Run the policy decision point image: 
```sh
docker run --rm -p8080:8080 ghcr.io/boozallen/aissemble-policy-decision-point:1.8.0-SNAPSHOT
```
3. Verify the service is running: `curl localhost:8080/api/healthcheck`
4. Create a JWT for the user _aissemble_:
```sh
curl -X POST -H "Content-Type: application/json" localhost:8080/api/authenticate -d '{
  "username": "aissemble",
  "password": ""
}'
```
5. Submit a decision request with the token generated in the previous step and verify `PERMIT` is returned
```sh
curl -X POST -H "Content-Type: application/json" localhost:8080/api/pdp -d '{
  "jwt": "TOKEN",
  "resource": "",
  "action": "data-access"
}'
```

csun-cpointe:
Tests passed!!

PR #232, opened 2024-07-24
#216 extensions security mirgration

habibimoiz:
#216 - Baton migration script for refactoring aiops references in extensions-security. 
  - Baton migration scripts for any files that need to be renamed.
 - Baton migrations for any Objects that need to be renamed.

PR #257, opened 2024-08-02
 #255 fix PDP image and change Dockerfile delimiter

ewilkins-csi:
The rename of the security module in #232 broke the default Policy Decision Point image, because of the class rename of `AissembleAttributeExtension` was not propagated to the `pdp.xml` file. The demo/example policy was also broken because the subject/user name and the URN of the attribute that the `LocalAttributePoint` class was looking for was changed, but the test-attribute.json and test-policy.xml included in the image was not updated. Further, the example JWT token was for the wrong subject name.

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -589,6 +589,7 @@
                             <image>
                                 <name>${docker.repo.id}/${dockerImageName}:${dockerImageVersion}</name>
                                 <build>
+                                    <filter>@</filter>
                                     <args>
                                         <DOCKER_BASELINE_REPO_ID>${docker.baseline.repo.id}/</DOCKER_BASELINE_REPO_ID>
                                         <VERSION_AISSEMBLE>${version.aissemble}</VERSION_AISSEMBLE>
--- a/extensions/extensions-docker/aissemble-policy-decision-point/README.md
+++ b/extensions/extensions-docker/aissemble-policy-decision-point/README.md
@@ -5,7 +5,7 @@ on a Quarkus server.
 
 To start this container run
 
-`docker run -p 8780:8080 boozallen/aissemble-policy-decision-point:AISSEMBLE_VERSION`
+`docker run -p 8780:8080 ghcr.io/boozallen/aissemble-policy-decision-point:AISSEMBLE_VERSION`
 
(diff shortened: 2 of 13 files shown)


build-parent/pom.xml
```diff
@@ -589,6 +589,7 @@
                             <image>
                                 <name>${docker.repo.id}/${dockerImageName}:${dockerImageVersion}</name>
                                 <build>
+                                    <filter>@</filter>
```
ewilkins-csi, line 592: I: Primary change

extensions/extensions-docker/aissemble-policy-decision-point/README.md
```diff
@@ -16,15 +16,15 @@ POST localhost:8780/api/pdp
 Body:
 ``` json
 {
-  "jwt": "eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJqdGkiOiI3ZDZmMWZlNS05YzZiLTQ1ZGEtODlmMS0yMDM5YWIwNWZhNDEiLCJzdWIiOiJhaW9wcyIsImF1ZCI6ImF1ZGllbmNlIiwibmJmIjoxNjI0OTc2MDI5LCJpYXQiOjE2MjQ5NzYwMjksImV4cCI6MTkyNDk4MDc5NywiaXNzIjoiYWlvcHMuYXV0aG9yaXR5In0.eUBC2ink77XRf5n5JIXlLZR-fBiRmGrqo1TBFz46yZhWDY38dsh30flELE8gO5SG2rUSIe-VmmjSny8PFLNGwy5MGLwr9z56HoH7OrejJeEzCa1yBl67VWgUZhoDy3RzvARfdBnUstfigHYeQA2ECvW-b2kppYJPVUNX2uKmwfZupwqCGqIX56s7qntV0dUAjpC_KiZ3fjUz1HXqK_evWos0xPVT8XOB2ZADhh87kf7LmocYQ4Y-Z_fsou6jqYh1lQT8WeI2AKskE613nSqmTA2bax5-dOFXKWKLy8t5glyjkdqFZVrLrNkK9tXqNYpZ8efIkZKOu7T9TlsvkHU1XQ",
```
ewilkins-csi, line 19: I: Technically this token is still valid and can be used, but it doesn't behave the same way as you'd expect given the docs (i.e. returning PERMIT for the example policy.)

carter-cundiff (approved):

Issue #238, opened 2024-07-26
Feature: As a user, I want aiSSEMBLE 1.8.0 released, so I have access to the latest features. (Finalize)


ewilkins-csi:
## Description
We are ready to cut the release for 1.8.0. We are still in process of getting GitHub actions ready for the release so will continue to use the old infrastructure for this release.

Before release:
- [x] Resolve Jenkins Agent image build failure

Stretch goal:
- [ ] Need to include the GPG key to sign the Github commit for the release job

After release:
- [x] 1.8.0 artifacts have been published to public repositories
- [x] 1.8.0 release notes have been published
- [x] 1.8.0 documentation been published
- [x] Create 1.9.0 milestone and fix versions
   - [x] Move in-flight tickets/issues to 1.9.0 for both Jira and Github
- [x] Clean up 1.8.0 specific migration code from `foundation-upgrade` module
- [x] Remove `python-linting-migration` from the archetype pom file
- [x] Reset the draft release note

## Test Strategy/Script
1. Create a new project from the 1.8.0 archetype:
```
mvn archetype:generate -DarchetypeGroupId=com.boozallen.aissemble \
                       -DarchetypeArtifactId=foundation-archetype \
                       -DarchetypeVersion=1.8.0 \
                       -DgroupId=org.test \
                       -DartifactId=verify-release \
                       -DprojectGitUrl=test.org/verify-release.git \
                       -DprojectName="Verify 1.8.0 Release" \
&& cd verify-release
```
2. Add a simple pipeline to verify-release-pipeline-models 
[SparkJavaDataDeliveryBasic.json](https://github.com/user-attachments/files/16502964/SparkJavaDataDeliveryBasic.json)

3. Build the project: ./mvnw clean install
4. Follow manual actions
5. Repeat prior two steps until no more manual actions are printed
6. Do an uncached build so the docker target directories are correct: ./mvnw clean install -Dmaven.build.cache.skipCache
7. Deploy the project via Tilt and verify it can be deployed

## References/Additional Context
Add any other context, links, or screenshots about the feature request here.

csun-cpointe:
DoD completed with @ewilkins-csi

nartieri:
Testing passed! 🎉

PR #260, opened 2024-08-05
#238 aiSSEMBLE 1.8 post-release cleanup

csun-cpointe:


test/src/test/resources/docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
 # Test docker image
-FROM ghcr.io/boozallen/aissemble-quarkus:1.8.0-SNAPSHOT
+FROM ghcr.io/boozallen/aissemble-quarkus:1.8.0
```
ewilkins-csi, line 2: S: Needs to be updated to 1.9.0-SNAPSHOT. Same with `test-app/apps/hello-world/Chart.yaml`

csun-cpointe: ah.. good catch!

ewilkins-csi (approved):

csun-cpointe:
PR is ready to merge but require an admin to merge since it requires the signed commit override

Issue #258, opened 2024-08-05
Feature: update policy-decision-point to package service into zip instead of uber-jar


jaebchoi:
## Description
In the past we have packaged jar files using the uber-jar pattern. This has turned out to be problematic for downstream projects due to how plugin.xml is compiled. For this reason, we now package them into a zip file. policy-decision-point still use the old pattern and need to be migrated

## Definition of Done

- Package policy-decision-point app in its respective docker container without using an uber-jar 
- Ensure extension security and docker  aiSSEMBLE build is succeed. 
- Run integration test.
- Run archtype test.
- Test to ensure that policy-decision-point is functioning correctly by hitting their endpoints in downstream projects

## Test Strategy/Script

- Create a downstream project using the below command. You can call it whatever you want. I will use sample22 as an example.
```
mvn archetype:generate \
-DarchetypeGroupId=com.boozallen.aissemble \
-DarchetypeArtifactId=foundation-archetype \
-DarchetypeVersion=1.9.0-SNAPSHOT \
-DgroupId=com.sample22 \
-DartifactId=sample22 \
-DprojectGitUrl=url \
-DprojectName=sample22 
```

- Add the below SimpleDataDeliveryExample.json pipeline to sample22-pipeline-models/src/main/resources/pipelines
```
{
  "name":"SimpleDataDeliveryExample",
  "package":"com.boozallen.aissemble.documentation",
  "type":{
    "name":"data-flow",
    "implementation":"data-delivery-spark"
  },
  "dataLineage": true,
  "steps":[
    {
      "name":"IngestData",
      "type":"synchronous",
      "dataProfiling":{
        "enabled":false
      }
    }
  ]
}
```
- Run mvn clean install + manual actions until done
- Run tilt up and wait for all resources to be in a Ready state.

> NOTE: if policy-decision-point happens to fail with cacheKey, try building mvn clean install without cache.
>  i.e ImageBuild: failed to compute cache key

- Make sure the policy-decision-point container is up and running without any errors

- Using Rest Client,  Make a GET call to get health check API
`GET localhost:8080/api//healthcheck`
- Using a REST client like Postman to make authn request to get token
- Set the URL to localhost:8080/api/authenticate
- Set the Action to POST
- Set the body type to JSON
- Set the body content to:
```
{
  "username": "aissemble",
  "password": ""
}
```

- Click Send
- Expected result:
`JWT token`

- Use above token to make request to PDP  
- Set the URL to localhost:8080/api/pdp
[...]

PR #262, opened 2024-08-06
#258 update policy-decision-point to package service into zip instead of uber-jar

jaebchoi:
In the past we have packaged jar files using the uber-jar pattern. This has turned out to be problematic for downstream projects due to how plugin.xml is compiled. For this reason, we now package them into a zip file. policy-decision-point still use the old pattern and need to be migrated

Github issue:
https://github.com/boozallen/aissemble/issues/258

### DIFF
--- a/extensions/extensions-data-lineage/extensions-data-lineage-http-consumer-service/pom.xml
+++ b/extensions/extensions-data-lineage/extensions-data-lineage-http-consumer-service/pom.xml
@@ -139,26 +139,6 @@
                     <skipOriginalJarRename>true</skipOriginalJarRename>
                 </configuration>
             </plugin>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-dependency-plugin</artifactId>
-                <executions>
-                    <execution>
-                        <id>unpack-docker-resources</id>
-                        <phase>prepare-package</phase>
-                        <goals>
-                            <goal>unpack-dependencies</goal>
-                        </goals>
-                        <configuration>
-                            <outputDirectory>${project.build.directory}</outputDirectory>
-                            <includeArtifactIds>extensions-data-lineage</includeArtifactIds>
-                            <type>zip</type>
-                            <classifier>app</classifier>
-                            <failOnMissingClassifierArtifact>true</failOnMissingClassifierArtifact>
-                        </configuration>
-                    </execution>
-                </executions>
-            </plugin>
             <!-- Packages the quarkus app and its dependencies into a zip-->
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
--- a/extensions/extensions-security/extensions-policy-decision-point-service/pom.xml
+++ b/extensions/extensions-security/extensions-policy-decision-point-service/pom.xml
@@ -12,10 +12,6 @@
     <name>aiSSEMBLE::Extensions::Security::Policy Decision Point Service</name>
     <description>Policy Decision Point Services</description>
 
-    <properties>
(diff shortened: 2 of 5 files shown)


extensions/extensions-security/extensions-policy-decision-point-service/pom.xml
```diff
@@ -85,6 +81,46 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-source-plugin</artifactId>
             </plugin>
+            <plugin>
```
jaebchoi, line 84: I: the plug in step of prepare package should be at last, because it was trying to archive before quarkus-app were generated.

extensions/extensions-security/extensions-policy-decision-point-service/pom.xml
```diff
@@ -85,6 +81,46 @@
...
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
```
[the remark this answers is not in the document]
jaebchoi: I agree with you that unpacking is unnecessary as we do this in docker/extension-pdp. 
I will clean it up and it seems like data-lineage-http-consumer does the same ( have unpacking dependencies in java modules) do you think it is worth clean up with different ticket?

extensions/extensions-security/extensions-policy-decision-point-service/src/main/resources/quarkus-app.xml
```diff
@@ -0,0 +1,29 @@
...
+	</formats>
+
+	<fileSets>
+		<fileSet>
+			<directory>${project.build.directory}/quarkus-app</directory>
+			<outputDirectory>quarkus-app/</outputDirectory>
+		</fileSet>
+	</fileSets>
+
+</assembly>
```
carter-cundiff, line 29: A: Think this file is using tabs instead of spaces.

carter-cundiff (approved):

extensions/extensions-security/extensions-policy-decision-point-service/pom.xml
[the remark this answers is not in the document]
ewilkins-csi: we should just clean it up now

extensions/extensions-security/extensions-policy-decision-point-service/src/main/resources/quarkus-app.xml
jaebchoi: Ah yeah I'll update that real quick!

extensions/extensions-data-lineage/extensions-data-lineage-http-consumer-service/pom.xml
```diff
@@ -139,26 +139,6 @@
                     <skipOriginalJarRename>true</skipOriginalJarRename>
                 </configuration>
             </plugin>
-            <plugin>
```
jaebchoi, line 142: I: As mentioned above, this is unnecessary, so will be cleaning up as part of this PR.

extensions/extensions-data-lineage/extensions-data-lineage-http-consumer-service/src/main/resources/quarkus-app.xml
```diff
@@ -8,22 +8,22 @@
   #L%
   -->
 <assembly
-	xmlns="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.2"
```
jaebchoi, line 11: I: Format change. this was all tabs.

Issue #252, opened 2024-08-01
Feature: As an aiSSEMBLE user, I want helm charts for necessary infrastructure so that I can easily deploy my project


carter-cundiff:
## Description
Currently, every aissemble project must deploy argocd in order to deploy aissemble.  These deployments are not actively maintained on most (if not all) projects.  We should formalize their support with a helm chart that includes support for ArgoCD, nginx ingress, and GPU Operator (optional). Additionally, This will greatly help conformance with IA guidelines.

## DOD
- [x] Create helm charts for the following resources:
  - [x] ArgoCD - utilize [Argo helm chart](https://github.com/argoproj/argo-helm/tree/main/charts/argo-cd)
  - [x] Nginx ingress - utilize [Kubernetes helm chart](https://github.com/kubernetes/ingress-nginx/tree/main/charts/ingress-nginx)
    - [x] Configure ArgoCD to work with the ingress
  - [ ]  GPU Operator (Stretch Goal) - utlilize [NVIDIA helm chart](https://github.com/NVIDIA/gpu-operator/blob/main/deployments/gpu-operator/Chart.yaml)
- [x] Charts should include reasonable defaults and guidance for deploying them
- [x] Create a parent helm chart with options to enable/disable the the above sub-charts
  - [x] Include jenkins in this chart

## Test Strategy/Script
- **Before Testing**:
  - Update your rancher desktop settings to disable Traefik (Preferences -> Kubernetes -> Enable Traefik). Your rancher desktop will need to restart after the change.
  - Install the Argo CD CLI if you have not already: https://argo-cd.readthedocs.io/en/stable/cli_installation/
- **OTS Only**:
  - Run the following to build the chart locally:
    ```
    mvn clean install -pl :aissemble-infrastructure-chart -Dmaven.build.cache.skipCache=true
    ```
  - Use the local `*.tgz` file for all `helm` commands, for example:
    ```
    helm install aissemble-infrastructure YOUR/PATH/aissemble-infrastructure-chart-1.9.0-SNAPSHOT.tgz
    ```
- Install the helm chart with the following command:
```
helm install aissemble-infrastructure oci://ghcr.io/boozallen/aissemble-infrastructure-chart --version 1.9.0-SNAPSHOT
```
- Wait for all the pods to be completed/running using the following command:
```
kubectl get pods --all-namespaces | grep -v 'kube-system'
```
- In a separate terminal, run the following command to allow us to reach our ingress controller:
```
[...]

PR #265, opened 2024-08-08
#252 Create helm chart with necessary infrastructure for deploying a project

carter-cundiff:


extensions/extensions-helm/pom.xml
```diff
@@ -27,7 +27,6 @@
         <module>aissemble-spark-application-chart</module>
         <module>aissemble-spark-operator-chart</module>
         <module>aissemble-lineage-http-consumer-chart</module>
-        <module>aissemble-jenkins-chart</module>
```
carter-cundiff, line 30: I: Jenkins chart is migrated to become a part of the infrastructure chart.

foundation/foundation-mda/src/main/resources/profiles.json
```diff
@@ -866,37 +866,6 @@
 			}
 		]
 	},
-	{
```
carter-cundiff, line 869: I: No longer need Jenkins deployment templates as it's part of infrastructure chart. This chart was never officially released, so no harm in deleting the foundation-mda profile and targets.

extensions/extensions-helm/aissemble-infrastructure-chart/templates/namespaces.yaml
```diff
@@ -0,0 +1,31 @@
+---
+{{- if index .Values "argo-cd" "enabled" }}
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: {{ index .Values "argo-cd" "namespaceOverride" }}
+  annotations:
+    "helm.sh/hook": pre-install
+    "helm.sh/hook-weight": "-10"
```
csun-cpointe, line 9: Q: what does the -10 mean?

carter-cundiff: The weight determines the order the kubernetes resources are installed in a helm chart. Each resource has a default weight of 0. The resources are then installed with the lowest weight first (negative to positive) So the weight of -10 is to ensure this resource is installed first, since it's the namespace that needs to exist before the other resources can be installed.

extensions/extensions-helm/aissemble-infrastructure-chart/tests/namespaces_test.yaml
```diff
@@ -0,0 +1,91 @@
+suite: infrastructure
+templates:
+  - namespaces.yaml
+tests:
+  - it: Should contain a Argo CD namespace
+    documentIndex: 0
+    asserts:
+      - containsDocument:
+          kind: Namespace
+          apiVersion: v1
```
csun-cpointe, line 10: will the document also have the name info? maybe we can also assert `name: argocd` like in the below test

carter-cundiff: Yes that can be done too!

extensions/extensions-helm/aissemble-infrastructure-chart/values.template.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+# Default values for argocd
+argo-cd:
+  enabled: true
+  fullnameOverride: argocd
+  namespaceOverride: argocd
+  crds:
+    keep: false # Remove CRDs on chart uninstall
+  global:
+    domain: argocd.localdev.me
```
csun-cpointe, line 13: can this domain be the override if it's in a test/prod env?

extensions/extensions-helm/aissemble-infrastructure-chart/templates/namespaces.yaml
```diff
@@ -0,0 +1,31 @@
...
+---
+{{- end }}
+{{- if index .Values "ingress-nginx" "enabled" }}
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: {{ index .Values "ingress-nginx" "namespaceOverride" }}
```
ewilkins-csi, line 26: S: I know we're setting the override values in our default, but I think best practice here would still be to default to `Release.Namespace`. This would allow a user to unset these overrides to control the namespace at deployment time. Obviously, if the subcharts don't provide support for `Release.Namespace` it's not useful but most chart developers are trending that direction as a best practice.

ewilkins-csi (approved):

extensions/extensions-helm/aissemble-infrastructure-chart/templates/namespaces.yaml
carter-cundiff: Yes they all provide support for `Release.Namespace` as their default. Thought behind overriding the namespaces here was so you can deploy the helm chart in one command, but not end up with all your charts in the same namespace. I'll move the guidance to setting the namespaces as part of the helm install command in the docs, so we can preserve the `Release.Namespace` functionality.

carter-cundiff: Due to issues in helm with creating a namespace as part of a helm chart that's the same as the `Release.Namespace` this default could not be included. Instead the namespace is only created when the override is present and differs from the `Release.Namespace`.

PR #271, opened 2024-08-13
Zip pipeline-invocation-service dependency jars

J-Clingerman:


extensions/extensions-docker/aissemble-pipeline-invocation/src/main/resources/docker/Dockerfile
```diff
@@ -8,4 +8,4 @@ FROM @docker.baseline.repo.id@/boozallen/aissemble-quarkus:@project.version@ AS
 LABEL org.opencontainers.image.source="https://github.com/boozallen/aissemble"
 
 COPY --from=builder /usr/local/bin/helm /usr/local/bin/helm
-COPY --chown=default target/dockerbuild/*.jar $JAVA_APP_DIR/
+COPY --chown=default target/quarkus-app/ $JAVA_APP_DIR/
```
ewilkins-csi, line 11: S: We should copy the 4 sub directories individually, and in the order that the other quarkus image builds do. This allows us to have the more static stuff like libraries/dependencies cached/resused by docker, and only update the small portion that's likely to change (the actual app.)

ewilkins-csi (approved):
Just one bit of fairly straight-forward feedback. Otherwise, looks good!

extensions/extensions-docker/aissemble-pipeline-invocation/src/main/resources/docker/Dockerfile
J-Clingerman: Smart! Just made the change, thanks Emily.

Issue #273, opened 2024-08-15
BUG: The spark operator doesnt have write access for certain apigroups


J-Clingerman:
## Description
The **sparkoperator rbac.yaml** chart was updated for downstream projects. For the api group, **apiextensions.k8s.io**, the rule is set to only get resources. The **pipeline-invocation-service**, which uses the **sparkoperator**, requires create access to create a pipeline. This issue causes the following error when attempting to create a pipeline via the pipeline-invocation service:

```
customresourcedefinitions.apiextensions.k8s.io is forbidden: User "system:serviceaccount:default:sparkoperator" cannot create resource "customresourcedefinitions" in API group "apiextensions.k8s.io" at the cluster scope 
```
## Definition of Done
- Update the rbac.yaml chart to give the service account, **spark-operator**, the rules **get**, **create**, **update**, and **delete** for the api group **apiextensions.k8s.io**
  - The **spark-operator** is dependent on **aissemble-spark-operator-chart** --> This is the rbac.yaml that needs to be updated
  - Use the previous custom chart that was being used --> Currently we are using the communities spark operator chart that doesnt have the correct access
- Successfully build aissemble
- Run integration test
- Run archetype test
- Create a new downstream project
  - Build the downstream project and complete the manual actions
  - Look inside of the downstream projects <project-name>**/src/main/resources/apps/spark-operator/charts** and check to see if the rbac.yaml file was updated inside the zip file
- Test the pipeline invocation service 

## Testing Steps
- Clone aissemble or navigate to your aissemble directory
`git clone https://github.com/boozallen/aissemble.git`
- Build aissemble
`mvn clean install`
- Create a downstream project using the below command:
```

mvn archetype:generate \
-DarchetypeGroupId=com.boozallen.aissemble \
-DarchetypeArtifactId=foundation-archetype \
-DarchetypeVersion=1.9.0-SNAPSHOT \
-DgroupId=com.issue273 \
-DartifactId=issue273 \
-DprojectGitUrl=url \
-DprojectName=issue273   

```
- Add the below SimpleDataDeliveryExample.json pipeline to issue273/src/main/resources/pipelines

```
{
  "name":"SimpleDataDeliveryExample",
  "package":"com.boozallen.aissemble.documentation",
  "type":{
    "name":"data-flow",
    "implementation":"data-delivery-spark"
  },
  "steps":[
    {
      "name":"IngestData",
      "type":"synchronous",
      "dataProfiling":{
        "enabled":false
      }
    }
[...]

PR #274, opened 2024-08-15
248 Fix GitHub Actions build so integration tests succeed

aaron-gary:
Fixed ClusterRole and ClusterRoleBinding to allow integration tests to succeed

### DIFF
--- a/extensions/extensions-helm/aissemble-spark-operator-chart/pom.xml
+++ b/extensions/extensions-helm/aissemble-spark-operator-chart/pom.xml
@@ -34,4 +34,23 @@
         </plugins>
     </build>
 
+    <profiles>
+        <profile>
+            <id>gh-build</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>${group.helm.plugin}</groupId>
+                        <artifactId>helm-maven-plugin</artifactId>
+                        <configuration>
+                            <!-- Skip dry run due to rbac issues with api groups on EKS cluster -->
+                            <skipDryRun>true</skipDryRun>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+    </profiles>
+
+
 </project>
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ on:
         description: "Branch you want to build"
         required: true
         type: string
-        default: 'dev'
+        default: "dev"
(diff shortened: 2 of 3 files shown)


.github/resources/helm_it_role_binding.yaml
```diff
@@ -0,0 +1,14 @@
+# Permissions required by the GitHub Actions Runner to perform integration tests
```
aaron-gary, line 1: I'm looking for feedback on the ClusterRole and ClusterRoleBinding yaml files.  Should these be merged in?  They are run against the self hosted EKS cluster, but only once.  If we wanted to create a new EKS cluster to host our build jobs then these would come in handy.  Just not sure where to put them.

extensions/extensions-helm/aissemble-spark-operator-chart/pom.xml
```diff
@@ -34,4 +34,23 @@
...
+        <profile>
+            <id>gh-build</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>${group.helm.plugin}</groupId>
+                        <artifactId>helm-maven-plugin</artifactId>
+                        <configuration>
+                            <!-- Skip dry run due to rbac issues with api groups on EKS cluster -->
+                            <skipDryRun>true</skipDryRun>
```
aaron-gary, line 47: This can be removed when https://github.com/boozallen/aissemble/issues/273 is completed

.github/resources/helm_it_role_binding.yaml
cpointe-ibllanos: I'm OK with the permissions as is as long as they're bound to the GitHub service account namespace, which seems to be the case.

As for where to put these - I've seen conventions on other teams where repos are stood up for architecture design files like yaml and terraform stuff. I'm OK where they are now though as it gives new users an easy basis for setting up their own EKS clusters for testing against forks -- maybe just provide documentation in the README about what they should be used for.

cpointe-ibllanos (approved):

.github/resources/helm_it_role_binding.yaml
aaron-gary: I kind of like the idea of storing this in a IAC repo.  I'm going to look into it.

aaron-gary: I decided to yank them from this PR.  FYI

PR #277, opened 2024-08-15
Grant spark-operator write access for the required apigroup

J-Clingerman:
Found the previous rbac.yaml file we were using to define access for reasources and just copied over that. Currently, we were using the community spark-operator rbac.yaml which didnt allow write access for the apiextensions.k8s.io apigroup.

### DIFF
--- a/extensions/extensions-helm/aissemble-spark-operator-chart/templates/rbac.yaml
+++ b/extensions/extensions-helm/aissemble-spark-operator-chart/templates/rbac.yaml
@@ -1,0 +1,140 @@
+{{- /*
+aiSSEMBLE Custom rbac.yaml
+
+Required custom rbac.yaml file that grants the sparkoperator service account 
+create, delete, and update access to the apigroup apiextensions.k8s.io. 
+
+This is necessary for the pipeline-invocation-service to create instances of the 
+SparkApplication CRD to submit pipelines to the Spark Operator for execution.
+*/}}
+
+{{- if or .Values.rbac.create .Values.rbac.createClusterRole }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "spark-operator.fullname" (index .Subcharts "spark-operator") }}
+  annotations:
+    "helm.sh/hook": pre-install, pre-upgrade
+    "helm.sh/hook-delete-policy": hook-failed, before-hook-creation
+    "helm.sh/hook-weight": "-10"
+  labels:
+    {{- include "spark-operator.labels" (index .Subcharts "spark-operator") | nindent 4 }}
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - "*"
+- apiGroups:
+  - ""
+  resources:
+  - services
+  - configmaps
+  - secrets
+  verbs:
+  - create
+  - get
+  - delete
+  - update
+  - list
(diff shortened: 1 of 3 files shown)


extensions/extensions-helm/aissemble-spark-operator-chart/templates/rbac.yaml
```diff
@@ -0,0 +1,130 @@
+{{- if or .Values.rbac.create .Values.rbac.createClusterRole }}
```
ewilkins-csi, line 1: A: Can we include a template comment on the permissions that are being changed from the default Spark Operator RBAC and why they are necessary? Will help prevent accidental deletion in the future.

extensions/extensions-helm/aissemble-spark-operator-chart/values.template.yaml
```diff
@@ -56,4 +61,8 @@ spark-operator:
 
     sparkoperator:
       # -- Optional name for the operator service account
-      name: "sparkoperator"
\ No newline at end of file
+      name: "sparkoperator"
+
+rbac:
+  # -- Set to True in order to enable overriding with our own RBAC template
+  createClusterRole: True
```
ewilkins-csi, line 68: S: There was also some documentation in the README about this custom RBAC that was removed.  Should probably add that back as well.

J-Clingerman: Re-added the rbac.yaml information into the README.

extensions/extensions-helm/aissemble-spark-operator-chart/templates/rbac.yaml
ewilkins-csi: Let's make it an actual [template comment](https://helm.sh/docs/chart_best_practices/templates/#comments-yaml-comments-vs-template-comments) so it doesn't get included in the final resource yaml. Also I don't think the comment is quite correct.

The community Spark Operator chart works perfectly fine without these permissions. Looking back at our offline conversation, I think this is so that the _pipeline invocation service_ can create _instances_ of the SparkApplication CRD (which is novel behavior by the invocation service) to submit pipelines _to the Spark Operator_ for execution.

ewilkins-csi: Although, this does make me wonder why we have to update the spark operator service account to affect the operation of the pipeline invocation service.  Seems like probably the long term solution here is that pipeline invocation service should have it's own service account with the minimal permissions it needs instead of sharing with Spark Operator.

J-Clingerman: > Let's make it an actual [template comment](https://helm.sh/docs/chart_best_practices/templates/#comments-yaml-comments-vs-template-comments) so it doesn't get included in the final resource yaml. Also I don't think the comment is quite correct.
> 
> The community Spark Operator chart works perfectly fine without these permissions. Looking back at our offline conversation, I think this is so that the _pipeline invocation service_ can create _instances_ of the SparkApplication CRD (which is novel behavior by the invocation service) to submit pipelines _to the Spark Operator_ for execution.

I will change the format to the template comment - I was unfamiliar with this comment type thank you! Also, you are correct, the change is required for the pipeline-invocation-service. I wasn't sure if I should add that information. Making the necessary changes now, thank you Emily.

Issue #266, opened 2024-08-08
Feature: Add the ability to inject config store values via local properties files.


csun-cpointe:
## Description
Following up ticket to [#209](https://github.com/boozallen/aissemble/issues/209), now that we have webhook service up and running, we want to be able to inject configs into a properties files that are mounted on the container.

## DOD
Acceptance criteria required to realize the requested feature
- [x] Update the mutating webhook to filter resources:
   - [x] On resource creation only
   - [x] With appropriate metadata label
  ~- [ ] Of Kind ConfigMap~
- [x] Inject the required config store property values into the newly created ConfigMap
   - [x] Define a substitution key that the config store will override with the appropriate value
   - [x] Create a patch for injecting the value to the configMap yaml
- [x] Update the Config Store documentation with detail how to inject values to the ConfigMap
## BDD Scenario:
Scenario: The configuration service can inject values to newly created configMap

## Test Strategy/Script
- Create a downstream project from the archetype using the following command:
  ```
  mvn archetype:generate -U -DarchetypeGroupId=com.boozallen.aissemble \
  -DarchetypeArtifactId=foundation-archetype \
  -DarchetypeVersion=1.9.0-SNAPSHOT \
  -DgroupId=com.test \
  -DartifactId=test-266 \
  -DprojectGitUrl=test.url \
  -DprojectName=test-266 \
  && cd test-266
  ```
- Add a simple pipeline to test-266-pipeline-models
[SparkPipeline.json](https://github.com/user-attachments/files/16603518/SparkPipeline.json)

- Add to the fermenter-mda plugin executions in test-266-deploy/pom.xml
```
<execution>
    <id>configuration-store</id>
    <phase>generate-sources</phase>
    <goals>
        <goal>generate-sources</goal>
    </goals>
    <configuration>
        <basePackage>com.boozallen.aissemble.test</basePackage>
        <profile>configuration-store-deploy-v2</profile>
        <!-- The property variables below are passed to the Generation Context and utilized
             to customize the deployment artifacts. -->
        <propertyVariables>
            <appName>configuration-store</appName>
        </propertyVariables>
    </configuration>
</execution>
```
- Build the project with `mvn clean install` and complete the manual action
- Download and unzip the attached 
[266-helper.zip](https://github.com/user-attachments/files/16605025/266-helper.zip)

- Mac users
  - In the root directory, find the `Tiltfile-config` file and copy the src directory to the root of the project and update the `configuration_files_path` to the following:
```
[...]

PR #278, opened 2024-08-15
#266 config store property value injection

csun-cpointe:


foundation/foundation-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
```diff
@@ -65,6 +65,7 @@
 				<include>devops/**</include>
 				<include>jenkinsPipelineSteps.groovy</include>
 				<include>.mvn/**</include>
+				<include>Tiltfile-config</include>
```
ewilkins-csi, line 68: C: Let's not introduce this here. I don't think this is what we want the final user experience for the config store to be. So for now, we can use this approach to test as we're developing but before we ship we'll need a better solution.

foundation/foundation-configuration-store/README.md
```diff
@@ -7,21 +7,41 @@ please refer to the [aiSSEMBLE&trade; Github Pages](https://boozallen.github.io/
 
 ## Developer Guidance
 
-### Deserializing YAML Content into Java Objects with snakeyaml
-The `ConfigLoader` leverages [snakeyaml](https://bitbucket.org/snakeyaml/snakeyaml/src/master/) to parse `YAML` files define configurations and associated metadata. This module contains two classes, `YamlProperty` and `YamlConfig`, which are the objects that house the deserialized information from the configuration `YAML` files. An example of properly formatted configurations files:
+### Deserializing Properties Content into Java Objects with Krausening
+The `ConfigLoader` leverages [Krausening](https://github.com/TechnologyBrewery/krausening/tree/dev/krausening) to parse properties files define configurations and associated metadata, which enables the configuration value encryption for security reasons. 
+When defining a property, the property name should be the group name and property content should be in key=value format. An example of properly formatted configurations files:
```
ewilkins-csi, line 12: A: This line seems incorrect but I'm not sure what we're trying to say. Specifically "_the property name should be the group name_" sounds off. I don't know what that means.

foundation/foundation-configuration-store/pom.xml
```diff
@@ -149,6 +149,12 @@
             <artifactId>kubernetes-client</artifactId>
             <version>6.10.0</version>
         </dependency>
+        <dependency>
+            <groupId>com.flipkart.zjsonpatch</groupId>
+            <artifactId>zjsonpatch</artifactId>
+            <version>0.4.16</version>
```
ewilkins-csi, line 155: S: I'm very reluctant about adding a dependency on a pre 1.0 version of a library that hasn't been updated in nearly a year.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -33,31 +48,124 @@
...
         AdmissionRequest request = admissionReviewRequest.getRequest();
 
-        AdmissionResponse response = new AdmissionResponseBuilder()
-                                                    .withAllowed(true)
-                                                    .withUid(request.getUid())
-                                                    .build();
+        JsonFactory factory = JsonFactory.builder()
+                .enable(JsonReadFeature.ALLOW_BACKSLASH_ESCAPING_ANY_CHARACTER)
+                .enable(JsonReadFeature.ALLOW_UNESCAPED_CONTROL_CHARS)
+                .enable(JsonReadFeature.ALLOW_JAVA_COMMENTS)
```
ewilkins-csi, line 79: Q: Why allow Java Comments?

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -33,31 +48,124 @@
...
+        return null;
+    }
+
+    private StringBuilder replaceConfigStoreKeyWithConfigValue(String source) {
+        StringBuilder target = new StringBuilder();
+        String[] data = source.split(CONFIG_STORE_INJECT_START);
+        // if the config store key is found
+        if (data.length > 1) {
+            target.append(data[0]);
+            logger.info("Config store properties injection requested: ");
```
ewilkins-csi, line 144: Q: Why end with a colon here? Did we expect more info to be added, like maybe the property key?

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -33,31 +48,124 @@
...
+            return property.getValue();
+        } catch (Exception e) {
+            logger.error(String.format("Error reading property - groupName: %s, propertyName: %s", groupName, propertyName), e);
+        }
+        return null;
+    }
+
+    private StringBuilder replaceConfigStoreKeyWithConfigValue(String source) {
+        StringBuilder target = new StringBuilder();
+        String[] data = source.split(CONFIG_STORE_INJECT_START);
```
ewilkins-csi, line 140: C: I don't think this allows multiple configs in one YAML.  You might use a Scanner with the delimiter set to your inject start constant.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -33,31 +48,124 @@
...
+    }
+
+    public String getProperty(String groupName,String propertyName) {
+        PropertyKey propertyKey = new PropertyKey(groupName, propertyName);
+        Property property;
+        try {
+            property = this.configLoader.read(propertyKey);
+            return property.getValue();
+        } catch (Exception e) {
+            logger.error(String.format("Error reading property - groupName: %s, propertyName: %s", groupName, propertyName), e);
```
ewilkins-csi, line 133: S: We should just re-throw an exception here and allow it to bubble up to the top where we return the fail response to k8s.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -33,31 +48,124 @@
...
+                .enable(JsonReadFeature.ALLOW_UNESCAPED_CONTROL_CHARS)
+                .enable(JsonReadFeature.ALLOW_JAVA_COMMENTS)
+                .enable(JsonReadFeature.ALLOW_YAML_COMMENTS)
+                .enable(JsonReadFeature.ALLOW_SINGLE_QUOTES)
+                .build();
+
+        ObjectMapper mapper = new ObjectMapper(factory);
+        AdmissionResponse response;
+        try {
+            JsonNode requestJson = mapper.readTree(new StringReader(mapper.writeValueAsString(request)));
```
ewilkins-csi, line 87: S: This is really odd to me. Why are we writing the request to a string, then to a generic JSON object, then back to a string? Why not just do:
```
String source = mapper.wratieValueAsString(request.getObject());
StringBuilder target = replaceConfigStorKeyWithConfigValue(source);
```

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -33,31 +48,124 @@
...
+        try {
+            property = this.configLoader.read(propertyKey);
+            return property.getValue();
+        } catch (Exception e) {
+            logger.error(String.format("Error reading property - groupName: %s, propertyName: %s", groupName, propertyName), e);
+        }
+        return null;
+    }
+
+    private StringBuilder replaceConfigStoreKeyWithConfigValue(String source) {
```
ewilkins-csi, line 138: A: If we fix the multi-replacement issue by making this method replace all injection points, we should rename this method so it doesn't sound singular.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -33,31 +48,124 @@
...
+                Map<String, String> keyMap = convertToKeyMap(key);
+                String value = getProperty(keyMap.get(GROUP_NAME), keyMap.get(PROPERTY_NAME));
+                target.append(value);
+                if (data[i].length() > contentIndex + 1) {
+                    target.append(data[i].substring(contentIndex+1));
+                }
+            }
+        }
+
+        return target;
```
ewilkins-csi, line 157: A: Based on how you use this return value, seems cleaner to return the String instead of StringBuilder.  Also, I feel like the name could be a little better. Maybe something like `updatedJson` or `processedBody`.

foundation/foundation-configuration-store/src/test/java/com/boozallen/aissemble/configuration/mutating/webhook/MutatingWebhookSteps.java
```diff
@@ -12,30 +12,59 @@
...
-    private ObjectMapper objectMapper = new ObjectMapper();
+    private ConfigMap configMap;
+    private Map<String, String > configMapData;
+    private String responsePatch;
+
+    @Given("a ConfigMap definition that contains the substitution key exists")
+    public void aConfigMapDefinitionThatContainsTheSubstituionKeyExists() {
+        final String propertyKey = "groupName=aws-credentials;propertyName=AWS_ACCESS_KEY_ID";
+        configMapData = new HashMap<>();
+        configMapData.put("AWS_ACCESS_KEY_ID", ConfigMutatingWebhook.CONFIG_STORE_INJECT_START.replaceAll("\\\\", "") + propertyKey + ConfigMutatingWebhook.CONFIG_STORE_INJECT_END);
```
ewilkins-csi, line 52: A: Instead of writing the string literal for INJECT_START as a regex and then trying to unescape it here, you could store the exact string in the literal and then use `Pattern.quote` anywhere it needs to be passed in as a regex.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
csun-cpointe: yeah that makes sense. Let me simplify it and parse to Json later for comparison.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
csun-cpointe: it actually does multiple replacement since there is a loop and it seems String.split actually provides simplest api for it.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
csun-cpointe: I can remove it but just put it there in case.

foundation/foundation-configuration-store/pom.xml
csun-cpointe: yeah but it does seem this is already included in the quarkus-resteasy-jackson api so it actually can be cleaned up.

ewilkins-csi: No we should leave it if we're interfacing with it directly.  Best practice is to always list the dependencies you directly use as direct dependencies, even when they're available transitively.  This makes it much easier to maintain as transitive dependencies are removed/updated.

ewilkins-csi (approved):
Just two final comments. Otherwise LGTM.

PR #296, opened 2024-08-26
 #270 Adjust prune workflow

chang-annie:
- Update prune workflow to use `dataaxiom/ghcr-cleanup-action@v1`
- Pass in all aiSSEMBLE docker image names through a matrix
- Pass in list of release versions to the `exclude-tags` parameter (e.g., `1.7.0,1.8.0`)
- Use `keep-n-tagged: 2` to ensure that the two most recent NON-release tagged images are saved (e.g., `1.9.0-SNAPSHOT` & `1.8.0-rc8`)

### DIFF
--- a/.github/workflows/prune.yml
+++ b/.github/workflows/prune.yml
@@ -1,4 +1,4 @@
-# Prunes old aissemble-* untagged containers
+# Prunes old aissemble-* containers, excluding all release tags and the two most recent non-release tags
 
 name: Prune ghcr.io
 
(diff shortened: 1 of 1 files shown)


.github/workflows/prune.yml
```diff
@@ -10,19 +10,74 @@ on:
...
-          cut-off: Two days ago UTC
-          account-type: org
-          org-name: boozallen
-          keep-at-least: 2
-          untagged-only: true
-          dry-run: false
+          log-level: info
+          package: ${{ matrix.package-name }}
           token: ${{ secrets.GHCR_IO_TOKEN }}
+          exclude-tags: ${{ needs.fetch_image_info.outputs.release-values }}
```
chang-annie, line 82: I: `exclude-tags` is set to the value of the `release-values` output from phase 1, which is a string containing a comma separated list of release versions (e.g., `1.7.0,1.8.0`)

.github/workflows/prune.yml
```diff
@@ -10,19 +10,74 @@ on:
...
+  
+  # Phase 2: loop through output of phase 1 to prune images
+  prune_ghcr:
+    # Tells workflow to wait for the "fetch_image_info" phase to finish
+    needs: fetch_image_info
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        package-name: ${{ fromJson(needs.fetch_image_info.outputs.matrix-values) }}
```
chang-annie, line 73: I: Here, we're identifying the matrix values that will be used repeat the prune step (aka the list of docker image names). We have to repeat this step for each docker image because the `ghcr-cleanup-action` action only allows us to identify one package to prune at a time.

The matrix values are given as a JSON array containing the image names. For example,
```
["aissemble-vault","aissemble-quarkus","aissemble-airflow","aissemble-fastapi","aissemble-hive-service","aissemble-hive-mysql",...]
```

.github/workflows/prune.yml
```diff
@@ -10,19 +10,74 @@ on:
...
+                -H "Authorization: Token ${GH_TOKEN}" \
+                --paginate "/orgs/boozallen/packages/container/${name}/versions" \
+                | jq -r '.[] | .metadata.container.tags[] | select(test("^\\d+\\.\\d+\\.\\d+$"))')
+          
+              # append all release tags to all_release_versions array
+              all_release_versions+=("${release_versions}")
+          done
+          
+          # replace all whitespaces with newline (required for sort) and then sort all release versions while removing any duplicates
+          all_release_versions_sorted=$(echo "${all_release_versions[@]}" | tr ' ' '\n'| sort -u )
```
chang-annie, line 50: I: At this point,`all_release_versions` looks something like:
```
1.8.0
1.7.0 1.8.0
1.7.0 1.7.0 1.8.0
1.7.0 1.8.0
1.7.0 1.8.0
...
```

I found that the simplest way to remove duplicates here is to use `sort -u`, which first sorts them numerically, then removes duplicates. This requires that we have all elements separated by `\n`. 

Other solutions considered were: 
- use the [awk](https://stackoverflow.com/questions/53742223/how-to-remove-repeated-rows-in-awk) command to remove duplicates
- only adding new elements to `all_release_versions` if it does not already exist

.github/workflows/prune.yml
```diff
@@ -10,19 +10,74 @@ on:
...
+          
+          # replace the newlines with commas and then remove the trailing comma
+          all_release_versions=$(echo "${all_release_versions_sorted[@]}" | tr '\n' ',' | sed 's/,$//')
```
chang-annie, line 53: I: At this point, `all_release_versions` looks something like this:
```
1.7.0
1.8.0
```
To achieve the format we need, we replace all `\n` with commas to get:
```
1.7.0,1.8.0,
```
Lastly, we need to remove the trailing comma to get:
```
1.7.0,1.8.0
```

ewilkins-csi (approved):

.github/workflows/prune.yml
rohanmars: Seeing how you had to implement the exclude-tags with an external regex it makes me think regex support should be directly supported on the exclude-tags setting in the GitHub cleanup action. I'll work on adding that function in the ghcr-cleanup-action.

chang-annie: thanks @rohanmars!! we had trouble finding a clean up action that supported multi-arch images and regex so that would be a great feature to add.

rohanmars: Okay great. Yes, I had a hard time finding a multi-arch action so I ended up just building one, haha. I'll add regex support soon. I think it might make sense also to add a "packages" mode on the action so it could support different packages (images) without having todo a matrix. I think to think about that more.

chang-annie: that would be great, thanks again!

rohanmars:
I've added multiple package and regex support to v1.0.11/v1 now.

You should be able to cleanup without any script magic using something like:

```yaml
      - name: Prune ghcr
        uses: dataaxiom/ghcr-cleanup-action@v1
        with:
          log-level: info
          packages: "^aissemble(?!.*?-chart).*$"
          expand-packages: true
          exclude-tags: "^\\d+\\.\\d+\\.\\d+$"
          use-regex: true
          token: ${{ secrets.GHCR_IO_TOKEN }}
          keep-n-tagged: 2
          dry-run: true
```

If you want to switch/test that that would be great. No worries if not. Note, I haven't tested the exact configuration above, but have set it up from looking at your script, so I have set the dry-run to true, which you might want to use first. You wouldn't need any custom script anymore (the fetch_image_info job). The packages regex might need to be tweaked, I figured that out from looking at your jq query.

chang-annie:
> I've added multiple package and regex support to v1.0.11/v1 now.
> 
> You should be able to cleanup without any script magic using something like:
> 
> ```yaml
>       - name: Prune ghcr
>         uses: dataaxiom/ghcr-cleanup-action@v1
>         with:
>           log-level: info
>           packages: "^aissemble(?!.*?-chart).*$"
>           expand-packages: true
>           exclude-tags: "^\\d+\\.\\d+\\.\\d+$"
>           use-regex: true
>           token: ${{ secrets.GHCR_IO_TOKEN }}
>           keep-n-tagged: 2
>           dry-run: true
> ```
> 
> If you want to switch/test that that would be great. No worries if not. Note, I haven't tested the exact configuration above, but have set it up from looking at your script, so I have set the dry-run to true, which you might want to use first. You wouldn't need any custom script anymore (the fetch_image_info job). The packages regex might need to be tweaked, I figured that out from looking at your jq query.

This is great, thank you! We'll be trying out the new features in the near future and will reach out with feeback, if needed.

chang-annie:
> near futu

Hey @rohanmars, we just implemented these new features and they worked smoothly and as expected! We tested the new configuration and we were able to completely remove the custom scripting. A big thank you!

rohanmars:
That's great news @chang-annie . And thanks for the inspiration and requirement, adding that capability to the action greatly improves it, I've already used this new capability in my own projects!

Let me know if there are any issues with it.

PR #297, opened 2024-08-26
#270-2 Fix to remove empty values from release_values

chang-annie:
This is to remove any rows that are blank from the `all_release_versions` array before saving them to output. Without this added logic, if we have any docker images that do not have any release versions, the `exclude-tags` parameter gets set to `,1.7.0,1.8.0` (with an extra comma at the front).

### DIFF
--- a/.github/workflows/prune.yml
+++ b/.github/workflows/prune.yml
@@ -49,8 +49,8 @@ jobs:
           # replace all whitespaces with newline (required for sort) and then sort all release versions while removing any duplicates
           all_release_versions_sorted=$(echo "${all_release_versions[@]}" | tr ' ' '\n'| sort -u )
           
-          # replace the newlines with commas and then remove the trailing comma
-          all_release_versions=$(echo "${all_release_versions_sorted[@]}" | tr '\n' ',' | sed 's/,$//')
+          # remove any blank rows, replace the newlines with commas, then remove the trailing comma
+          all_release_versions=$(echo "${all_release_versions_sorted[@]}" | sed -r '/^\s*$/d'| tr '\n' ',' | sed 's/,$//')
           
           # save to output
           echo "release_values=${all_release_versions}" >> $GITHUB_OUTPUT


.github/workflows/prune.yml
```diff
@@ -49,8 +49,8 @@ jobs:
           # replace all whitespaces with newline (required for sort) and then sort all release versions while removing any duplicates
           all_release_versions_sorted=$(echo "${all_release_versions[@]}" | tr ' ' '\n'| sort -u )
           
-          # replace the newlines with commas and then remove the trailing comma
-          all_release_versions=$(echo "${all_release_versions_sorted[@]}" | tr '\n' ',' | sed 's/,$//')
+          # remove any blank rows, replace the newlines with commas, then remove the trailing comma
```
chang-annie, line 52: I: ` sed -r '/^\s*$/d'` command checks for any rows that only contain whitespace and deletes it

.github/workflows/prune.yml
```diff
@@ -49,8 +49,8 @@ jobs:
...
+          all_release_versions=$(echo "${all_release_versions_sorted[@]}" | sed -r '/^\s*$/d'| tr '\n' ',' | sed 's/,$//')
```
ewilkins-csi, line 53: Q: Can you not just use a `-z` test of `release_versions` on line 46?

ewilkins-csi (approved):

.github/workflows/prune.yml
chang-annie: @ewilkins-csi similar to the `-z` zero length check, we could use the `-n` not empty check and loop through the `release_versions` array
```
  # Iterate over each element in release_versions
  for version in "${release_versions[@]}"; do
      # Check if the element is not empty
      if [ -n "$version" ]; then
          # Append the element to all_release_versions
          all_release_versions+=("$version")
      fi
  done
```
in our case,  I think adding the command to remove blank rows at the end is fine since we don't have that many docker images

Issue #284, opened 2024-08-21
Feature: Provide migration path for projects transitioning from v1 Spark Operator helm chart to v2


carter-cundiff:
## Description
The spark operator v2 helm chart was previously created and set as the default in new projects, however, migration instructions were never created for updating existing projects. We need to create guidance for users to update their existing spark operator chart to the new v2 one.

## DOD
- [x] Deprecate the v1 profile: `aissemble-spark-operator-deploy`
  - [x] Message should include that v1 is only compatible with `1.7.0` image
  - [x] Update migration to skip updating spark operator v1 image version
  - [x] Fix link in deprecation warning messages 
- [x] Update chart README with instructions for migrating to the v2 chart:
  - [x] Include new property locations and any specialty instructions
- [x] Update documentation noting the migration is available in the following locations:
  - [x] Create `1.8.1` patch release notes
    - [x] Link to general migration instructions in antora docs
  - [x] Update list of available V2 profiles on [containers.adoc antora page](https://boozallen.github.io/aissemble/aissemble/current/containers.html)

## Implementation Notes
- [x] Create `1.8.1` patch release branch off of `1.8.0` tag
  - [x] Run `mvn versions:set` command to update to `1.8.1-SNAPSHOT`
  - [x] Manually update two files in `/tests`
  - [x] Update antora docs
- [x] Create feature branch off patch release branch
- [x] Merge into `1.8.1` patch release branch
- [x] Cherry pick back to `dev`

### Spark Operator Upgrade
- **OTS Only:**
  - Build the entire aiSSEMBLE project to create all the `1.8.1-SNAPSHOT` artifacts locally
- Create a archetype project on version `1.7.0`:
```
mvn archetype:generate -U -DarchetypeGroupId=com.boozallen.aissemble \
  -DarchetypeArtifactId=foundation-archetype \
  -DarchetypeVersion=1.7.0 \
  -DgroupId=com.test \
  -DartifactId=test-284 \
  -DprojectGitUrl=test.url \
  -DprojectName=test-284 \
  && cd test-284
```
- Add the [SparkPipeline.json](https://github.com/user-attachments/files/16752612/SparkPipeline.json) file to the `test-284-pipeline-models/src/main/resources/pipelines` directory
- Run `mvn clean install` until all the manual actions are complete 
[...]

PR #299, opened 2024-08-27
#284 Create migration instructions for transitioning from v1 Spark Operator helm chart to v2

carter-cundiff:


DRAFT_RELEASE_NOTES.md
```diff
@@ -1,5 +1,8 @@
 # Major Additions
 
+## Helm Migration
+The Spark Operator Helm chart has been migrated to the `v2` structure.  To migrate your helm charts to use the v2 pattern, please follow the [instruction](https://boozallen.github.io/aissemble/aissemble/current/containers.html#_kubernetes_artifacts_upgrade) in the technical documentation.
```
ewilkins-csi, line 4: A: IMO, instead of approaching this like we would a normal v2 migration, we should say the issue we're patching (that the v1 chart is not compatible with 1.8) and clarify that projects created after 1.2.10 will already have the v2 chart.

ewilkins-csi (approved):

PR #318, opened 2024-09-06
Fix CI profile to publish into nexus

jaebchoi:
Looks like downstream projects are publishing into test.pypi.org.
That was because build-parent pom file was setting useDevRepository to true, in which downstream project inherit those configuration.
Update the config to correctly publish to nexus.

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -146,6 +146,10 @@
         <version.helm.plugin>6.14.3</version.helm.plugin>
         <version.helm.min>3.12.0</version.helm.min>
         <version.snakeyaml>2.0</version.snakeyaml>
+
+        <!-- Used for publishing artifacts into internal nexus repo-->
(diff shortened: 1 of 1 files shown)


build-parent/pom.xml
```diff
@@ -146,6 +146,10 @@
         <version.helm.plugin>6.14.3</version.helm.plugin>
         <version.helm.min>3.12.0</version.helm.min>
         <version.snakeyaml>2.0</version.snakeyaml>
+
```
d-ryan-ashcraft, line 149: C (Critical): these do not belong here.  This is an open-source project, so we don't push anything into an internal repository nor reference an internal repository in its code base.

Please close this PR and remove the branch.

jaebchoi: I think I misunderstood the intention here.
I got it now, We need to create CI profile in the Video processing not aissemble root. closing.

PR #319, opened 2024-09-06
285 GitHub actions deploy artifacts

aaron-gary:


cwoods-cpointe (approved):

.github/workflows/build.yml
```diff
@@ -87,10 +87,14 @@ jobs:
           distribution: 'temurin'
       - name: Create Docker Builder Config File
         run: sudo touch /etc/buildkitd.toml
+      # Generate the settings.xml for ghcr.io, pypi, & dev-pypi server profiles
+      - name: Create settings.xml
+        run: |
+          echo "<settings><servers><server><id>ghcr.io</id><username>${{ secrets.GHCR_IO_USERNAME }}</username><password>${{ secrets.GHCR_IO_TOKEN }}</password></server><server><id>pypi</id><username>${{ secrets.PYPI_USERNAME }}</username><password>${{ secrets.PYPI_TOKEN }}</password></server><server><id>dev-pypi</id><username>${{ secrets.TEST_PYPI_USERNAME }}</username><password>${{ secrets.TEST_PYPI_TOKEN }}</password></server> </servers></settings>" > $HOME/.m2/settings.xml
```
csun-cpointe, line 93: Q: will this show the token in the log?

csun-cpointe (approved):

jocobtt (approved):

.github/workflows/build.yml
ewilkins-csi, line 93: Q: Did we look into using the `setup-java` feature for [creating the Maven settings](https://github.com/actions/setup-java?tab=readme-ov-file#maven-options) file? If not, might be a better alternative in the long run.

aaron-gary: Correct.  The GH Actions api will always try to mask any logging/console output where sensitive values are involved.

aaron-gary: Yes, I looked into it, but it seems to have some limitations, such as it assumes you are setting a single server.  I'll give it another look, though

ewilkins-csi: No worries! I thought there might be something preventing us from using it, but wanted to make sure we were aware of it.  It might make sense to open a feature request with them if they don't already have something similar on their backlog.

aaron-gary: Great idea. I'll look into opening a feature request

Issue #322, opened 2024-09-09
Feature: Generate Config Store by default and ensure proper documentation is in place under Supporting Component


carter-cundiff:
## Description
Now that the config store MVP is ready to go, we would like to enable it to be generated automatically into downstream projects by default. This includes creating manual actions for adding downstream and providing documentation outlining how to use it.

## DOD
- [x] Enable configuration-store deployment by default in downstream project
  - [x] Create manual action for adding the configuration-store execution to the deploy pom for projects without it
    - [x] **Stretch Goal:** #219
    - [x] Verify krausening is the default store backing and vault is not generated in 
- [x] Update the related documentation
  - [x] Release notes link to antora doc page
  - [x] Move the current [antora doc guide ](https://boozallen.github.io/aissemble/aissemble/current/guides/guides-configuration-store.html) to a page within supporting components
  - [x] Update content to insure it's all relevent to the MVP
    - [x] Remove details about policy configuration 

## Test Strategy/Script
**Test New Project**
- **OTS Only:**
  - Run the following: `mvn clean install -pl :foundation-mda -Dmaven.build.cache.skipCache=true`

- Create project from `1.9.0-SNAPSHOT`:
```
mvn archetype:generate -U -DarchetypeGroupId=com.boozallen.aissemble \
  -DarchetypeArtifactId=foundation-archetype \
  -DarchetypeVersion=1.9.0-SNAPSHOT \
  -DgroupId=com.test \
  -DartifactId=test-322\
  -DprojectGitUrl=test.url \
  -DprojectName=test-322 \
  && cd test-322
```
- Add the [SparkPipeline.json](https://github.com/user-attachments/files/16752612/SparkPipeline.json) file to the `test-322-pipeline-models/src/main/resources/pipelines` directory
- Run `mvn clean install -Dfermenter.display.message.keys=true` once and verify you see the following manual action as part of the `test-322-deploy/pom.xml` group
```
[WARNING] Message Key: test-322-deploy/pom.xml_execution_configuration-store
<execution>
    <id>configuration-store</id>
    <phase>generate-sources</phase>
    <goals>
        <goal>generate-sources</goal>
    </goals>
    <configuration>
        <basePackage>com.test</basePackage>
        <profile>configuration-store-deploy-v2</profile>
        <!-- The property variables below are passed to the Generation Context and utilized
                to customize the deployment artifacts. -->
        <propertyVariables>
            <appName>configuration-store</appName>
        </propertyVariables>
    </configuration>
[...]

Issue #309, opened 2024-08-29
BUG: Configuration Store `supplementalVolumes.readOnly = true` is causing the Out of Sync status in ArgoCD


csun-cpointe:
## Description
When using argocd to deploy and sync the `configuration-store` app, the sync status will remain `Out of Sync`. When clicking into the pod, it will show the live manifest and desired manifest are different. The difference is `supplementalVolumes.readyOnly is true`. We need to investigate the root cause of the issue and make the fix.

![Screenshot 2024-08-29 at 12 37 24 PM](https://github.com/user-attachments/assets/90c6a534-60dc-4682-b4c9-df51f8f9d054)

![Screenshot 2024-08-29 at 12 37 40 PM](https://github.com/user-attachments/assets/37049b45-5ae6-4298-9138-0f8868e201d4)

## Steps to Reproduce
Clear, specific, and detailed steps taken to enable reproduction of the bug for investigation.
- Create a downstream project with baseline version 1.9.0-SNAPSHOT
```
mvn archetype:generate -U -DarchetypeGroupId=com.boozallen.aissemble \
  -DarchetypeArtifactId=foundation-archetype \
  -DarchetypeVersion=1.9.0-SNAPSHOT \
  -DgroupId=com.test \
  -DartifactId=test-defect \
  -DprojectGitUrl=test.url \
  -DprojectName=test-defect \
  && cd test-defect
```
- Add the [SparkPipeline.json](https://github.com/user-attachments/files/16781488/SparkPipeline.json) file to the test-defect-pipeline-models/src/main/resources/pipelines directory
- Add to the fermenter-mda plugin executions in test-defect-deploy/pom.xml
```
<execution>
    <id>configuration-store</id>
    <phase>generate-sources</phase>
    <goals>
        <goal>generate-sources</goal>
    </goals>
    <configuration>
        <basePackage>com.boozallen.aissemble.test</basePackage>
        <profile>configuration-store-deploy-v2</profile>
        <!-- The property variables below are passed to the Generation Context and utilized
             to customize the deployment artifacts. -->
        <propertyVariables>
            <appName>configuration-store</appName>
        </propertyVariables>
    </configuration>
</execution>
```
- Run mvn clean install until all the manual actions are complete
- Once the manual actions are complete, run mvn clean install -Dmaven.build.cache.skipCache=true once to get any remaining manual actions
[...]

PR #329, opened 2024-09-12
#322 Set config store to be included by default

carter-cundiff:
- Update documentation to live under supporting components


docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -0,0 +1,172 @@
...
+It then provides a standardized way of accessing them, allowing the environment specific configurations to be dynamically provided to 
+their respective resources within the project at runtime.
+
+=== Setup
+To setup the configuration store, one must first create properties files for storing the configurations. Each individual configuration is 
+represented as a property key value pair within the file. These files must then be made accessible to the configuration store kubernetes 
+resource for ingestion. See the applicable sections below for more details on defining configurations as properties, creating environment 
+specific property overrides, mounting the properties, and specifying the configuration store backend storage type.
+
+==== Defining Configuration Properties
```
ewilkins-csi, line 14: S: I think we have an opportunity to offload a lot of this explanation by linking the Krausening docs.  I'd also suggest that for now we don't bring in the concept of a "group" until it's absolutely necessary (i.e. during the Usage docs where there is an explicit parameter call "group" that has to be provided.)

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -0,0 +1,172 @@
...
+
+== Overview
+The Universal Configuration Store is a tool that enables the various configurations for a project to be centrally defined and managed. 
+It then provides a standardized way of accessing them, allowing the environment specific configurations to be dynamically provided to 
+their respective resources within the project at runtime.
+
+=== Setup
+To setup the configuration store, one must first create properties files for storing the configurations. Each individual configuration is 
+represented as a property key value pair within the file. These files must then be made accessible to the configuration store kubernetes 
+resource for ingestion. See the applicable sections below for more details on defining configurations as properties, creating environment
```
ewilkins-csi, line 11: A: Any time we say "See XYZ..." we should just link the section we're talking about.  In-window links for any internal navigation and new-window links for any external navigation.  I'd also suggest where we can working the link more organically into a sentence rather than outright saying "See XYZ". E.g. The first sentence could read _"...must first [define the configurations] through properties files."_ where _define the configurations_ links to the relevant section.

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -0,0 +1,172 @@
...
+Here is an example of a base properties file named `messaging.properties`:
+[source,properties]
+----
+connector=smallrye-kafka
+topic=baseTopic
+----
+This file defines two properties, `connector` and `topic`, each with the group name `messaging`.
+
+===== Environment Specific Overrides
+In addition to the base properties file introduced above, one can optionally specify a secondary properties file that houses environment
```
ewilkins-csi, line 29: S: Again I think we can largely lean on Krausening docs here and only call out the aiSSEMBLE-specific stuff.

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -0,0 +1,172 @@
...
+[source,properties]
+----
+connector=smallrye-kafka
+topic=ciTopic
+auth-token=auth-secret
+----
+
+NOTE: The `topic` property is overwritten when the environment specific value `ciTopic`.
+
+==== Mounting Properties
```
ewilkins-csi, line 49: S: I think this would be more consumable if the over-arching section was "Deploying" and we frame everything in that context.  We could have a quick intro paragraph that talks about/links the Helm chart documentation, and what the key goals are of the deployment setup (making the configuration properties files available to the service and choosing the backing store)

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -0,0 +1,172 @@
...
+[source,yaml]
+----
+  env:
+    BASE_PROPERTY: <URI housing base property configurations>
+    ENVIRONMENT_PROPERTY: <URI housing CI-specific property overrides/augmentations>
+----
+
+==== Backend Storage Options
+The universal configuration store offers robust production ready configuration storage using https://developer.hashicorp.com/vault[Vault,role=external,window=_blank]. This is an excellent solution for large 
+enterprise applications. For smaller applications looking for a more light weight but still robust storage option, the configuration store
```
ewilkins-csi, line 76: A: Let's talk about the light-weight solution first and mention it's the default.  And then we can segue into "For enterprise applications that have more robust requirements.."

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -0,0 +1,172 @@
...
+
+The default property storage is Krausening. To change it to use one of the other options, set the environment
+variable `STORAGE_CLASS` to the desired value. The values currently available are `krausening`, `vault`, and `inMemory`.
+
+=== Usage
+After the configurations are ingested and reconciled within the store, the next step is to access them within a project using one of 
+the two available methods: the API endpoint or kubernetes resource injection. See the applicable sections below for more details on 
+utilizing these approaches.
+
+==== API Endpoint
```
ewilkins-csi, line 88: A: Maybe "REST Endpoint" instead of API?

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -0,0 +1,172 @@
...
+to access any configurations from within a project.
+
+.Endpoints
+[cols="2,1,4"]
+|===
+| Endpoint Path | HTTP Method | Description
+
+| `/aissemble-properties/{groupName}/{propertyName}`
+| `GET`
+| Returns a JSON object containing the value of a property with the given group name and property name.
```
ewilkins-csi, line 99: A: I would define the URL path items as you would Query params here.  Allows us to more cleanly tie the "group name" to the source properties file name.

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -0,0 +1,172 @@
...
+connector=smallrye-kafka
+topic=ciTopic
+auth-token=auth-secret
+----
+
+NOTE: The `topic` property is overwritten when the environment specific value `ciTopic`.
+
+==== Mounting Properties
+In order for the store to read the configurations, the properties files will need to be placed into a https://kubernetes.io/docs/concepts/storage/persistent-volumes/#lifecycle-of-a-volume-and-claimpersistent[Persistent Volume,role=external,window=_blank] (PV)
+mounted to the configuration store kubernetes resource. It is assumed a PV will meet most projects needs and is the most widely accepted
```
ewilkins-csi, line 51: S: We've got a fine line to walk here between convenience/comprehensiveness in the antora docs, and confining chart-specific details to the chart documentation.  I think what we have is a pretty good split, but I think we could pull it up even a little higher here. For instance instead of getting into "Persistent Volumes" specifically, we can just say the high-level goal:  Attaching the properties to the local file system of the service through a mounted volume.   I don't have great specifics here, but maybe the key question is: Could someone with very little Kubernetes/Helm knowledge read the paragraph and get good understanding of _what's_ happening and why (even if they don't fully understand the "how").

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -0,0 +1,172 @@
...
+NOTE: Multiple configuration properties files can be stored at either URI.
+
+Since it is common practice to define separate Helm chart values for each environment of a project's development lifecycle, it is 
+encouraged to define one shared base URI and respective environment specific URIs, each housing the relevant overrides.
+
+The following Helm chart configuration demonstrates the URI specifications for an example CI deployment:
+[source,yaml]
+----
+  env:
+    BASE_PROPERTY: <URI housing base property configurations>
```
ewilkins-csi, line 70: S: This is an example of the type of detail that I think is maybe too much for these docs.  There are two risks with this example: 1) it muddies the waters of the overall explanation of the feature and generally how it works, and 2) it's very likely to fall out of date as the chart changes.

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -0,0 +1,172 @@
...
+enterprise applications. For smaller applications looking for a more light weight but still robust storage option, the configuration store 
+can be set to use https://github.com/TechnologyBrewery/krausening[Krausening,role=external,window=_blank]. The configuration store can also be set up to use in memory storage, this is best
+for quick local testing.
+
+The default property storage is Krausening. To change it to use one of the other options, set the environment
+variable `STORAGE_CLASS` to the desired value. The values currently available are `krausening`, `vault`, and `inMemory`.
+
+=== Usage
+After the configurations are ingested and reconciled within the store, the next step is to access them within a project using one of 
+the two available methods: the API endpoint or kubernetes resource injection. See the applicable sections below for more details on
```
ewilkins-csi, line 85: A: Let's make sure we've capitalized proper nouns like _"Kubernetes"_ throughout the page.

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -0,0 +1,172 @@
...
+[source,json]
+----
+{
+  "groupName": "messaging",
+  "name": "topic",
+  "value": "ciTopic"
+} 
+----
+
+NOTE: The name of the configuration service may vary depending on the projects helm chart configuration, by default it is
```
nartieri, line 123: A: missing apostrophe - project's

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -0,0 +1,172 @@
...
+`http://configuration-store.config-store-ns.svc:8083`
+
+==== Resource Injection
```
nartieri, line 126: A: Should this be _Kubernetes_ Resource Injection to match what we called it above?

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -0,0 +1,172 @@
...
+The configuration store provisions a https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/[kubernetes admission webhook,role=external,window=_blank] as part of its deployment. This webhook is used for injecting property values
+from the configuration store directly into kubernetes resources as they deployed to the cluster.
```
nartieri, line 128: A: recommend minor edit to "...as they _are_ deployed..."

?: Something that struck me as I was reading through this page is we give examples of using what we have built, but don't necessarily give examples of when to use them, or why you would choose one approach over another. You can tell me to pound sand if that is not something we typically do and is why we did not do it here. But I do wonder if providing some help to the user would be useful. E.g., when would they want to use the REST endpoint/API vice Kubernetes Resource Injection; when should they use Krausening vs Vault, and maybe some others. Anyway, it is just a bit of food for thought.

nartieri (changes requested):
Added some substantive and administrative feedback. Happy to discuss as needed.

docs/modules/ROOT/pages/configuration-store.adoc
carter-cundiff: Looking at the [krausening related section for this](https://github.com/TechnologyBrewery/krausening/tree/dev/krausening/), while it does explain how the extension works, I think it introduces too many unrelated details. Specifically the stuff the env var `KRAUSENING_EXTENSIONS` and the code block for retrieving the properties.

Would rather just keep a straightforward example here that just focus on the properties files.

docs/modules/ROOT/pages/configuration-store.adoc
carter-cundiff: Update hyperlinks to flow with sentence

docs/modules/ROOT/pages/configuration-store.adoc
carter-cundiff: Refactored

docs/modules/ROOT/pages/configuration-store.adoc
carter-cundiff: Brought up to higher level by removing kubernetes specific wording.

docs/modules/ROOT/pages/configuration-store.adoc
carter-cundiff: Moved technical portion to helm chart README

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -0,0 +1,172 @@
...
+connector=smallrye-kafka
+topic=baseTopic
+----
+This file defines two properties, `connector` and `topic`, each with the group name `messaging`.
+
+===== Environment Specific Overrides
+In addition to the base properties file introduced above, one can optionally specify a secondary properties file that houses environment
+specific configurations. These configurations will then used to override their respective base configurations.
+
+Continuing with the example above, suppose one defines an additional properties file named `messaging.properties` like the following:
```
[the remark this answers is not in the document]
carter-cundiff: In theory they would need to be the same name but in different directories for the environment ovverride to work. This is no longer applicable though as I removed the property file names from this section to make it less confusing.

docs/modules/ROOT/pages/configuration-store.adoc
carter-cundiff: Fixed

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -0,0 +1,176 @@
...
+connector=smallrye-kafka
+topic=baseTopic
+----
+This file defines two configuration properties: `connector` and `topic`.
+
+TIP: For sensitive configuration values, Krausening supports https://github.com/TechnologyBrewery/krausening/blob/dev/krausening/README.md#krausening-in-four-pints-leveraging-jasypt-for-encryptingdecrypting-properties[encrypting and decrypting properties,role=external,window=_blank]
+
+===== Environment Specific Overrides
+In addition to the properties file introduced above, one can optionally specify a secondary properties file that houses environment
+specific configurations. These configurations will then used to override their respective base configurations.
```
ewilkins-csi, line 30: A:
```suggestion
specific configurations. These configurations will override their respective base configurations.
```

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -0,0 +1,176 @@
...
+----
+connector=smallrye-kafka
+topic=ciTopic
+auth-token=auth-secret
+----
+
+NOTE: The `topic` property is overwritten when the environment specific value `ciTopic`.
+
+==== Deploying the Configuration Store
+The https://github.com/boozallen/aissemble/blob/{git-tree}/extensions/extensions-helm/aissemble-configuration-store-chart/README.md[configuration store helm chart,role=external,window=_blank] is used for easily deploying it to a Kubernetes cluster. As part of the deployment, one must
```
ewilkins-csi, line 50:
```suggestion
The https://github.com/boozallen/aissemble/blob/{git-tree}/extensions/extensions-helm/aissemble-configuration-store-chart/README.md[Configuration Store Helm chart,role=external,window=_blank] can easily deploy the store to a Kubernetes cluster. As part of the deployment, one must
```

docs/modules/ROOT/pages/configuration-store.adoc
carter-cundiff: The backend storage section gives a little insight into why you would want to use the different options.

In terms of the REST Endpoint vs kubernetes injection, I think the injection is the preferred method and can be used in replacement of the REST Endpoint in most use cases. However, at the end of the day it really comes down to user preference if they'd rather use the endpoint.

docs/modules/ROOT/pages/configuration-store.adoc
carter-cundiff: Updated

docs/modules/ROOT/pages/configuration-store.adoc
ewilkins-csi: Not sure if we're capitalizing "Configuration Store". I'm fine with either, but Helm should be capitalized.

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -0,0 +1,176 @@
...
+connector=smallrye-kafka
+topic=ciTopic
+auth-token=auth-secret
+----
+
+NOTE: The `topic` property is overwritten when the environment specific value `ciTopic`.
+
+=== Deploying the Configuration Store
+The https://github.com/boozallen/aissemble/blob/{git-tree}/extensions/extensions-helm/aissemble-configuration-store-chart/README.md[configuration store helm chart,role=external,window=_blank] is used for easily deploying it to a Kubernetes cluster. As part of the deployment, one must
+<<mounting-properties-files,mount the properties files>> to the Kubernetes resource allowing the configuration store to access the custom properties. Additionally,
```
ewilkins-csi, line 51: A:
```suggestion
provide the store access to the custom configurations by <<mounting-properties-files,mount the properties files>> to the Kubernetes resource. Additionally,
```

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -0,0 +1,176 @@
...
+the <<backend-storage-types,backend storage type>> can be adjusted to fit a project's needs.
+
+==== Mounting Properties Files
+Properties files are attached to the local file system of the configuration store within Kubernetes through a mounted volume. To mount
+a volume to the configuration store, one must set the helm chart's https://github.com/boozallen/aissemble/blob/{git-tree}/extensions/extensions-helm/aissemble-configuration-store-chart/README.md#persistent-volume-properties[persistent volume properties,role=external,window=_blank]. Then the helm chart's https://github.com/boozallen/aissemble/blob/{git-tree}/extensions/extensions-helm/aissemble-configuration-store-chart/README.md#environment-variables[environment 
+variables,role=external,window=_blank] must be updated with the corresponding URI(s) to the properties files within that volume. It is required to specify
```
ewilkins-csi, line 57: A: I think it might be better to use the word "location" over "URI", since (at least for now) we are _assuming_ the URI will be a local `file://` URI.

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -0,0 +1,176 @@
...
+secondary URI for properties files that house the environment specific configurations. 
+
+NOTE: Multiple configuration properties files can be stored at either URI.
+
+=== Backend Storage Types
+By default, the configuration store uses Krausening for it backend property storage. This provides a more light weight but still robust 
+storage option perfect for smaller applications.
+
+For enterprise applications that have more robust requirements, it is recommended to utilize https://developer.hashicorp.com/vault[Vault,role=external,window=_blank]. Additionally, the configuration 
+store can be set up to use in memory storage, this is best for quick local testing.
```
ewilkins-csi, line 68:
```suggestion
store can be set up to use in-memory storage &mdash; this is best for unit testing.
```

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -0,0 +1,176 @@
...
+
+Given an example properties file named `messaging.properties`:
+[source,properties]
+----
+connector=smallrye-kafka
+----
+
+The property name would be `connector` and group name would be `messaging`.
+
+=== REST Endpoint
```
ewilkins-csi, line 87: Q: Do we have a client API that we want to reference here? (Can't recall if we created one yet.)

carter-cundiff: At this time we do not

docs/modules/ROOT/pages/configuration-store.adoc
carter-cundiff: Capitalized all helms and config stores

nartieri (approved):

extensions/extensions-helm/aissemble-jenkins-chart/values.yaml
```diff
@@ -1,10 +0,0 @@
-# Default values for aissemble-jenkins.
```
carter-cundiff, line 1: I: Delete file mistakenly committed to dev

foundation/foundation-configuration-store/README.md
```diff
@@ -19,30 +19,4 @@ AWS_SECRET_ACCESS_KEY=base-secret-access-key
 ```
 
 ### Handling Load Status
-The `ConfigLoader` leverages `load-status` to detect whether properties were already fully loaded or only partially loaded due to an interrupted process. The code logic ensures that a `fully-loaded` status is set when properties are loaded successfully. This status is then used to skip loading if the properties were successfully loaded previously, thus avoiding unnecessary reloading during deployment refreshes. For details on this behavior, refer to the implementation code and comments in the `ConfigLoader` class.
-
-### Inject Configuration Value
```
carter-cundiff, line 24: I: Moved content to antora docs

foundation/foundation-mda/src/main/resources/templates/deployment/configuration-store/v2/configuration-store.values.yaml.vm
```diff
@@ -21,7 +21,6 @@ aissemble-configuration-store-chart:
       supplementalVolumeMounts:
         - name: configurations
           mountPath: "/configurations"
-          readOnly: true
```
cwoods-cpointe, line 24: q: do we not we want to keep the original PVC as read only?

carter-cundiff: Ideally yes, but it introduces a bug with argoCD (#309). So this is a temp fix for MVP until we get time to further investigate the issue.

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/ManualActionNotificationService.java
```diff
@@ -944,7 +961,7 @@ private NotificationParams configureNotification(File optionalRoot, String profi
         params.setPomFilePath(deployDir + File.separator + "pom.xml");
         params.setProfileConfiguration("<profile>" + profile + "</profile>");
         params.setPomFile(new File(params.getPomFilePath()));
-        params.setKey(getMessageKey(params.getPomFilePath(), "execution", appName));
```
ewilkins-csi, line 947: A/Q: Feels like it would be more consistent to have the relative path to the POM file here.  Can we not relativize against the root like we have elsewhere? (Not sure why it's called "optionalRoot" given the code just uses it directly for `setPomFilePath`.)

ewilkins-csi (approved):

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/ManualActionNotificationService.java
carter-cundiff: Update to set key with relative path as part of this function. Also update `optionalRoot` to `rootDir`.

PR #332, opened 2024-09-12
#3 Added aiSSEMBLE GitHub Action release job

jacksondelametter:
This is a PR to add the aiSSEMBLE GitHub Action release job. This action mimics the Jenkins release job found here https://github.boozallencsn.com/aissemble/aissemble/blob/fix-ci/devops/JenkinsfileRelease.groovy

The release job aims to do the following

- Parses the inputted version to get its major, minor, and patch counterparts
- Executes a pre release update to the release branch's Antora documentation 
- Cuts a release of the release branch
- Executes a post release update to the release branch's Antora documentation

### DIFF
--- a/.github/actions/split_version/action.yaml
+++ b/.github/actions/split_version/action.yaml
@@ -1,0 +1,44 @@
+name: 'Split Version'
+description: 'Splits the inputed version into major, minor, and patch components'
+inputs:
+  version: 
+    required: true
+  release-branch:
+    required: true
+  existing-version:
+    required: true
+outputs:
+  major-minor-version:
+    value: ${{ steps.split.outputs.major-minor-version }}
+  patch-version:
+    value: ${{ steps.split.outputs.patch-version }}
+  is-pre-release:
+    value: ${{ steps.split.outputs.is-pre-release }}
+  release-branch:
+    value: ${{ steps.split.outputs.release-branch }}
+  existing-version:
+    value: ${{ steps.split.outputs.existing-version }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Split
+      id: split
+      shell: bash
+      # Gets minor version 1.9.0-rc1 -> 1.9 
+      # Gets patch version 1.9.0-rc1 -> 1.9.0
+      # Determines if release is a pre release candidate
+      # Gets the release branch inputs.release-branch != "" ? inputs.release-branch : aissemble-release-$mino_version 
+      # Gets the existing version inputs.existing-version != "" ? inputs.existing-version : $patch_version-SNAPSHOT 
+      run: |
+        major=$(echo ${{ inputs.version }} | cut -d "." -f 1) 
+        minor=$(echo ${{ inputs.version }} | cut -d "." -f 2) 
+        patch=$(echo ${{ inputs.version }} | cut -d "." -f 3 | cut -d "-" -f 1) 
+        minor_version=$major.$minor
+        patch_version=$major.$minor.$patch
+        echo "major-minor-version=$minor_version" >> $GITHUB_OUTPUT 
+        echo "patch-version=$patch_version" >> $GITHUB_OUTPUT
+        echo "is-pre-release=$( [ "${{ inputs.version }}" != "$patch_version" ] && echo true || echo false)" >> $GITHUB_OUTPUT
+        echo "release-branch=$( [ "${{ inputs.release-branch }}" != "" ] && echo "${{ inputs.release-branch }}" || echo "aissemble-release-$minor_version")" >> $GITHUB_OUTPUT
+        echo "existing-version=$( [ "${{ inputs.existing-version }}" != "" ] && echo "${{ inputs.existing-version }}" || echo "$patch_version-SNAPSHOT")" >> $GITHUB_OUTPUT
+
--- a/.github/actions/update_antora/action.yaml
+++ b/.github/actions/update_antora/action.yaml
@@ -1,0 +1,37 @@
+name: 'Update Antora docs'
+description: 'Updates the Antora docs'
+inputs:
+  version:  
+    required: true
+  major-minor-version:  
+    required: true
+  patch-version:  
+    required: true
+  is-pre-release:  
+    required: true
+  release-branch:  
+    required: true
+  custom-facet-old-version:  
+    required: true
+  custom-facet-new-version:  
+    required: true
+    
+runs:
+  using: "composite"
+  steps:
+    - name: Update
+      shell: bash
+      run: |
+        sed -i "s/^version: .*/version: ${{ inputs.patch-version }}/" docs/antora.yml
+        version=$( ${{ inputs.is-pre-release }} && echo ${{ inputs.version }} || echo ${{ inputs.major-minor-version }} )
+        sed -i "s/^display_version: .*/display_version: $version/" docs/antora.yml
+        sed -i "s/^prerelease: .*/prerelease: ${{ inputs.is-pre-release }}/" docs/antora.yml
+        sed -i "s/^    is-pre-release: .*/    is-pre-release: ${{ inputs.is-pre-release }}/" docs/antora.yml
+        find docs/modules/ROOT/attachments/ -type f -exec sed -i 's|${{ inputs.custom-facet-old-version }}|${{ inputs.custom-facet-new-version }}|g' {} +
+        git config --global user.email "aissemble@bah.com"
+        git config --global user.name "aiSSEMBLE Team"
+        git diff
+        git add docs/antora.yml docs/modules/ROOT/attachments/
+        git commit -S -m "Updating Antora documentation for version ${{ inputs.version }}"
+        git push origin HEAD:${{ inputs.release-branch }}
+
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,110 @@
+# This workflow will build a Java project with Maven, and cache/restore any dependencies to improve the workflow execution time
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-java-with-maven
+
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+name: Release aissemble
+
+on:
+  workflow_dispatch:
+    inputs:
+      releaseVersion:
+        description: "The version you want to release aiSSEMBLE on."
+        required: true
+        type: string
+      developmentVersion:
+        description: "The version that dev should be on after this release."
+        required: true
+        type: string
+      releaseBranch:
+        description: "OPTIONAL: The branch to use for running the release."
+        required: false
+        type: string
+      existingVersion:
+        description: "OPTIONAL: The existing aiSSEMBLE version. Defaults to patchVersion-SNAPSHOT"
+        required: false
+        type: string
+
+jobs:
+  build:
+
+    runs-on: arc-runner-set-aissemble
+    env:
+      DOCKER_CONFIG: /home/runner/.docker
+      RELEASE_VERSION: ${{ inputs.releaseVersion }}
+      NEXT_DEVELOPMENT_VERSION: ${{ inputs.developmentVersion }}
+      RELEASE_BRANCH: ${{ inputs.releaseBranch }}
+      EXISTING_VERSION: ${{ inputs.existingVersion }}
+
+    steps:
+      - name: Import GPG key
+        uses: crazy-max/ghaction-import-gpg@v6
+        with:
+          gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
+          passphrase: ${{ secrets.GPG_PRIVATE_KEY_PASSPHRASE }}
+      - name: Install git
+        run: |
+          sudo apt-get update
+          sudo apt install -y git
+      - uses: actions/checkout@v4
+      - name: Split version
+        id: split_version
+        uses: ./.github/actions/split_version
+        with:
+          version: ${{ env.RELEASE_VERSION }}
+          release-branch: ${{ env.RELEASE_BRANCH }}
+          existing-version: ${{ env.EXISTING_VERSION }}
+      - name: Print release parameters
+        run: | 
+          echo "Releasing with the following parameters"
+          echo "Release version: ${{ env.RELEASE_VERSION }}"
+          echo "Next Development version: ${{ env.NEXT_DEVELOPMENT_VERSION }}"
+          echo "Major minor version: ${{ steps.split_version.outputs.major-minor-version }}"
+          echo "Patch Version: ${{ steps.split_version.outputs.patch-version }}"
+          echo "Is pre release version: ${{ steps.split_version.outputs.is-pre-release }}"
+          echo "Release branch: ${{ steps.split_version.outputs.release-branch }}"
+          echo "Existing version: ${{ steps.split_version.outputs.existing-version }}"
+      - name: Checkout release branch
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ steps.split_version.outputs.release-branch }}
+      - name: Update Antora Docs
+        uses: ./.github/actions/update_antora
+        with:
+          version: ${{ env.RELEASE_VERSION }}
+          major-minor-version: ${{ steps.split_version.outputs.major-minor-version }} 
+          patch-version: ${{ steps.split_version.outputs.patch-version }}
+          is-pre-release: ${{ steps.split_version.outputs.is-pre-release }}
+          release-branch: ${{ steps.split_version.outputs.release-branch }}
+          custom-facet-old-version: ${{ steps.split_version.outputs.existing-version }}
+          custom-facet-new-version: ${{ env.RELEASE_VERSION }}
+      - name: Install dependencies
+        uses: ./.github/actions/install_dependencies
+        with:
+          docker-username: ${{ secrets.DOCKERHUB_USERNAME }}
+          docker-token: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Create settings.xml
+        run: |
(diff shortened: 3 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,140 @@
...
+        required: false
+        default: false
+        type: boolean
+
+jobs:
+  build:
+
+    runs-on: arc-runner-set-aissemble
+    env:
+      DOCKER_CONFIG: /home/runner/.docker
```
jacksondelametter, line 41: Using env variables to easily switch out using workflow_dispatch input values and hardcoded values for testing!

.github/workflows/release.yml
```diff
@@ -0,0 +1,140 @@
...
+        type: string
+      releaseBranch:
+        description: "OPTIONAL: The branch to use for running the release."
+        required: false
+        type: string
+      existingVersion:
+        description: "OPTIONAL: The existing aiSSEMBLE version. Defaults to patchVersion-SNAPSHOT"
+        required: false
+        type: string
+      dryRun:
```
jacksondelametter, line 30: Allows us to toggle running the release job in dryRun mode.

aaron-gary (approved):
Looks good to me, but I haven't done a release on aiSSEMBLE so be sure you get a second opinion from baseline.

.github/workflows/release.yml
```diff
@@ -0,0 +1,140 @@
...
+  build:
+
+    runs-on: arc-runner-set-aissemble
+    env:
+      DOCKER_CONFIG: /home/runner/.docker
+      releaseVersion: ${{ inputs.releaseVersion }}
+      developmentVersion: ${{ inputs.developmentVersion }}
+      releaseBranch: ${{ inputs.releaseBranch }}
+      existingVersion: ${{ inputs.existingVersion }}
+      dryRun: ${{ inputs.dryRun }}
```
carter-cundiff, line 46: A: Recommend making the env name ALL_CAPS if possible.

.github/workflows/release.yml
```diff
@@ -0,0 +1,140 @@
...
+          java-version: '11'
+          distribution: 'temurin'
+      - name: Create Docker Builder Config File
+        run: sudo touch /etc/buildkitd.toml
+      - name: Login to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - uses: actions/checkout@v4
```
carter-cundiff, line 87: Q: `- uses: actions/checkout@v4` this is used twice, is that needed?

.github/actions/split_version/action.yaml
```diff
@@ -0,0 +1,47 @@
...
+    description: 'Version to split'
+    required: true
+  release-branch:
+    description: 'Release branch'
+    required: true
+  existing-version:
+    description: 'Existing version'
+    required: true
+outputs:
+  minor-version:
```
carter-cundiff, line 14: S: Would update this variable to be `major-minor-version`, sort of misleading currently.

.github/actions/split_version/action.yaml
```diff
@@ -0,0 +1,47 @@
...
+    description: "Patch version"
+    value: ${{ steps.split.outputs.patch-version }}
+  is-pre-release:
+    description: "Is pre-release version"
+    value: ${{ steps.split.outputs.is-pre-release }}
+  release-branch:
+    description: "Is pre-release version"
+    value: ${{ steps.split.outputs.release-branch }}
+  existing-version:
+    description: "Is pre-release version"
```
carter-cundiff, line 27: S: Update descriptions

.github/actions/split_version/action.yaml
```diff
@@ -0,0 +1,47 @@
...
+      run: |
+        major=$(echo ${{ inputs.version }} | cut -d "." -f 1) 
+        minor=$(echo ${{ inputs.version }} | cut -d "." -f 2) 
+        patch=$(echo ${{ inputs.version }} | cut -d "." -f 3 | cut -d "-" -f 1)  
+        minor_version=$major.$minor
+        patch_version=$major.$minor.$patch
+        echo "minor-version=$minor_version" >> $GITHUB_OUTPUT 
+        echo "patch-version=$patch_version" >> $GITHUB_OUTPUT
+        echo "is-pre-release=$( [ "${{ inputs.version }}" != "$patch_version" ] && echo true || echo false)" >> $GITHUB_OUTPUT
+        echo "release-branch=$( [ "${{ inputs.releaseBranch }}" != "" ] && echo "${{ inputs.releaseBranch }}" || echo "aissemble-release-$minor_version")" >> $GITHUB_OUTPUT
```
carter-cundiff, line 45: Q: Have you verified this works having the `echo` and part of the `&&` conditional?

jacksondelametter: Yes this works!

.github/workflows/release.yml
```diff
@@ -0,0 +1,142 @@
...
+          minor-version: ${{ steps.split_version.outputs.minor-version }} 
+          patch-version: ${{ steps.split_version.outputs.patch-version }}
+          is-pre-release: ${{ steps.split_version.outputs.is-pre-release }}
+          release-branch: ${{ steps.split_version.outputs.release-branch }}
+          custom-facet-old-version: ${{ steps.split_version.outputs.existing-version }}
+          custom-facet-new-version: ${{ env.RELEASE_VERSION }}
+      - name: Create settings.xml
+        run: |
+          echo "<settings><servers><server><id>ossrh</id><username>${{ secrets.SONATYPE_CENTRAL_REPO_TOKEN_USER }}</username><password>${{ secrets.SONATYPE_CENTRAL_REPO_TOKEN_KEY }}</password></server><server><id>ghcr.io</id><username>${{ secrets.GHCR_IO_USERNAME }}</username><password>${{ secrets.GHCR_IO_TOKEN }}</password></server><server><id>pypi</id><username>${{ secrets.PYPI_USERNAME }}</username><password>${{ secrets.PYPI_TOKEN }}</password></server><server><id>dev-pypi</id><username>${{ secrets.TEST_PYPI_USERNAME }}</username><password>${{ secrets.TEST_PYPI_TOKEN }}</password></server> </servers></settings>" > $HOME/.m2/settings.xml
```
carter-cundiff, line 123: Q: Should we be using a differ pypi server for the release? I believe test pypi is only for our snapshots.

.github/workflows/release.yml
jacksondelametter: Yes it is needed, it is a bug with using custom actions. You have to checkout the current branch your running on for them to show up. The other `checkout` checks out the release branch.

.github/workflows/release.yml
jacksondelametter: Changed

.github/workflows/release.yml
carter-cundiff: Recommend grouping them together (if possible) and leaving a comment explaining that.

.github/workflows/release.yml
jacksondelametter: Its being uploaded to pypi not test pypi, I verified this.

.github/workflows/release.yml
jacksondelametter: I cant group them unfortunately but I can leave a comment

.github/workflows/release.yml
carter-cundiff: Safe to remove this then? `<id>dev-pypi</id><username>${{ secrets.TEST_PYPI_USERNAME }}</username><password>${{ secrets.TEST_PYPI_TOKEN }}</password>`?

jacksondelametter: I am not sure if it is safe to remove!

.github/actions/split_version/action.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - name: Split
+      id: split
+      shell: bash
+      run: |
+        git branch
+        major=$(echo ${{ inputs.version }} | cut -d "." -f 1) 
+        minor=$(echo ${{ inputs.version }} | cut -d "." -f 2) 
+        patch=$(echo ${{ inputs.version }} | cut -d "." -f 3 | cut -d "-" -f 1)  
+        minor_version=$major.$minor
+        patch_version=$major.$minor.$patch
```
carter-cundiff, line 42: A: Would be helpful to have comments explaining what these vars are given inputs, for example from the old release script:
```
//grab second group (first embedded capture) in first match: 1.4.2-rc.3 -> 1.4.2

//grab second group (first embedded capture) in first match: 1.4.2 -> 1.4
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,142 @@
...
+        uses: ./.github/actions/split_version
+        with:
+          version: ${{ env.RELEASE_VERSION }}
+          release-branch: ${{ env.RELEASE_BRANCH }}
+          existing-version: ${{ env.EXISTING_VERSION }}
+      - name: Print release parameters
+        run: | 
+          echo "Releasing with the following parameters"
+          echo "Release version: ${{ env.RELEASE_VERSION }}"
+          echo "Development version: ${{ env.DEVELOPMENT_VERSION }}"
```
carter-cundiff, line 101: Optional: Change `Development version` -> `Next Development Version` and `env.DEVELOPMENT_VERSION` -> `env.NEXT_DEVELOPMENT_VERSION`

.github/actions/split_version/action.yaml
jacksondelametter: Added comments to better explain

.github/workflows/release.yml
jacksondelametter: Changed

.github/workflows/release.yml
```diff
@@ -0,0 +1,143 @@
...
+        description: "OPTIONAL: The existing aiSSEMBLE version. Defaults to patchVersion-SNAPSHOT"
+        required: false
+        type: string
+      dryRun:
+        description: "OPTIONAL: Used to run the release job as a dry run, skipping the deploy phase"
+        required: false
+        default: false
+        type: boolean
+  push:
+    branches: [ "3-add-release" ]
```
carter-cundiff, line 36: This will be removed before merge? Does it need a different value?

.github/workflows/release.yml
ewilkins-csi: I don't think you'll need a double checkout once this is merged to dev.  Do you have a link to the info you found saying you would need that?

.github/actions/split_version/action.yaml
```diff
@@ -0,0 +1,52 @@
...
+    description: "Major-minor version"
+    value: ${{ steps.split.outputs.major-minor-version }}
+  patch-version:
+    description: "Patch version"
+    value: ${{ steps.split.outputs.patch-version }}
+  is-pre-release:
+    description: "Is pre-release version"
+    value: ${{ steps.split.outputs.is-pre-release }}
+  release-branch:
+    description: "Is pre-release version"
```
ewilkins-csi, line 24: A: Update description

.github/actions/split_version/action.yaml
```diff
@@ -0,0 +1,52 @@
...
+    required: true
+  release-branch:
+    description: 'Release branch'
+    required: true
+  existing-version:
+    description: 'Existing version'
+    required: true
+outputs:
+  major-minor-version:
+    description: "Major-minor version"
```
ewilkins-csi, line 15: S: In general these descriptions are just restating the name of the output which doesn't seem helpful.  We should either make them more explanatory or just omit them if we can.

.github/actions/update_antora/action.yaml
```diff
@@ -0,0 +1,44 @@
...
+  patch-version:  
+    description: 'Patch version'
+    required: true
+  is-pre-release:  
+    description: 'If the release is a pre-release version'
+    required: true
+  release-branch:  
+    description: 'Release branch'
+    required: true
+  custom-facet-old-version:
```
ewilkins-csi, line 19: Q: Why are these inputs instead of just being derived from `version`?

.github/actions/update_antora/action.yaml
```diff
@@ -0,0 +1,44 @@
...
+        version=$( ${{ inputs.is-pre-release }} && echo ${{ inputs.version }} || echo ${{ inputs.major-minor-version }} )
+        sed -i "s/^display_version: .*/display_version: $version/" docs/antora.yml
+        sed -i "s/^prerelease: .*/prerelease: ${{ inputs.is-pre-release }}/" docs/antora.yml
+        sed -i "s/^    is-pre-release: .*/    is-pre-release: ${{ inputs.is-pre-release }}/" docs/antora.yml
+        find docs/modules/ROOT/attachments/ -type f -exec sed -i 's|${{ inputs.custom-facet-old-version }}|${{ inputs.custom-facet-new-version }}|g' {} +
+        git config --global user.email "aissemble@bah.com"
+        git config --global user.name "aiSSEMBLE Team"
+        git diff
+        git add .
+        git commit -S -m "Updating scripts and templates to ${{ inputs.version }}"
```
ewilkins-csi, line 42: A: I'd make the commit message more accurate. This is likely a carryover from when we updated more stuff manually instead of through the build process.

.github/workflows/release.yml
ewilkins-csi: C: This _needs_ to be removed before merging.  This workflow should be on dispatch only.

.github/workflows/release.yml
```diff
@@ -0,0 +1,143 @@
...
+  build:
+
+    runs-on: arc-runner-set-aissemble
+    env:
+      DOCKER_CONFIG: /home/runner/.docker
+      RELEASE_VERSION: ${{ inputs.releaseVersion || '1.9.0-rc2' }}
+      NEXT_DEVELOPMENT_VERSION: ${{ inputs.developmentVersion || '1.10.0-SNAPSHOT' }}
+      RELEASE_BRANCH: ${{ inputs.releaseBranch }}
+      EXISTING_VERSION: ${{ inputs.existingVersion }}
+      DRY_RUN: ${{ inputs.dryRun || true }}
```
ewilkins-csi, line 48: C: We should not be including these defaults at all. It just introduces risk/confusion.  If this script is updated/worked on in the future and such mechanisms are helpful, the author can add this stuff in.  However, I would argue that we should be testing by filling in the parameters manually anyway.

.github/workflows/release.yml
```diff
@@ -0,0 +1,143 @@
...
+        with:
+          default: 3.11.4
+      - name: Install Poetry
+        uses: snok/install-poetry@v1
+      - name: Install Helm
+        run: |
+          curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3
+          chmod 700 get_helm.sh
+          ./get_helm.sh
+      - name: Install Helm Unittest Plugin
```
ewilkins-csi, line 73: S: This is just a copy of what we're doing in `build.yaml` right? Strongly suggest we pull these out into a common place if so.  We're already introducing the complexity of composite actions, and this is the core use case for them.

.github/workflows/release.yml
ewilkins-csi: yes it should be

.github/workflows/release.yml
```diff
@@ -0,0 +1,143 @@
...
+          custom-facet-new-version: ${{ env.RELEASE_VERSION }}
+      - name: Create settings.xml
+        run: |
+          echo "<settings><servers><server><id>ossrh</id><username>${{ secrets.SONATYPE_CENTRAL_REPO_TOKEN_USER }}</username><password>${{ secrets.SONATYPE_CENTRAL_REPO_TOKEN_KEY }}</password></server><server><id>ghcr.io</id><username>${{ secrets.GHCR_IO_USERNAME }}</username><password>${{ secrets.GHCR_IO_TOKEN }}</password></server><server><id>pypi</id><username>${{ secrets.PYPI_USERNAME }}</username><password>${{ secrets.PYPI_TOKEN }}</password></server><server><id>dev-pypi</id><username>${{ secrets.TEST_PYPI_USERNAME }}</username><password>${{ secrets.TEST_PYPI_TOKEN }}</password></server> </servers></settings>" > $HOME/.m2/settings.xml
+      - name: Release aiSSEMBLE
+        run: | 
+          ./mvnw release:clean release:prepare release:perform -s $HOME/.m2/settings.xml -U -B -Pci,gh-build -DdryRun=${{ env.DRY_RUN }} -DpushChanges=true -DreleaseVersion=${{ env.RELEASE_VERSION }} -DdevelopmentVersion=${{ env.NEXT_DEVELOPMENT_VERSION }} -Dgpg.keyname=aissemble@bah.com -DsignTag=true -Dmaven.build.cache.enabled=false
```
ewilkins-csi, line 127: S: IMO the dry-run isn't going to be used enough to warrant having it checked in to dev and it just adds noise to the parameters. Esp since we aren't using it for things like the git commits at earlier steps.

.github/workflows/release.yml
jacksondelametter: Yes, this is only for testing purposes. It will be removed before being merged.

.github/actions/update_antora/action.yaml
jacksondelametter: When updating the custom facet schemas, what we change differs between the pre release Antora docs update and the post release.

.github/workflows/release.yml
jacksondelametter: This is only for testing purposes. This will be deleted before being merged.

.github/workflows/release.yml
jacksondelametter: Ok this is actually not a bug. It looks like if you dont do a initial checkout, the local aissemble repository is actually not present in the runner. By doing an initial checkout, the repository is present and the branch being ran on is checked out.

.github/workflows/release.yml
jacksondelametter: Removed

.github/actions/split_version/action.yaml
jacksondelametter: Removed descriptions

.github/actions/update_antora/action.yaml
jacksondelametter: Changed to "Updating Antora documentation to version ..."

.github/workflows/release.yml
jacksondelametter: Added instal dependencies composite action.

.github/workflows/release.yml
jacksondelametter: Removed

Issue #310, opened 2024-08-29
BUG: Configuration Store: `"aissemble-configuration-store-webhook-certs" not found` error when disabling the webhook


csun-cpointe:
## Description
Deploying the configuration-store app using argocd with the `webhook.enable` set to `false`, the configuration store app status becomes `Degraded` and the `configuration-store` pod fails with the `MountVolume.SetUp failed for volume "certs" : secret "aissemble-configuration-store-webhook-certs" not found` error
[
![Screenshot 2024-08-29 at 4 12 42 PM](https://github.com/user-attachments/assets/86522af3-92e0-41d8-bb4b-c35d1404089d)
](url)
![Screenshot 2024-08-29 at 4 03 50 PM](https://github.com/user-attachments/assets/f9d8ee11-6bfd-463b-ae61-dcc2691e8415)

## Steps to Reproduce
Clear, specific, and detailed steps taken to enable reproduction of the bug for investigation.
- Create a downstream project with baseline version 1.9.0-SNAPSHOT
```
mvn archetype:generate -U -DarchetypeGroupId=com.boozallen.aissemble \
  -DarchetypeArtifactId=foundation-archetype \
  -DarchetypeVersion=1.9.0-SNAPSHOT \
  -DgroupId=com.test \
  -DartifactId=test-310 \
  -DprojectGitUrl=test.url \
  -DprojectName=test-310 \
  && cd test-310
```
- Add the [SparkPipeline.json](https://github.com/user-attachments/files/16781488/SparkPipeline.json) file to the test-defect-pipeline-models/src/main/resources/pipelines directory

- Add to the fermenter-mda plugin executions in test-defect-deploy/pom.xml
```
<execution>
    <id>configuration-store</id>
    <phase>generate-sources</phase>
    <goals>
        <goal>generate-sources</goal>
    </goals>
    <configuration>
        <basePackage>com.boozallen.aissemble.test</basePackage>
        <profile>configuration-store-deploy-v2</profile>
        <!-- The property variables below are passed to the Generation Context and utilized
             to customize the deployment artifacts. -->
        <propertyVariables>
            <appName>configuration-store</appName>
        </propertyVariables>
    </configuration>
</execution>
```
- Run mvn clean install until all the manual actions are complete
- Once the manual actions are complete, run mvn clean install -Dmaven.build.cache.skipCache=true once to get any remaining manual actions

[...]

PR #336, opened 2024-09-13
#310 enable quarkus flexible configuration

csun-cpointe:


extensions/extensions-helm/aissemble-configuration-store-chart/templates/webhook-serviceaccount.yaml
```diff
@@ -1,4 +1,4 @@
-{{ if .Values.webhook.serviceAccount.create }}
+{{ if and .Values.webhook.enable .Values.webhook.serviceAccount.create }}
```
aaron-gary, line 1: Q: I'm not familiar with this syntax.  It looks like the statement has the "and" in the wrong order, but it may just be my inexperience with helm(?)

aaron-gary (approved):

extensions/extensions-helm/aissemble-configuration-store-chart/templates/webhook-serviceaccount.yaml
csun-cpointe: yeah.. helm template functions follow the syntax `functionName arg1 arg2...` and operators are implemented as functions.

Issue #330, opened 2024-09-12
Feature: aiSSEMBLE-specific Hive MySQL image should be deprecated


ewilkins-csi:
## Description
With the introduction of the v2 Spark Infrastructure chart, we've officially moved the default Hive Metastore chart to use the community MySql chart/image.  As such, the aiSSEMBLE-specific image should be deprecated and removed.

## DOD
- [x] The aiSSEMBLE build no longer produces the `aissemble-hive-mysql` image
- [x] Generating the v1 Hive DB chart should produce a values file that uses the 1.8.1 version of the `aissemble-hive-mysql` image
- [x] Upgrading with the v1 Hive DB chart should not update the image tag to version 1.9.0

## Test Script
1. **Final Test:** Verify the **1.9.0-SNAPSHOT** tag of the [`aissemble-hive-service` image](https://github.com/boozallen/aissemble/pkgs/container/aissemble-hive-service) has been published more recently than the same tag for the [`aissemble-hive-mysql` image](https://github.com/boozallen/aissemble/pkgs/container/aissemble-hive-mysql)
2. **OTS:** Verify the module **aiSSEMBLE::Extensions::Docker::Hive MySql** does not appear in the build summary
3. Create a project from the current archetype
   ```sh
   mvn archetype:generate -B '-DarchetypeGroupId=com.boozallen.aissemble' \
                          '-DarchetypeArtifactId=foundation-archetype' \
                          '-DarchetypeVersion=1.9.0-SNAPSHOT' \
                          '-DgroupId=org.test' \
                          '-Dpackage=org.test' \
                          '-DprojectGitUrl=test.org/test.git' \
                          '-DprojectName=Test Project' \
                          '-DartifactId=ots-330' \
                          '-DdockerProjectRepositoryUrl=test-registry/' \
   && cd ots-330
   ```
4. Add the old v1 Hive DB profile to the fermenter-mda executions in _ots-330-deploy/pom.xml_
   ```xml
                       <execution>
                           <id>hive-metastore-db</id>
                           <phase>generate-sources</phase>
                           <goals>
                               <goal>generate-sources</goal>
                           </goals>
                           <configuration>
                               <basePackage>org.test</basePackage>
                               <profile>hive-metastore-db-deploy</profile>
                               <propertyVariables>
                                   <appName>hive-metastore-db</appName>
                               </propertyVariables>
                           </configuration>
                       </execution>
   ```
5. Run the build: `mvn clean install`
[...]

PR #341, opened 2024-09-16
[#330] deprecate hive mysql image

ewilkins-csi:
Removes the aiSSEMBLE Hive DB image from the build entirely and sets the old hive-metastore-db v1 chart to use the 1.8.1 version of the image. Also removed this chart from the v1 Baton migration so it would not be updated beyond 1.8.1.


pom.xml
```diff
@@ -6,6 +6,7 @@
         <groupId>com.boozallen.aissemble</groupId>
         <artifactId>aissemble-parent</artifactId>
         <version>4</version>
+        <relativePath/>
```
ewilkins-csi, line 9: I: Just fixes a Maven warning about an invalid relative path. See https://maven.apache.org/ref/3.8.1/maven-model/maven.html#class_parent

extensions/extensions-docker/pom.xml
```diff
@@ -59,7 +59,7 @@
     <modules>
         <module>aissemble-quarkus</module>
         <module>aissemble-fastapi</module>
-        <module>aissemble-hive</module>
+        <module>aissemble-hive-service</module>
```
ewilkins-csi, line 62: I: Since there aren't two hive images now, the structure was flattened.

cwoods-cpointe (approved):

carter-cundiff (approved):

PR #369, opened 2024-09-26
Java 17 upgrade

ewilkins-csi:
Now that 1.9 has been cut and dev is on 1.10 we can merge in the Java upgrade branch.  This will allow us to work directly off of dev instead of managing a feature branch.


carter-cundiff (approved):

carter-cundiff:
Q: Will we need to upgrade the CI jobs to use jdk 17?

ewilkins-csi:
> Q: Will we need to upgrade the CI jobs to use jdk 17?

Addressed in the latest commit

Issue #152, opened 2024-06-14
BUG: Records with the same package as a data pipeline cannot be used as inbound/outbound types


ewilkins-csi:
## Description
When the package name of a record matches the package name of a Spark data pipeline, the record cannot be used as an inbound or outbound record type because a NPE is thrown.  It's unclear if this would happen with PySpark pipelines.

## Steps to Reproduce
1. Create an aissemble project
2. Add a record MetaModel with the package `org.test`
3. Add a `data-delivery-spark` pipeline MetaModel with the package `org.test` that sets the `inbound` or `outbound` `recordType` of a step to the record created in step 2
4. Run the build and a NPE is thrown during `generate-sources`

## Expected Behavior
The build should successfully generate sources via Fermenter.

## Actual Behavior
A `NullPointerException` is thrown.

## Definition of Done
- Reproduce issue as mentioned in Step to Reproduce section in above.
- Fix Null pointer Exception when Record package matches with Pipeline Package when specifying inbound or outbound Record Type.
- Run Integration test and archetype test to make sure successful run.
- Follow Step to Reproduce and Verify There's no NullPointerException Thrown.

## Testing Steps
- Pull latest dev branch and build using **mvn clean install**
- Create an aissemble project using following command
```
mvn archetype:generate -U -DarchetypeGroupId=com.boozallen.aissemble \
  -DarchetypeArtifactId=foundation-archetype \
  -DarchetypeVersion=1.10.0-SNAPSHOT\
  -DgroupId=com.boozallen.aissemble.test \
  -DartifactId=issue-152\
  -DprojectGitUrl=git.example.com/issue-152\
  -DprojectName="issue-152”
```
- Create a record MetaModel with the package called org.test.aissemble.data.profiling (Contact James Choi for example Record)
- Create Dictonary MetaModel (doesn't have to match any package and Contact James Choi for example Dictonary)
- Create a data-delivery-spark pipeline MetaModel with the package org.test.aissemble.data.profiling that sets the inbound or outbound recordType of a step to the record created in above (Contact James Choi for example pipeline)
- cd into downstream project.
- Add a data delivery spark pipeline, record and dictionary json files to an aissemble project by saving attached json into following respectively.
```
{downstream-project}/{downstream-project}-pipeline-models/src/main/resources/pipelines
{downstream-project}/{downstream-project}-pipeline-models/src/main/resources/records
{downstream-project}/{downstream-project}-pipeline-models/src/main/resources/dictionary
```
[...]

PR #378, opened 2024-09-27
#152 Records with the same package as a data pipeline cannot be used as inbound/outbound types

jaebchoi:
### Description
When the package name of a record matches the package name of a Spark data pipeline, the record cannot be used as an inbound or outbound record type because a NPE is thrown.

[Issue #152 ](https://github.com/boozallen/aissemble/issues/152)

### DIFF
--- a/foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/java/JavaStepDataRecordType.java
+++ b/foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/java/JavaStepDataRecordType.java
@@ -22,9 +22,6 @@ import com.boozallen.aiops.mda.metamodel.element.StepDataRecordType;
  */
 public class JavaStepDataRecordType extends BaseStepDataRecordTypeDecorator {
 
-    private AIOpsModelInstanceRepostory modelRepository = ModelInstanceRepositoryManager
-            .getMetamodelRepository(AIOpsModelInstanceRepostory.class);
-
     /**
      * {@inheritDoc}
      */
@@ -41,7 +38,7 @@ public class JavaStepDataRecordType extends BaseStepDataRecordTypeDecorator {
         String fullyQualifiedRecordType = null;
 
         String packageName = getPackage();
-        if (StringUtils.isNotBlank(packageName) && !packageName.equals(modelRepository.getBasePackage())) {
+        if (StringUtils.isNotBlank(packageName) ){
             fullyQualifiedRecordType = packageName + "." + getName();
         }
 


foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/java/JavaStep.java
```diff
@@ -453,6 +453,12 @@ public String getLowerCamelCaseName() {
     }
 
     private void addRecordImports(String importName) {
+        if(importName == null)
+        {
+            throw new GenerationException("Step '" + getName() + "' has outbound or inbound Record Type that have "
+                    + "null import name.");
+        }
+
```
jaebchoi, line 461: I: this is to have better error thrown than plain NPE

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/java/JavaStepDataRecordType.java
```diff
@@ -41,7 +38,7 @@ public String getFullyQualifiedType() {
         String fullyQualifiedRecordType = null;
 
         String packageName = getPackage();
-        if (StringUtils.isNotBlank(packageName) && !packageName.equals(modelRepository.getBasePackage())) {
+        if (StringUtils.isNotBlank(packageName) ){
```
jaebchoi, line 41: I: Not sure this is intended but it was initially there as part of 0.20.0 milestone to make recordtype functionality.

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/java/JavaStep.java
ewilkins-csi: Q: Did you see this happen in testing? I would have expected the model validation phase to throw an error before we got to this point.  Not bad to have this either way, but if we're missing something in validation it'd be good to fill that gap.

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/java/JavaStepDataRecordType.java
ewilkins-csi: The original thought process was probably that we don't need an import if it's in the same package.  Which _should_ still be true.  If we remove this and that results in an import, then the import statement would be harmless but also show up as unnecessary in most IDEs.

ewilkins-csi (approved):

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/java/JavaStepDataRecordType.java
jaebchoi: Interesting, thought it would alway match base model and pipeline package. Good to know, Thank you!

PR #381, opened 2024-09-30
331 dockerhub rate limit - Resolve 429 Too Many Requests Error 

colinpalmer-pro:
Resolves the `429 Too Many Requests - Server message: toomanyrequests: You have reached your pull rate limit` Docker Hub Error.

The following changes ensure we are using our docker credentials throughout our build process:
- Explicitly state `docker.io` repo in base images
  - This allows fabric8 to pick up the registry and create a config object. Without the registry specified, fabric8 doesn't create the config object
- Update fabric8 to use own `.docker` config location
  - fabric8 cleans up after its done so when using the default `.docker` config location, fabric8 removes the docker credentials. So any module that performs a pull outside of fabric8 was no longer authenticated

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -599,7 +599,7 @@
                                     </args>
                                     <buildx>
                                         <builderName>default</builderName>
-                                        <dockerStateDir>~/.docker</dockerStateDir>
+                                        <dockerStateDir>~/.docker/fabric8</dockerStateDir>
                                         <platforms>
                                             <platform>${docker.platforms}</platform>
                                         </platforms>
--- a/extensions/extensions-docker/aissemble-fastapi/src/main/resources/docker/Dockerfile
+++ b/extensions/extensions-docker/aissemble-fastapi/src/main/resources/docker/Dockerfile
@@ -1,6 +1,6 @@
 # Script for creating base FastAPI Docker image
 
-FROM python:3.11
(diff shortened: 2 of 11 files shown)


extensions/extensions-docker/aissemble-fastapi/src/main/resources/docker/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Script for creating base FastAPI Docker image
 
-FROM python:3.11
+FROM docker.io/python:3.11
```
colinpalmer-pro, line 4: i: explicitly stating the registry so fabric8 knows to authenticate to docker.io

build-parent/pom.xml
```diff
@@ -599,7 +599,7 @@
                                     </args>
                                     <buildx>
                                         <builderName>default</builderName>
-                                        <dockerStateDir>~/.docker</dockerStateDir>
+                                        <dockerStateDir>~/.docker/fabric8</dockerStateDir>
                                         <platforms>
```
colinpalmer-pro, line 603: i: providing a separate .docker directory so fabric8 uses the default location but does not override/delete the default `~.docker/config.json`

d-ryan-ashcraft:
Thanks for working through this @colinpalmer-pro!

ewilkins-csi (approved):

Issue #379, opened 2024-09-27
Feature: Add unit tests for pyproject.toml.vm Python version updates


meliz19:
## Description
This is a follow on to #366. In that ticket, I updated the Python version in the `pyproject.toml.vm` files. The purpose of this ticket is to write unit tests to verify the Python version updates in all of the `pyproject.toml.vm` files.

## DOD
- [x]  Write unit tests for the following files to verify the Python version updates
- [x] `foundation/foundation-mda/src/main/resources/templates/data-delivery-data-records/combined.pyproject.toml.vm`
- [x] `foundation/foundation-mda/src/main/resources/templates/data-delivery-data-records/separate.core.pyproject.toml.vm`
- [x] `foundation/foundation-mda/src/main/resources/templates/data-delivery-data-records/separate.spark.pyproject.toml.vm`
- [x] `foundation/foundation-mda/src/main/resources/templates/data-delivery-pyspark/pyproject.toml.vm`
- [x] `foundation/foundation-mda/src/main/resources/templates/general-mlflow/pyproject.toml.vm`
- [x] `foundation/foundation-mda/src/main/resources/templates/inference/pyproject.toml.vm`
- [x] `foundation/foundation-mda/src/main/resources/templates/sagemaker-training/pyproject.toml.vm`

## Test Strategy/Script
*Test 1: OTS*
- From the root project directory, run the following command. Verify it builds successfully. 
```
cd foundation/foundation-mda && ~/aissemble/mvnw clean install && cd -
```
*Test 2*
- Verify the unit tests passed in the CI build
## References/Additional Context
None

meliz19:
Completed DoD with @d-ryan-ashcraft

meliz19:
Completed OTS with @cwoods-cpointe @habibimoiz

chang-annie:
test steps are passing! CI build succeeded and all the unit tests passed.

PR #387, opened 2024-10-02
#379 Add unit tests to check pyproject.toml velocity template Python …

meliz19:
…version updates

### DIFF
--- a/foundation/foundation-mda/src/test/java/com/boozallen/aiops/mda/generator/PyprojectGenerationSteps.java
+++ b/foundation/foundation-mda/src/test/java/com/boozallen/aiops/mda/generator/PyprojectGenerationSteps.java
@@ -1,0 +1,119 @@
+package com.boozallen.aiops.mda.generator;/*-
+ * #%L
+ * aiSSEMBLE::Foundation::MDA
+ * %%
+ * Copyright (C) 2021 Booz Allen
+ * %%
+ * This software package is licensed under the Booz Allen Public License. All Rights Reserved.
+ * #L%
+ */
+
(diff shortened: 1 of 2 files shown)


foundation/foundation-mda/src/test/java/com/boozallen/aiops/mda/generator/PyprojectGenerationSteps.java
```diff
@@ -0,0 +1,119 @@
...
+    public void a_pipeline_with_step_type_using_the_implementation(String pipelineType, String pipelineStepType, String pipelineImplementation) throws IOException {
+        String pipelineName = pipelineType + "-pipeline";
+        String pipelineStepName = pipelineStepType + "-step";
+        
+        createAndSavePipeline(pipelineName, pipelineType, pipelineImplementation, pipeline -> {
+            try {
+                StepElement pipelineStep = createPipelineStep(pipelineStepName, pipelineStepType);
+                pipeline.addStep(pipelineStep);
+            } catch (Exception e) {
+                System.err.println("Exception: " + e.getMessage());
```
ewilkins-csi, line 68: S: Something like `logger.error("Failed to add new step to test pipeline", e)` will provide more useful output.  It will print both the exception message and the full stack trace, while also giving info about what we were attempting when the error occurred.

meliz19: Great point. I'll make the update. Thank you!

carter-cundiff (approved):

Issue #363, opened 2024-09-24
Feature: JDK 17 Upgrade Support Series - Semantic data (incl. data transform & encryption)


csun-cpointe:
## Description
In #133  we modified the build-parent to use JDK 17. When building aiSSEMBLE with the updated build-parent, some modules do not build successfully. This is a series of ticket to fix broken modules. This issue will focus on the semantic data (including data transform and data encryption) functionality

## DOD
Acceptance criteria required to realize the requested feature
The builds for following modules work:
- [x] `extensions-encryption`
- [x] `extensions-transform`
- [x] `aissemble-vault`
- [ ] Note the migration needed if there are template changes
- [ ] Smoke test success on the `extenions-encryption` module (all tests are run with spit module setup)
   - [x] AES encryption
   - [x] Custom dictionary
   - [x] Data validation
   - [ ] Simple persistent format

## Test Strategy/Script
- Create a downstream project using java 17
   ```
             mvn archetype:generate -DarchetypeGroupId=com.boozallen.aissemble \                  
                       -DarchetypeArtifactId=foundation-archetype \
                       -DarchetypeVersion=1.10.0-SNAPSHOT \
                       -DgroupId=org.test \
                       -DartifactId=test-363 \
                       -DprojectGitUrl=test.org/test-363.git \
                       -DprojectName="Test 363" && cd test-363

   ```
- Unzip the attached pipeline [resources.zip](https://github.com/user-attachments/files/17237632/resources.zip) and replace the `-pipeline-models/src/main/resources` folder:
- In the -shared/pom.xml file update to use split data record
     ```
                        <configuration>
                            <basePackage>com.boozallen</basePackage>
   -                        <profile>aissemble-data-records-combine-module</profile>
   +                        <profile>aissemble-data-records-separate-module</profile>
                        </configuration>
     ```
- Repeatedly run `mvn clean install -Dmaven.build.cache.skipCache` and resolve the manual actions until no more manual actions
**Note**: **Do not include** the data-access related modules
**Note**: Before include the `spark-pipeline` to the `pipelines` module, **update** the `-pipelines/spark-pipeline/pom.xml` data record dependency to use `-data-record-spark` module
```
        <dependency>
            <groupId>${project.groupId}</groupId>
-           <artifactId>test-363-data-records</artifactId>
+           <artifactId>test-363-data-records-spark</artifactId>
            <version>${project.version}</version>
        </dependency>
```
[...]

PR #396, opened 2024-10-04
#363 fix the spark-application build issue

csun-cpointe:
Specifying the Ivy user directory will override the Ivy property `ivy.default.ivy.user.dir`, which will also disallow the ivy cache to use `/home/runner/.ivy2/cache` on the runner for the local test. 

Since we only use the pod cache for the dev deployments as the we only set `storeCacheOnNode: true` for the spark-operator as part of `values-dev.yaml`, we will move the ivy cache directory to the spark application `-dev-values` file

### DIFF
--- a/foundation/foundation-mda/src/main/resources/templates/deployment/spark-operator/spark-application-dev-values.yaml.vm
+++ b/foundation/foundation-mda/src/main/resources/templates/deployment/spark-operator/spark-application-dev-values.yaml.vm
@@ -2,6 +2,7 @@ sparkApp:
     spec:
       image: "${projectName}-spark-worker-docker:latest"
       sparkConf:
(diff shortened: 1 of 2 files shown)


foundation/foundation-mda/src/main/resources/templates/deployment/spark-operator/spark-application-dev-values.yaml.vm
```diff
@@ -2,6 +2,7 @@ sparkApp:
     spec:
       image: "${projectName}-spark-worker-docker:latest"
       sparkConf:
+        spark.jars.ivy: "/opt/spark/.ivy2"
```
csun-cpointe, line 5: I: Migration to come just to unblock CI for now

Issue #400, opened 2024-10-08
TASK: Overwrite the ivy user directory in the spark-application chart


csun-cpointe:
## Description
For java 17 upgrade, we have upgraded the `aissemble-spark` base image, which has a different home directory than the `/opt/spark/` directory so to make sure Ivy cache can find the right directory (_Ivy cache is default to the environment's home directory. e.g.: ~/.ivy2_), we will need to override the ivy user directory by specifying the `spark.jars.ivy` directory in the chart. 

Currently, to fix the build issue, we have moved the `spark.jars.ivy` from the spark application chart to the dev values file to avoid the local test issue. However, since our local test is not using the docker container but only copy over the spark configuration from the yaml file to configure the spark session in the local environment, we should _not_ copy `spark.jars.ivy` directory for the spark tester setup but leave it as default so that it can still use the default home directory and when we deploying the spark application to different environment, we don't need to set `spark.jars.ivy` path in the values file.

## DOD
Acceptance criteria required to complete the work
- [ ] Move the `spark.jars.ivy` configuration to the aissemble-spark-application chart
- [ ] In python or java test code, avoid `spark.jars.ivy` configuration being setup in the spark

## Test Strategy/Script
How will this item be verified?
- Pull the latest `aissemble` repo

- **OTS Only:**
  - Since the `aissemble-spark-application-chart` is not being pushed up to GitHub packages yet, we will make some local modification
     - In the`foundation/foundation-mda/src/main/resources/templates/data-delivery-common/pipeline.chart.pom.configuration.vm`, modify the line 32 to point to your local directory `<<path to your baseline repo>>/extensions/extensions-helm/aissemble-spark-application-chart`
     - In the `foundation/foundation-mda/src/main/resources/templates/data-delivery-common/pipeline.chart.pom.configuration.vm`, modify the line 24 to be `{path = "/path to your baseline repo/extensions/extensions-data-delivery/aissemble-extensions-data-delivery-spark-py", develop = true}`
     - Run `mvn clean install -pl :foundation-mda -Dmaven.build.cache.skipCache`
- Run `mvn -B clean install -Parchetype-test -pl :foundation-archetype` and verify the build is success.

## References/Additional Context
As needed

ewilkins-csi:
**DoD/OTS** looks good ✅

carter-cundiff:
Testing passed:
![image](https://github.com/user-attachments/assets/149c93ec-a1cc-4572-acb4-878377c9eded)

PR #402, opened 2024-10-08
#400 overwrite ivy user directory in aissemble-spark-application-chart

csun-cpointe:
For java 17 upgrade, we have upgraded the aissemble-spark base image, which has a different home directory than the /opt/spark/ directory so to make sure Ivy cache can find the right directory (Ivy cache is default to the environment's home directory. e.g.: ~/.ivy2), we will need to override the ivy user directory by specifying the spark.jars.ivy directory in the chart.

Currently, to fix the build issue, we have moved the `spark.jars.ivy` from the spark application chart to the dev values file to avoid the local test issue. However, since our local test is not using the docker container but only copy over the spark configuration from the yaml file to configure the spark session in the local environment, we should not copy spark.jars.ivy directory for the spark tester setup but leave it as default so that it can still use the default home directory and when we deploying the spark application to different environment, we don't need to set spark.jars.ivy path in the values file.

### DIFF
--- a/extensions/extensions-data-delivery/aissemble-extensions-data-delivery-spark-py/src/data_delivery_spark_py/test_utils/spark_session_manager.py
+++ b/extensions/extensions-data-delivery/aissemble-extensions-data-delivery-spark-py/src/data_delivery_spark_py/test_utils/spark_session_manager.py
@@ -47,7 +47,11 @@ def create_standalone_spark_session(sparkapplication_path: str) -> SparkSession:
             ",".join(spec_map["spec"]["deps"]["repositories"]),
         )
     for key, value in spec_map["spec"]["sparkConf"].items():
-        builder = builder.config(key, value)
+        logger.info("sparkConf - key: {}, value: {}: ".format(key, value))
+        if "spark.jars.ivy" != key.lower():
+            builder = builder.config(key, value)
+        else:
+            logger.info("Skip spark config - key: {}: ".format(key))
     return builder.getOrCreate()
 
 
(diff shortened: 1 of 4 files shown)


extensions/extensions-data-delivery/aissemble-extensions-data-delivery-spark-py/src/data_delivery_spark_py/test_utils/spark_session_manager.py
```diff
@@ -47,7 +47,11 @@ def create_standalone_spark_session(sparkapplication_path: str) -> SparkSession:
             ",".join(spec_map["spec"]["deps"]["repositories"]),
         )
     for key, value in spec_map["spec"]["sparkConf"].items():
-        builder = builder.config(key, value)
+        logger.info("sparkConf - key: {}, value: {}: ".format(key, value))
+        if "spark.jars.ivy" != key.lower():
+            builder = builder.config(key, value)
+        else:
+            logger.info("Skip spark config - key: {}, value: {}: ".format(key, value))
```
ewilkins-csi, line 54: A: Can probably just print the key to be a little less noisy since we just printed the key and value above.

foundation/foundation-mda/src/main/resources/templates/cucumber.spark.test.base.harness.java.vm
```diff
@@ -92,7 +92,11 @@ public abstract class SparkTestBaseHarness implements EventListener {
 
         for (Map.Entry<String, String> entry : sparkApplicationConfig.spec().sparkConf().entrySet()) {
             logger.warn(entry.getKey() + " " + entry.getValue());
-            builder = builder.config(entry.getKey(), entry.getValue());
+            if (!entry.getKey().equalsIgnoreCase("spark.jars.ivy")) {
+                builder = builder.config(entry.getKey(), entry.getValue());
+            } else {
+                logger.warn(String.format("Skip spark config: %s - %s", entry.getKey(), entry.getValue()));
```
ewilkins-csi, line 98: A: Recommend info here since we expect this value to always exist and be skipped

ewilkins-csi (approved):

Issue #302, opened 2024-08-27
Feature: Add imagePullSecrets support to Policy Decision Point chart


lemuelwatts:
## Description
The v2 charts for Policy Decision Point do not currently allow for the user to add an imagePullSecret so that images can be pulled from private image registries. We would like to add this functionality to the chart so that it can be used more easily in downstream projects that require the use of a private image registry.

## DOD
- [x] Add Policy Decision Point chart reference to imagePullSecret
- [x] Verify Policy Decision Point chart can be rendered that includes the imagePullSecret field and value
- [x] Add unit test around adding support for imagePullSecret.
- [x] Update README to support imagePullSecrets to Policy Decision Point Chart.
- [x] Verify aissemble build is successful.
- [x] Verify Downstream project that uses PDP with imagePullSecret is successful.

## Test Strategy/Script

- On Github, Go to Settings,
- On left tab Go to Developer setting.
- Click on personal access token (classic) and generate classic token.
Scope should be something like this:
![Screenshot 2024-10-10 at 10 12 37 AM](https://github.com/user-attachments/assets/52f3f271-28ec-42b6-9839-e237d549121c)

- Copy the gpg key ( personal access token) 
- Create secret with this command
` kubectl create secret docker-registry regcred --docker-server=ghcr.io --docker-username={your github name} --docker-password={your Personal access token}`

- Check to see if you have any PDP docker images by running command
`docker image`
something like this:
`{projectname}-policy-decision-point-docker                    tilt-16367d62958476b5   16367d629584   23 hours ago    417MB
`
NOTE: if you don't have any pdp-docker image, run downstream project with pdp and deploy, then this docker image will be generated. 

- Change Docker tag to your private registry.
`docker tag {projectname}-policy-decision-point-docker:{tilt-tag} ghcr.io/{github username}/issue-302-policy-decision-point-docker:1.0.0-SNAPSHOT`

- push docker tag to your private registry 
`docker push ghcr.io/{github username}/issue-302-policy-decision-point-docker:1.0.0-SNAPSHOT`

- check to confirm you have package uploaded
`https://github.com/{your github username}?tab=packages`

- pull in latest dev branch of aissemble
- make sure build is successful
Build downstream project using following command:
```
 mvn archetype:generate -U -DarchetypeGroupId=com.boozallen.aissemble \
  -DarchetypeArtifactId=foundation-archetype \
  -DarchetypeVersion=1.10.0-SNAPSHOT\
[...]

PR #405, opened 2024-10-09
#302 imagePullSecrets support to Policy Decision Point chart

jaebchoi:
The v2 charts for Policy Decision Point do not currently allow for the user to add an imagePullSecret so that images can be pulled from private image registries. We would like to add this functionality to the chart so that it can be used more easily in downstream projects that require the use of a private image registry.

Issue #302 https://github.com/boozallen/aissemble/issues/302

### DIFF
--- a/extensions/extensions-helm/aissemble-policy-decision-point-chart/README.md
+++ b/extensions/extensions-helm/aissemble-policy-decision-point-chart/README.md
@@ -9,18 +9,19 @@ helm install policy-decision-point oci://ghcr.io/boozallen/aissemble-policy-deci
 **Note**: *the version should match the aiSSEMBLE project version.*
 
 # Properties
-| Property                      | Description                                 | Required Override | Default                                                                            |
-|-------------------------------|---------------------------------------------|-------------------|------------------------------------------------------------------------------------|
-| app.name                      | Sets label for app.kubernetes.io/name       | No                | Chart.Name (aissemble-policy-decision-point-chart)                                 |
-| app.version                   | Sets label for app.kubernetes.io/version    | No                | Chart.AppVersion (aiSSEMBLE project version)                                       |
-| hostname                      | The hostname for the application            | No                | policy-decision-point                                                              |
-| image.name                    | The image name                              | Yes               | boozallen/aissemble-policy-decision-point                                          |
-| image.imagePullPolicy         | The image pull policy                       | No                | Always (ensures local docker image is pulled, rather than from Nexus repo)         |
-| image.dockerRepo              | The image docker repository                 | No                | ghcr.io/                                                                           |
-| image.tag                     | The image tag                               | No                | Chart.AppVersion                                                                   |
-| service.spec.ports            | The service spec ports                      | No                | - name: rest-api <br/>&emsp;&emsp;port: 8080 <br/>&emsp;&emsp;targetPort: 8080     |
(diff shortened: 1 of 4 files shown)


extensions/extensions-helm/aissemble-policy-decision-point-chart/README.md
```diff
@@ -9,18 +9,19 @@ helm install policy-decision-point oci://ghcr.io/boozallen/aissemble-policy-deci
...
+| app.name                     | Sets label for app.kubernetes.io/name                              | No                | Chart.Name (aissemble-policy-decision-point-chart)                                 |
+| app.version                  | Sets label for app.kubernetes.io/version                           | No                | Chart.AppVersion (aiSSEMBLE project version)                                       |
+| hostname                     | The hostname for the application                                   | No                | policy-decision-point                                                              |
+| image.name                   | The image name                                                     | Yes               | boozallen/aissemble-policy-decision-point                                          |
+| image.imagePullPolicy        | The image pull policy                                              | No                | Always (ensures local docker image is pulled, rather than from Nexus repo)         |
```
chang-annie, line 18: should this read "ensures local docker image is pulled, rather than from **ghcr.io** repo" instead of "Nexus repo"? Looks like we have a couple other `extensions-helm/*-chart/README.md` files that also still reference Nexus too. @ewilkins-csi

extensions/extensions-helm/aissemble-policy-decision-point-chart/README.md
```diff
@@ -9,18 +9,19 @@ helm install policy-decision-point oci://ghcr.io/boozallen/aissemble-policy-deci
...
+| image.tag                    | The image tag                                                      | No                | Chart.AppVersion                                                                   |
+| service.spec.ports           | The service spec ports                                             | No                | - name: rest-api <br/>&emsp;&emsp;port: 8080 <br/>&emsp;&emsp;targetPort: 8080     |
+| deployment.ports             | The deployment ports                                               | No                | - name: http-1 <br/>&emsp;&emsp;containerPort: 8080 <br/>&emsp;&emsp;protocol: TCP |
+| deployment.restartPolicy     | The deployment restart policy                                      | No                | Always                                                                             |
+| deployment.imagePullSecrets  | The secrets to use for pulling any of the images used by this Pod. | No                | []                                                                                 |
```
chang-annie, line 24: I think `pod` can be lowercase here!

extensions/extensions-helm/aissemble-policy-decision-point-chart/README.md
carter-cundiff: Always means it will always pull from the remote repo, so this is not accurate as well

extensions/extensions-helm/aissemble-policy-decision-point-chart/templates/deployment.yaml
```diff
@@ -24,6 +24,10 @@ spec:
         {{- toYaml . | nindent 12}}
       {{- end }}
       serviceAccountName: {{ .Values.deployment.serviceAccountName }}
+      {{- if .Values.deployment.imagePullSecrets }}
```
carter-cundiff, line 27: A: Would be cleaner to use `{{- with ...` and `{{- toYaml .` like the securityContext section above.

carter-cundiff (approved):

extensions/extensions-helm/aissemble-policy-decision-point-chart/templates/deployment.yaml
jaebchoi: Sounds good, will update to with

PR #410, opened 2024-10-10
#379 Investigate slow nightly CI build time & improve it

meliz19:
I have two commits in this PR: one includes the changes I made to speed up the `aissemble-extensions-model-training-api-sagemaker` build time in case someone wants to use the sagemaker code and the second removes all of the Sagemaker modules and related documentation.


test/test-mda-models/test-machine-learning-model/aissemble-machine-learning-sagemaker-training/pyproject.toml
```diff
@@ -16,18 +16,15 @@ python = ">=3.8"
 sagemaker-training = ">=4.6.1"
 mlflow = ">=2.16.2"
 
-# Poetry lock was timing out attempting to resolve dependencies
```
meliz19, line 19: Cleaned up unnecessary explicit dependency imports

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/generator/MachineLearningPipelineGenerator.java
```diff
@@ -95,18 +94,15 @@ private boolean enableAutoTrain(MachineLearningPipeline pipeline) {
...
-        // this surrounding if-check was added to prevent piplines that use sagemaker training from
-        // throwing a null pointer exception when detecting if Onnx conversion code is needed to be generated
-        if (pipeline.getTrainingStep() != null) {
-            for (PostAction postAction : pipeline.getTrainingStep().getPostActions()) {
-                if (PipelineUtils.forOnnxModelConversion(postAction)) {
-                    containsOnnx = true;
-                    break;
-                }
+        
+        for (PostAction postAction : pipeline.getTrainingStep().getPostActions()) {
```
ewilkins-csi, line 98: A: I know the comment said it was for SageMaker pipelines, but the `getTrainingStep` method's contract is to return `null` if there is no training step.  So I think keeping the `null` check/guard here makes sense instead of just relying on program flow to only route pipelines with training steps to this method.

ewilkins-csi (approved):
🎉  So long, SageMaker, and thanks for all the fish!

Issue #398, opened 2024-10-07
Feature: Java 17 Upgrade Support Series - Pipeline persistence


ewilkins-csi:
## Description
In #356 we upgraded the bulk of Spark pipeline functionality to Java 17. This tackles the follow-on work of updating the persistence logic of pipeline steps.  The available persistence types are:
 - hive
 - delta-lake
 - rdbms
   - **Note:** this used to be the `postgres` option but was updated to be `rdbms`. However, because many of the Java configs still refer to postgres it's unclear if this was tested with anything other than postgres.
   - **Update:** Java's `postgres` and `rdbms` options are separate but entangled so neither works
 - elasticsearch
 - neo4j

## DOD
Acceptance criteria required to realize the requested feature
- [x] A pipeline can write to and read from the following data sources while running on the new Java 17 Spark Worker:
   - [x] hive
   - [x] delta-lake
   - ~rdbms~
     - ~MySql (fallback to Postgres if RDBMS hasn't been properly generalized)~
     - ~Clean up config names (if time allows)~
   ---- stretch goal ----
   - [x] elasticsearch
     - Did not test reading from ES as I couldn't quite figure out the syntax and it was a low-priority item 
   - [x] neo4j
     - _Note:_ There is a pre-existing issue with Neo4j that prevents _reading_ data from a table. The data can be saved, just not read.
 - [x] Write migrations for updates
   - [x] Migrate SparkApplication delta-lake dependencies
   - [x] Migrate POM delta-lake dependencies

### Main Test
1. Create a new project on 1.10.0-SNAPSHOT.
    ```sh
    mvn archetype:generate '-DarchetypeGroupId=com.boozallen.aissemble' \
                           '-DarchetypeArtifactId=foundation-archetype' \
                           '-DarchetypeVersion=1.10.0-SNAPSHOT' \
                           '-DgroupId=org.test' \
                           '-Dpackage=org.test' \
                           '-DprojectGitUrl=test.org/test.git' \
                           '-DprojectName=Final Test 398' \
                           '-DartifactId=final-398' \
    && cd final-398
    ```
2. Set your Java version to **_17_** if it is not currently
3. Add a [Spark Pipeline model](https://github.com/user-attachments/files/17343578/PersistPipeline.json) with a step for each persist type & mode.
4. Fully generate the project by running `mvn clean install` and following manual actions
[...]

PR #412, opened 2024-10-11
[#398] update Delta Lake dependencies to 3.2.1

ewilkins-csi:
Version 3.2.1 added support for Spark 3.5.3. Note the name of the `delta-core` artifacts changed to `delta-spark`.


foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/generator/SparkApplicationGenerator.java
```diff
@@ -115,20 +114,6 @@ public void generate(GenerationContext context) {
         vc.put("versionNeo4j", config.getNeo4jVersion());
         vc.put("versionAwsSdkBundle", config.getAwsSdkBundleVersion());
 
-        // Addresses a Spark bug in which the --packages argument was not respected in cluster mode (3.0.0 <= x < 3.4.0)
-        // We don't currently support Spark <3.x.x in aiSSEMBLE, but we include a check for it just in case we ever do.
```
ewilkins-csi, line 119: I: We don't support changing the Spark version from the default aiSSEMBLE version (currently 3.5.2). It's not really worth it to keep the complexity around supporting older versions since we don't allow it.  If we revisit this due to demand, the logic isn't too bad to re-introduce. (And would likely be a very small portion of the overall work that would need to be done to support multiple Spark versions.)

foundation/foundation-mda/src/main/resources/templates/deployment/spark-operator/spark-application-base-values.yaml.vm
```diff
@@ -9,34 +9,6 @@ sparkApp:
     #end
     mainApplicationFile: "local:///opt/spark/jobs/pipelines/${mainApplicationFile}"
     deps:
-      #if (!${usePackageDeps})
```
ewilkins-csi, line 12: I: Because we pulled out the logic for supporting older versions of Spark, `usePackageDeps` can be assumed to be `true`.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_10_0/DeltaSparkYamlMigration.java
```diff
@@ -0,0 +1,105 @@
...
+import org.slf4j.LoggerFactory;
+import org.technologybrewery.baton.BatonException;
+import org.yaml.snakeyaml.error.YAMLException;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.List;
+
+public class DeltaSparkYamlMigration extends AbstractAissembleMigration {
```
ewilkins-csi, line 25: A: Add class-level javadoc

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_10_0/DeltaSparkPomMigration.java
```diff
@@ -0,0 +1,104 @@
...
+                .filter(DeltaSparkPomMigration::isDeltaCore)
+                .forEach(deltaCoreDependencies::add);
+        return deltaCoreDependencies;
+    }
+
+    private static boolean isDeltaCore(Dependency dep) {
+        return dep.getGroupId().equals("io.delta") && dep.getArtifactId().startsWith("delta-core_");
+    }
+
+    private static PomModifications.Replacement replaceInTag(InputLocationTracker container, String tag, String contents) {
```
ewilkins-csi, line 98: I: Good candidate to bake into the `PomModifications` class in Baton

csun-cpointe (approved):
LGTM

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_10_0/DeltaSparkYamlMigration.java
```diff
@@ -0,0 +1,105 @@
...
+        if (file.getName().endsWith(".yaml")) {
+            try {
+                YamlUtils.YamlObject yaml = YamlUtils.loadYaml(file);
+                if (yaml.hasObject("sparkApp")) {
+                   yaml = yaml.getObject("sparkApp");
+                }
+                if (yaml.hasObject("spec")) {
+                   yaml = yaml.getObject("spec");
+                    if (yaml.hasObject("deps")) {
+                        yaml = yaml.getObject("deps");
```
csun-cpointe, line 39: Maybe create an utility function to get the object from a path. e.g.: `YamlUtils.getObjectFromPath("sparkApp.spec.deps")` ?

foundation/foundation-upgrade/src/test/resources/test-files/v1_10_0/DeltaSparkYamlMigration/migration/spark-app-values.yaml
```diff
@@ -0,0 +1,44 @@
...
+    mainClass: org.test.PersistPipelineDriver
+    mainApplicationFile: "local:///opt/spark/jobs/pipelines/persist-pipeline.jar"
+    deps:
+      packages:
+        - mysql:mysql-connector-java:8.0.30
+        - org.apache.hadoop:hadoop-aws:3.3.4
+        - com.amazonaws:aws-java-sdk-bundle:1.12.262
+        - org.neo4j:neo4j-connector-apache-spark_2.12:4.1.5_for_spark_3
+        - org.elasticsearch:elasticsearch-spark-30_2.12:8.9.0
+        - io.delta:delta-core_2.12:2.4.0 #comment
```
csun-cpointe, line 16: It's good that we tested with comment 👍

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/util/YamlUtils.java
```diff
@@ -193,6 +186,36 @@ public Collection<Object> values() {
...
+        private <T> boolean hasValue(Class<T> type, String... path) {
+            YamlObject parent = getParentIfExists(path);
+            if (parent == null) {
+                return false;
+            }
+            String key = path[path.length - 1];
+            return parent.containsKey(key) && type.isInstance(parent.get(key));
+        }
+
+        private <T> T getValue(String... path) {
```
ewilkins-csi, line 199: I: An argument could be made that we only need these two type-parameterized methods and the `hasObject/getObject` methods, but it can be difficult to understand how to work with an API that only has type-parameterized methods.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/util/YamlUtils.java
```diff
@@ -193,6 +186,36 @@ public Collection<Object> values() {
...
+            YamlObject parent = getParentIfExists(path);
+            if (parent == null) {
+                throw new NoSuchElementException("Parent object does not exist at " + Arrays.toString(path));
+            }
+            String key = path[path.length - 1];
+            return (T) parent.get(key);
```
ewilkins-csi, line 205: I: If we passed in a `Class` object for the type, we could provide a more helpful message when the type of the value doesn't match what's expected. E.g.:
```java
    if (!type.isInstance(parent.get(key)) {
        throw new IllegalArgumentException("Value for [" + key + "] is not of type " + type.getSimpleName());
    }
```

However, I guessed that in nearly every case the call site will contain the literal path to the value being accessed and so a couple extra steps to check the source code to find what value isn't of the expected type isn't that big of a deal and simplifies both the call site and the implementation of this method.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_10_0/DeltaSparkYamlMigration.java
ewilkins-csi: I just updated the `YamlObject` hasX/getX methods to accept full paths instead of single keys.  Prevents this nesting situation while not cluttering the `YamlObject` API.

Issue #359, opened 2024-09-24
Feature: JDK 17 Upgrade Support Series - Alerting


habibimoiz:
## Description
In https://github.com/boozallen/aissemble/issues/133 we modified the build-parent to use JDK 17. When building aiSSEMBLE with the updated build-parent, some modules do not build successfully. This series of ticket(s) will upgrade the Alerting modules. 

## DOD
- [x] Verify foundation-alerting, extensions-alerting, extensions-alerting-teams, and extensions-alerting-slack modules build and all tests pass. 
- [x] Clean up pom in the above modules to remove unnecessary dependencies and add undeclared transient dependencies 
  - [x] Use `mvn dependency:analyze`

## Test Strategy/Script
- Clone the aiSSEMBLE repository https://github.com/boozallen/aissemble
- In the project's root directory, execute `mvn clean install -pl :extensions-alerting-teams -am`, verify a successful build
- Execute `mvn clean install -pl :extensions-alerting-slack -am`, verify a successful build and Cucumber tests run

PR #418, opened 2024-10-17
#359 - Modified aissemble-alerting-slack tests to run

jacksondelametter:
PR is to verify the alerting modules are able to run with JDK 17. Fixed an issue with the extensions-alerting-slack module where the unit tests were not being ran on a build. Cleaned up unused dependencies and added transient dependency declarations in the pom files for the foundation-alerting, extensions-alerting, extensions-alerting-teams, and extensions-alerting-slack modules

### DIFF
--- a/extensions/extensions-alerting/extensions-alerting-slack/pom.xml
+++ b/extensions/extensions-alerting/extensions-alerting-slack/pom.xml
@@ -22,15 +22,36 @@
             <groupId>org.technologybrewery.krausening</groupId>
             <artifactId>krausening</artifactId>
         </dependency>
-
-        <!-- Logging Dependencies (from aiops-parent pom) -->
         <dependency>
-            <groupId>org.apache.logging.log4j</groupId>
-            <artifactId>log4j-slf4j2-impl</artifactId>
+            <groupId>com.boozallen.aissemble</groupId>
+            <artifactId>foundation-core-java</artifactId>
+            <version>${project.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.apache.logging.log4j</groupId>
-            <artifactId>log4j-core</artifactId>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>${version.junit}</version>
+            <scope>test</scope>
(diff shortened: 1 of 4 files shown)


extensions/extensions-alerting/extensions-alerting-slack/pom.xml
```diff
@@ -22,15 +22,40 @@
...
         <dependency>
-            <groupId>org.apache.logging.log4j</groupId>
-            <artifactId>log4j-slf4j2-impl</artifactId>
+            <groupId>com.boozallen.aissemble</groupId>
+            <artifactId>foundation-core-java</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.aeonbits.owner</groupId>
+            <artifactId>owner</artifactId>
```
ewilkins-csi, line 32: Q/A: Was this added because we reference Owner classes via our usage of Krausening? If so, I think it probably makes more sense to just add Krausening and not Owner, as we'd probably not use Owner outside of Krausening.

extensions/extensions-alerting/extensions-alerting-slack/pom.xml
```diff
@@ -22,15 +22,40 @@
...
+            <version>1.0.9</version>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>${version.junit}</version>
```
ewilkins-csi, line 38: S:
```suggestion
            <version>${version.junit}</version>
            <scope>test</scope>
```

jacksondelametter: Fixed.

extensions/extensions-alerting/extensions-alerting-slack/pom.xml
jacksondelametter: Yes it is and Krausening is already added as a dependency. I have removed the reference to the Owner classes.

ewilkins-csi (approved):

Issue #421, opened 2024-10-18
Feature: JDK17 Upgrade Series: Data and Model Lineage


csun-cpointe:
## Description
In #133  we modified the core libraries t to use JDK 17. As part of upgrade to java 17 series, this ticket will focus on the data and model lineage functionality.

## DOD
Acceptance criteria required to realize the requested feature
- [x] Upgrade data and model lineage to build with Java 17:
   - [x] Ensure all modules are compatible with java 17
      - [x] aissemble-nvidia
      - [x] foundation-lineage
         - [x] foundation-modle-lineage
         - [x] foundation-data-lineage-python
         - [x] foundation-data-lineage-consumer
         - [x] foundation-data-lineage-java
      - [x] extensions-data-lineage-http-consumer-service
      - [x] extensions-data-lineage-http-consumer
   - [x] Clean up pom to remove unnecessary dependencies
   - [x] Data lineage functionalities work as expected
      - [x] Spark Pipeline can send data-lineage message
      - [x] Model Training pipeline can send model-lineage message

## Test Strategy/Script
1. Using create a new aissemble-based project using the latest archetype snapshot.
```
mvn archetype:generate '-DarchetypeGroupId=com.boozallen.aissemble' \
                           '-DarchetypeArtifactId=foundation-archetype' \
                           '-DarchetypeVersion=1.10.0-SNAPSHOT' \
                           '-DgroupId=org.test' \
                           '-Dpackage=org.test' \
                           '-DprojectGitUrl=test.org/test.git' \
                           '-DprojectName=Final Test 421' \
                           '-DartifactId=test-421' \
    && cd test-421
```
2. Set your Java version to **_17_** if it is not currently
3. Under -model/src/main/resources/pipelines add both [ClassificationTraining.json](https://github.com/user-attachments/files/17482480/ClassificationTraining.json) and [SparkPipeline.json](https://github.com/user-attachments/files/17482481/SparkPipeline.json)
 files
4. Add below execution to the -deploy/pom.xml
```
                     <execution>
                        <id>lineage-http-consumer</id>
                        <phase>generate-sources</phase>
                        <goals>
                            <goal>generate-sources</goal>
                        </goals>
                        <configuration>
                            <basePackage>com.boozallen.aissemble.documentation</basePackage>
                            <profile>lineage-http-consumer-deploy-v2</profile>
                            <!-- The property variables below are passed to the Generation Context and utilized to customize the deployment artifacts. -->
                            <propertyVariables>
[...]

PR #428, opened 2024-10-23
#421 data/model lineage java17 upgrade

csun-cpointe:


foundation/foundation-lineage/foundation-data-lineage/foundation-data-lineage-java/src/main/java/com/boozallen/aissemble/data/lineage/DatasetFacet.java
```diff
@@ -57,7 +57,7 @@ public class DefaultDatasetFacet extends OpenLineage.DefaultDatasetFacet {
         private final URI schemaUrl;
 
         public DefaultDatasetFacet(String schemaUrl, URI producer) {
-            super(producer);
+            super(producer, null);
```
carter-cundiff, line 60: Q: Is this due to a change in the parent class? What parameter as we passing as null?

foundation/foundation-mda/src/main/resources/profiles.json
```diff
@@ -542,9 +542,6 @@
 			},
 			{
 				"name": "mlflowLineageProperties"
-			},
-			{
-				"name": "trainingGitkeepFile"
```
carter-cundiff, line 547: Q: What is this removing?

csun-cpointe: because the directory is not empty so we don't need the keep file there

foundation/foundation-lineage/foundation-data-lineage/foundation-data-lineage-java/src/main/java/com/boozallen/aissemble/data/lineage/DatasetFacet.java
csun-cpointe: it's a new param added to the constructor, which is used to mark for delete facet

foundation/foundation-mda/src/main/resources/templates/general-docker/mlflow.lineage.properties.vm
```diff
@@ -1,2 +1,3 @@
 data.lineage.enabled=true
-data.lineage.producer=${scmUrl}
\ No newline at end of file
+data.lineage.producer=${scmUrl}
+data.lineage.namespace=default
```
carter-cundiff, line 3: Q: Will is break model lineage in existing projects if they fail to add this new property? Might need to add to release notes. 

Also below in [training.base.py.vm](https://github.com/boozallen/aissemble/pull/428/files#diff-ab1998bd6ea1c97c36eedefef420e66a7c458bd28f078d525032b361cbaccb0d) we set the default namespace to the pipeline name. Why do we need to set it here too?

csun-cpointe: This is a miss for the setting default namespace functionality per OpenLineage expected usage. We had changes made for the pyspark pipeline but missed the model training.  
I suspect there are existing project using it because it should be broken right now because of the miss. The new add on property and changes will fix the code.

Yeah.. I will add to the release note

carter-cundiff: Is this namespace different from the namespace set in [training.base.py.vm](https://github.com/boozallen/aissemble/pull/428/files#diff-ab1998bd6ea1c97c36eedefef420e66a7c458bd28f078d525032b361cbaccb0d)?

carter-cundiff: Also `data.lineage.namespace` is the legacy property. If we're adding a new one, it should be in `data.lineage.${pipelineStep}.namespace` format.

csun-cpointe: yeah.. this is a customizable value. We can either set a default value for them or we can just add in the release doc and have them set it by themselves to match the pipeline name or whichever value the prefer.

carter-cundiff: Discussed offline, this namespace config is needed to set the default for the dataset lineage object. Without it the pipeline will fail to run.

carter-cundiff (approved):

Issue #4, opened 2024-04-28
Fix Cyclic helm tests in `test-mda-models`


d-ryan-ashcraft:
### Description

While migrating to the ghcr.io as aissemble's new helm repository, we encountered some cyclic tests in the the `test-mda-models` module.  The core issue is the `test-chart` execution of `maven-exec-plugin`, which generally looks like:
```xml
            <plugin>
                <groupId>org.codehaus.mojo</groupId>
                <artifactId>exec-maven-plugin</artifactId>
                <version>3.1.0</version>
                <executions>
                    <execution>
                        <id>test-chart</id>
                        <phase>compile</phase>
                        <goals>
                            <goal>exec</goal>
                        </goals>
                        <configuration>
                            <executable>helm</executable>
                            <arguments>
                                <argument>template</argument>
                                <argument>aissemble-spark-application</argument>
                                <argument>--repo</argument>
                                <argument>${aissemble.helm.repo}</argument>
                                <argument>--version</argument>
                                <argument>${version.aissemble}</argument>
                                <argument>--values</argument>
                                <argument>
                                    src/aissemble_test_data_delivery_pyspark_model_basic/resources/apps/pyspark-data-delivery-basic-base-values.yaml,tests/resources/apps/pyspark-data-delivery-basic-test-values.yaml
                                </argument>
                                <argument>-s</argument>
                                <argument>templates/deployment.yaml</argument>
                            </arguments>
                            <outputFile>target/apps/pyspark-data-delivery-basic-test-chart.yaml</outputFile>
                        </configuration>
                    </execution>
                </executions>
            </plugin>
```
In this block, we reference a remote helm repo:
```xml
                                <argument>--repo</argument>
                                <argument>${aissemble.helm.repo}</argument>
```

Of course in the process of migrating to a new helm repo, no charts have been deployed here yet, which uncovered the cycle.  We should likely do the following:
- update to use the local version of the chart
- refactor to at least pull this up to a common `pluginDependency` definition in `test-mda-models` so it isn't repeated almost verbatim in four different modules.  More ideally, we'd eliminate the use of the `maven-exec-plugin` for this altogether as it is a generally a pretty good sign we have a workaround in place.  Since we've forked the `helm-maven-plugin`, why not support this with it?

This impacts the following four modules:
- `test-mda-data-delivery-model`
- `test-mda-data-delivery-model-basic`
- `aissemble-test-data-delivery-pyspark-model`
[...]

Issue #430, opened 2024-10-23
BUG: FIx data delivery steps not injecting smallrye when running test-data-delivery-spark-model.


jaebchoi:
## Description
We have disabled some data delivery steps as those steps are not injecting smallrye reactive messaging correctly.

Here are missing steps 
```
		{
			"name": "VoidInboundAndMessagingOutboundAsync",
			"type": "asynchronous",
			"outbound": {
				"type": "messaging",
				"channelName": "outboundChannel"
			}
		},

		{
			"name": "MessagingOutboundWithCustomTypesAsync",
			"type": "asynchronous",
			"outbound": {
				"type": "messaging",
				"channelName": "outboundChannel",
				"recordType": {
					"name": "CustomRecord",
					"package": "com.boozallen.aiops.mda.pattern.record"
				}
			}
		},

		{
			"name": "MessagingOutboundWithCustomRecAsync",
			"type": "asynchronous",
			"outbound": {
				"type": "messaging",
				"channelName": "outboundChannel",
				"recordType": {
					"name": "CustomRec",
					"package": "com.boozallen.aiops.mda.pattern.record"
				}
			}
		},
```

It seems like any step with asynchronous type,  no inbound messaging , and messaging outbound is having issue with running step execution.

## Steps to Reproduce
- On `aissemble/test/test-mda-models/test-data-delivery-spark-model/src/main/resources/pipelines/SparkJavaDataDeliveryPatterns.json` Add missing 3 steps mentioned above.
- Run mvn clean install
- log will show Data Lineage Recording is executed repeatedly and does not proceed further.

## Expected Behavior
- Data Lineage Recording is executed only once and move on.

## Actual Behavior
- Data Lineage Recording is executed repeatedly.

jaebchoi:
DoD completed with @aaron-gary

jaebchoi:
OTS Completed with @chang-annie

cpointe-ibllanos:
Tested - did not experience stalling on data lineage recording in either CI or locally. Builds pass.

PR #431, opened 2024-10-23
#4 Fix Cyclic helm tests in test-mda-models

jaebchoi:
While migrating to the ghcr.io as aissemble's new helm repository, we encountered some cyclic tests in the the test-mda-models module. 

https://github.com/boozallen/aissemble/issues/4


test/test-mda-models/test-data-delivery-spark-model/src/main/resources/pipelines/SparkJavaDataDeliveryPatterns.json
```diff
@@ -68,14 +68,6 @@
...
 			}
 		},
-		{
-			"name": "VoidInboundAndMessagingOutboundAsync",
-			"type": "asynchronous",
-			"outbound": {
-				"type": "messaging",
-				"channelName": "outboundChannel"
-			}
-		},
```
jaebchoi, line 78: Seems to be having some issue with async messaging outbound without inbound messaging.
added https://github.com/boozallen/aissemble/issues/430 to follow up.

test/test-mda-models/test-data-delivery-spark-model/pom.xml
```diff
@@ -248,5 +228,17 @@
             <version>${version.cucumber.reporting.plugin}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jboss.resteasy</groupId>
+            <artifactId>resteasy-client</artifactId>
+            <version>3.15.6.Final</version>
+            <scope>test</scope>
```
jaebchoi, line 235: I: Had to pin this version as investigated for few days and other versions doesn't seems to be compatible with this error 
.AbstractMethodError: Receiver class org.jboss.resteasy.microprofile.config.ServletConfigSource does not define or inherit an implementation of the resolved method 'abstract java.util.Set getPropertyNames()' of interface org.eclipse.microprofile.config.spi.ConfigSource.

foundation/foundation-mda/src/main/resources/templates/cucumber.spark.test.base.harness.java.vm
```diff
@@ -34,7 +34,7 @@ public abstract class SparkTestBaseHarness implements EventListener {
 
 	private static final Logger logger = LoggerFactory.getLogger(SparkTestBaseHarness.class);
 	protected static final SparkConfig config = KrauseningConfigFactory.create(SparkConfig.class);
-    private static final File DEFAULT_SPARK_TEST_APPLICATION_SPEC = new File("./target/apps/${artifactId}-test-chart.yaml");
+    private static final File DEFAULT_SPARK_TEST_APPLICATION_SPEC = new File("./target/apps/aissemble-spark-application-chart/templates/deployment.yaml");
 #if (${pipeline.isHiveSupportNeeded()})
```
ewilkins-csi, line 38: C: I don't think this is correct when a downstream project is being created. We generate the application yaml files based on the pipeline name. This will break downstream project pipeline tests.

test/test-mda-models/aissemble-test-data-delivery-pyspark-model-basic/pom.xml
```diff
@@ -87,38 +87,16 @@
...
-                                </argument>
-                                <argument>-s</argument>
-                                <argument>templates/deployment.yaml</argument>
-                            </arguments>
-                            <outputFile>target/apps/pyspark-data-delivery-basic-test-chart.yaml</outputFile>
-                        </configuration>
-                    </execution>
-                </executions>
+                <groupId>${group.helm.plugin}</groupId>
+                <artifactId>helm-maven-plugin</artifactId>
```
ewilkins-csi, line 92: S: The original plugin setup was based on what we generate for downstream projects.  I like this solution better, but if we don't align it with the expected format for downstream projects, then we aren't doing a great job of verifying the behavior.

test/test-mda-models/test-data-delivery-spark-model/pom.xml
ewilkins-csi: Q: Would this end up impacting downstream projects? If so, we will need to make sure we're tracking and resolving this issue before the release of 1.10, as I'm assuming this was caused by the Java 17 upgrade.

chang-annie: can we add this version as a property in `build-parent/pom.xml` to keep things consistent

test/test-mda-models/aissemble-test-data-delivery-pyspark-model-basic/pom.xml
jaebchoi: Yes upon investigating helm-maven-plugin doesn't support  any config option to output specific format that we want. They have output directory to dump files but they don't have option to rename to what we want.

Unfortunately I am going to fall back to using exec-maven-plugin as it provides outputFile.

Great callout! definitely missed downstream project use case.

foundation/foundation-mda/src/main/resources/templates/cucumber.spark.test.base.harness.java.vm
jaebchoi: will update to use exec-maven-plugin and revert back to original!

test/test-mda-models/test-data-delivery-spark-model/pom.xml
jaebchoi: I can create downstream project for spark to see if it is impacting!

Issue #414, opened 2024-10-16
Feature: JDK 17 Upgrade Support Series - Filestore & Pipeline model config


habibimoiz:
## Description
In https://github.com/boozallen/aissemble/issues/133 we modified the build-parent to use JDK 17. When building aiSSEMBLE with the updated build-parent, some modules do not build successfully. This issue will focus on Filestore & Pipeline model config.

## DOD

- [x] Verify that S3 filestore works with s3 local. 
- [x] Verify that the version of Jcloud we are using is compatible with Java 17. If not, then we want to upgrade to the a version of Jcloud that IS compatible with Java 17. 
- [x] Verify that Step Configuration Metamodel Options are functional with the Java 17 upgrade. 

### DOD Questions: 

-  Should the scope of this ticket include storing files into the filestore through the pipeline?
   - if so, is there a simple pipeline example that we can use to test this out? 
   - **A:** yes, we want to be able to write and read from the and to the filestore from the pipeline in the downstream project. 

## Test Strategy/Script
**_Note:_** Test 3 is theoretically completed when you complete Test 1 and Test 2. 

**Test 1 & Test 2** Push a file to LocalStack S3 using the FileStore service / Fetch a file from LocalStack S3 using the FileStore service

1. Download the attached downstream project ([_issue414.zip_](https://github.com/boozallen/aissemble/issues/414#issuecomment-2435381351)). 
2. Follow the instructions in the README.md. 
3. If all the steps in the README.md pass, then this test is complete. 

**Test 3:** Create an arbitrary key value pair using the Step Configuration Metamodel options. 

1. Download the attached downstream project ([_issue414.zip_](https://github.com/boozallen/aissemble/issues/414#issuecomment-2435381351)). 
2. Open the following file: `src/main/resources/pipelines/SimpleDataDeliveryExample.json`
3. Verify the Step Configuration metatmodel option is configured: 
```
      "configuration": [
        {
          "key": "test",
          "value": "testValue"
        }
```
4. Run `mvn clean install` and verify that the project builds successfully. 

## References/Additional Context
- [File Stores](https://boozallen.github.io/aissemble/aissemble/current/file-storage-details.html)
- [Pipeline Step Configuration Options](https://boozallen.github.io/aissemble/aissemble/current/pipeline-metamodel.html#_pipeline_step_configuration_options)

habibimoiz:
DoD completed w/ @ewilkins-csi, @carter-cundiff & @jacksondelametter.

[...]

PR #435, opened 2024-10-24
#414 - JDK 17 Upgrade Support Series - Filestore & Pipeline model config

habibimoiz:
#414 - JDK 17 Upgrade Support Series - Filestore & Pipeline model config
- Modifications to ensure FileStore and Pipeline model config is functional with the JDK 17 upgrade.
- Fixing an outdated aiSSEMBLE docs link in a FileStore template file.

### DIFF
--- a/foundation/foundation-core-java/src/main/java/com/boozallen/aissemble/core/filestore/AbstractFileStore.java
+++ b/foundation/foundation-core-java/src/main/java/com/boozallen/aissemble/core/filestore/AbstractFileStore.java
@@ -45,10 +45,13 @@ public abstract class AbstractFileStore {
      * @return a fully connected context
      */
     public BlobStoreContext openContext() {
-        return ContextBuilder.newBuilder(fileStoreConfig.getProvider())
-                .credentials(fileStoreConfig.getAccessKeyId(), fileStoreConfig.getSecretAccessKey())
-                .overrides(fileStoreConfig.getOverrides())
-                .buildView(BlobStoreContext.class);
+        ContextBuilder contextBuilder = ContextBuilder.newBuilder(fileStoreConfig.getProvider())
+                .credentials(fileStoreConfig.getAccessKeyId(), fileStoreConfig.getSecretAccessKey());
+
+        if (fileStoreConfig.getOverrides() != null) {
+            contextBuilder.overrides(fileStoreConfig.getOverrides());
+        }
+        return contextBuilder.buildView(BlobStoreContext.class);
     }
 
     /**
--- a/foundation/foundation-core-java/src/main/java/com/boozallen/aissemble/core/filestore/EnvironmentVariableFileStoreConfig.java
+++ b/foundation/foundation-core-java/src/main/java/com/boozallen/aissemble/core/filestore/EnvironmentVariableFileStoreConfig.java
@@ -28,6 +28,7 @@ public class EnvironmentVariableFileStoreConfig extends AbstractFileStoreConfig 
 
     /**
      * Get the configurations for the given filestore name.
+     *
      * @param name the name of the filestore
      * @return the configurations from the environment variables.
      */
@@ -58,11 +59,12 @@ public class EnvironmentVariableFileStoreConfig extends AbstractFileStoreConfig 
     public Properties getOverrides() {
         // Convert namespaced override values to property names understood by JClouds
         final String overridesJsonString = this.config.get(getName() + "_FS_OVERRIDES");
-        final Gson gson = new Gson();
-        final Map overrides = gson.fromJson(overridesJsonString, Map.class);
-
         final Properties propertyOverrides = new Properties();
-        propertyOverrides.putAll(overrides);
+        if (overridesJsonString != null && !overridesJsonString.isBlank()) {
+            final Gson gson = new Gson();
+            final Map overrides = gson.fromJson(overridesJsonString, Map.class);
+            propertyOverrides.putAll(overrides);
+        }
         return propertyOverrides;
     }
 }
--- a/foundation/foundation-mda/src/main/resources/templates/file-store/file.store.impl.java.vm
+++ b/foundation/foundation-mda/src/main/resources/templates/file-store/file.store.impl.java.vm
@@ -5,8 +5,7 @@ import jakarta.enterprise.context.ApplicationScoped;
 /**
  * Utility for accessing the ${fileStore.name} file storage. To customize
  * the file store connection modify the environment variables as per the documentation at
- * https://boozallen.github.io/aissemble/current/pipeline-metamodel.html
- *
+ * https://boozallen.github.io/aissemble/aissemble/current/pipeline-metamodel.html
  * Please **DO** modify with your customizations, as appropriate.
  *
  * Originally generated from: ${templateName}


foundation/foundation-core-java/src/main/java/com/boozallen/aissemble/core/filestore/AbstractFileStore.java
habibimoiz, line 1: Repairs NPE error that was occurring when the `_FS_OVERRIDES` property was not provided.

foundation/foundation-core-java/src/main/java/com/boozallen/aissemble/core/filestore/EnvironmentVariableFileStoreConfig.java
habibimoiz, line 1: Repairs NPE error that was occurring when the `_FS_OVERRIDES` property was not provided.

foundation/foundation-mda/src/main/resources/templates/file-store/file.store.impl.java.vm
habibimoiz, line 1: Fixing broken link to the pipeline-metalmodel doc.

habibimoiz:
OTS completed with @carter-cundiff & @cwoods-cpointe.

foundation/foundation-core-java/src/main/java/com/boozallen/aissemble/core/filestore/AbstractFileStore.java
```diff
@@ -45,10 +45,14 @@ public AbstractFileStore(final String name) {
      * @return a fully connected context
      */
     public BlobStoreContext openContext() {
-        return ContextBuilder.newBuilder(fileStoreConfig.getProvider())
-                .credentials(fileStoreConfig.getAccessKeyId(), fileStoreConfig.getSecretAccessKey())
-                .overrides(fileStoreConfig.getOverrides())
-                .buildView(BlobStoreContext.class);
+        ContextBuilder contextBuilder = ContextBuilder.newBuilder(fileStoreConfig.getProvider());
+        contextBuilder.credentials(fileStoreConfig.getAccessKeyId(), fileStoreConfig.getSecretAccessKey());
```
carter-cundiff, line 49:
```suggestion
        ContextBuilder contextBuilder = ContextBuilder.newBuilder(fileStoreConfig.getProvider())
            .credentials(fileStoreConfig.getAccessKeyId(), fileStoreConfig.getSecretAccessKey());
```

carter-cundiff, line 49: Can be updated to chain these together:
```
        ContextBuilder contextBuilder = ContextBuilder.newBuilder(fileStoreConfig.getProvider())
            .credentials(fileStoreConfig.getAccessKeyId(), fileStoreConfig.getSecretAccessKey());
```

habibimoiz: Chained them together.

foundation/foundation-core-java/src/main/java/com/boozallen/aissemble/core/filestore/AbstractFileStore.java
```diff
@@ -45,10 +45,14 @@ public AbstractFileStore(final String name) {
...
      */
     public BlobStoreContext openContext() {
-        return ContextBuilder.newBuilder(fileStoreConfig.getProvider())
-                .credentials(fileStoreConfig.getAccessKeyId(), fileStoreConfig.getSecretAccessKey())
-                .overrides(fileStoreConfig.getOverrides())
-                .buildView(BlobStoreContext.class);
+        ContextBuilder contextBuilder = ContextBuilder.newBuilder(fileStoreConfig.getProvider())
+                .credentials(fileStoreConfig.getAccessKeyId(), fileStoreConfig.getSecretAccessKey());
+
+        if (!fileStoreConfig.getOverrides().isEmpty()) {
```
ewilkins-csi, line 51: Q: Why is this change necessary? Does an empty overrides map cause issues? I'd think since the getter never returns `null` then it'd be safe to just pass it in.

habibimoiz: Yes, when the Overrides map is empty, we are getting a NPE and it can't invoke the map. To avoid the NPE from being thrown, the logic will now check to see if that map is null, if its not then it will chain it to the `ContextBuilder`.

foundation/foundation-core-java/src/main/java/com/boozallen/aissemble/core/filestore/EnvironmentVariableFileStoreConfig.java
```diff
@@ -58,11 +59,12 @@ public String getSecretAccessKey() {
...
-        final Gson gson = new Gson();
-        final Map overrides = gson.fromJson(overridesJsonString, Map.class);
-
         final Properties propertyOverrides = new Properties();
-        propertyOverrides.putAll(overrides);
+        if (overridesJsonString != null) {
+            final Gson gson = new Gson();
+            final Map overrides = gson.fromJson(overridesJsonString, Map.class);
+            propertyOverrides.putAll(overrides);
+        }
```
JeffreyRoss, line 67: Q: Why the need for the null check?

foundation/foundation-core-java/src/main/java/com/boozallen/aissemble/core/filestore/AbstractFileStore.java
carter-cundiff: @habibimoiz I don't think we actually tested with the empty map, we only tested with a null map. It might be fine to always add the map now that we guarantee `.getOverrides()` will never return null.

ewilkins-csi: This actually ends up causing a new NPE if the contract for `getOverrides` changes and `null` is sometimes returned. FWIW, `ContextBuilder` would throw a NPE if `getOverrides` did return `null`. So we could null-guard this instead of calling `isEmpty`. Or we could just clarify in the JavaDocs of `getOverrides` that it returns an empty map if the value hasn't been provided.

habibimoiz: I am in favor of null-guarding it. 

Proposed change: 
```
    public BlobStoreContext openContext() {
        ContextBuilder contextBuilder = ContextBuilder.newBuilder(fileStoreConfig.getProvider())
                .credentials(fileStoreConfig.getAccessKeyId(), fileStoreConfig.getSecretAccessKey());

        if (fileStoreConfig.getOverrides() != null) {
            contextBuilder.overrides(fileStoreConfig.getOverrides());
        }
        return contextBuilder.buildView(BlobStoreContext.class);
    }
```

carter-cundiff: Let's verify is still works with an empty overrides as well.

ewilkins-csi (approved):
Looks good barring test of empty overrides map

foundation/foundation-core-java/src/main/java/com/boozallen/aissemble/core/filestore/AbstractFileStore.java
habibimoiz: Just tested out with both null and empty overrides and I am getting the expected outcome!

habibimoiz:
> Looks good barring test of empty overrides map

Just tested both empty and null and its not throwing an NPE error.

Issue #419, opened 2024-10-17
Feature: 1.9.3 Patch Release


ewilkins-csi:
## Description
A clear and concise description of the new feature. It can include what the problem is and what you want to happen, and is preferably focused on the business context or how this impacts downstream users.

## DOD
Issues to potentially patch:
- Required patches:
  - [x] Thrift server connection is unstable (found via Data Access)
  - [x] The SparkOperator ArgoCD Application should have the ServerSideApply option by default. We mentioned this as a manual workaround for 1.8, and adjusted the above migration to account for it, but we never added it to the actual template so new 1.9 projects would still be broken. (See https://github.com/boozallen/aissemble/pull/416/files#r1804830533)
- Nice to have:
  - [x] Spark Infrastructure charts use `.Release.Name` instead of `.Values.app.name` like our other charts (See ca7b34f69f3f14fa)
    - If we need to drop this, make sure we add upgrade info around updating the Tiltfile helm arguments to include release name 
  - [x] `argocd-template-sync-policy-configuration-migration` adds migration instructions to files that don't need it when the entire sync policy section is in the template wholesale - e.g. after the migration runs a second time.
    - This may be specific to the project that raised the issue, and not worth fixing broadly if the logic proves difficult
  - [x] Instructions for updating custom Spark Application values files (See https://github.com/boozallen/aissemble/pull/409/files#diff-5f8520547da4ccaf2b5d54fe85ec2567b904ee5cf79af0318a600614765cc013)

Patch release DOD:
- [x] Release artifacts published
- [x] Release notes published
- [x] Main 1.9 release notes updated to patch version
- [x] Docs published

## Test Strategy/Script
1. Create a new project from 1.9.3 (SNAPSHOT for OTS)
    ```sh
    mvn archetype:generate -DarchetypeGroupId=com.boozallen.aissemble \
                           -DarchetypeArtifactId=foundation-archetype \
                           -DarchetypeVersion=1.9.3 \
                           -DgroupId=org.test \
                           -Dpackage=org.test \
                           -DprojectGitUrl=test.org/issue-419.git \
                           -DprojectName="Issue 419 Test" \
                           -DartifactId=issue-419 \
    && cd issue-419
    ```
2. Add the attached [SparkPipeline.json](https://github.com/user-attachments/files/17492461/SparkPipeline.json) to _issue-419-pipeline-models/src/main/resources/pipelines_
3. Fully generate the project until no more manual actions are printed
[...]

PR #437, opened 2024-10-24
[#419] fix invalid ArgoCD templates

ewilkins-csi:


foundation/foundation-mda/src/main/resources/templates/deployment/argocd/spark-operator.yaml.vm
```diff
@@ -21,7 +21,6 @@ spec:
         - {{ . }}
         {{- end }}    
     {{ end }}
-  {{- with .Values.spec.syncPolicy }}
```
ewilkins-csi, line 24: I: Spurious `with` block that had no closing `end` tag and caused helm templating errors

cwoods-cpointe (approved):

PR #438, opened 2024-10-24
#4.  Fix Cyclic test in test mda models

jaebchoi:
While migrating to the ghcr.io as aissemble's new helm repository, we encountered some cyclic tests in the the test-mda-models module.

https://github.com/boozallen/aissemble/issues/4

### DIFF
--- a/test/test-mda-models/test-data-delivery-spark-model-basic/pom.xml
+++ b/test/test-mda-models/test-data-delivery-spark-model-basic/pom.xml
@@ -64,7 +64,7 @@
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>build-helper-maven-plugin</artifactId>
             </plugin>
-            <!--NB: inactive by default until https://github.com/boozallen/aissemble/issues/4 is resolved: -->
+            <!--TODO: 3 steps are disabled and shall be available once https://github.com/boozallen/aissemble/issues/430 is resolved: : -->
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>exec-maven-plugin</artifactId>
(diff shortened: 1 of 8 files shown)


test/test-mda-models/test-data-delivery-spark-model/src/main/resources/pipelines/SparkJavaDataDeliveryPatterns.json
```diff
@@ -68,14 +68,6 @@
...
 			}
 		},
-		{
-			"name": "VoidInboundAndMessagingOutboundAsync",
-			"type": "asynchronous",
-			"outbound": {
-				"type": "messaging",
-				"channelName": "outboundChannel"
-			}
-		},
```
jaebchoi, line 78: I: Seems to be having some issue with async messaging outbound without inbound messaging.
added https://github.com/boozallen/aissemble/issues/430 to follow up.

test/test-mda-models/test-data-delivery-spark-model-basic/pom.xml
```diff
@@ -239,5 +236,17 @@
             <version>${version.cucumber.reporting.plugin}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jboss.resteasy</groupId>
+            <artifactId>resteasy-client</artifactId>
+            <version>${version.test.resteasy}</version>
```
jaebchoi, line 242: I: Upon testing downstream project with spark pipeline I don't see any issue with building and tilitng up.
Screenshot 2024-10-24 at 10 53 59 AM
Screenshot 2024-10-24 at 10 54 39 AM

ewilkins-csi: To follow up on this, it appears this is an issue with the `foundation-alerting` import.  There is likely some version incompatibility that was caused by the Java 17 upgrades.  It's worth noting that it seems there is a bug where the mda generation always things alerting is disabled even when it isn't.  We will likely need to handle both of these issues before releasing 1.10.

ewilkins-csi (approved):

PR #439, opened 2024-10-24
[#4] switch from docker compose to testcontainers

ewilkins-csi:
Our pyspark test metamodel module was using docker-compose to start containers for integration tests instead of our usual approach of using testcontainers. This caused a failure in CI as the GitHub runner doesn't appear to have docker-compose. This changeset also includes some fixes to the other places we were using the testcontainers Python package. Namely:
 - We now use dynamic ports in extensions-encryption-python
 - SafeDockerContainer workaround was removed as the core problem was resolved in testcontainers 4.0


extensions/extensions-encryption/aissemble-extensions-encryption-vault-python/src/aissemble_encrypt/vault_config.py
```diff
@@ -64,18 +64,23 @@ def secrets_host_url(self):
         Returns the secrets host url.
         This should only be used for integration tests, and not otherwise.
         """
-        return self.properties["secrets.host.url"]
+        if "SECRETS_HOST_URL" in os.environ:
```
ewilkins-csi, line 67: I: Allows us to use a dynamic port and assign it at test time.

extensions/extensions-encryption/aissemble-extensions-encryption-vault-python/src/aissemble_encrypt/vault_config.py
```diff
@@ -64,18 +64,23 @@ def secrets_host_url(self):
...
+            # host and port of the aissemble-vault container
+            return os.environ["SECRETS_HOST_URL"]
+        else:
+            return self.properties["secrets.host.url"]
 
     @staticmethod
-    def validate_container_start():
+    def validate_container_start(port):
```
ewilkins-csi, line 74: I/A: I'm a little skeptical about including this method here in the Config class instead of just in the test classes, but I didn't feel strongly enough to move it.

test/test-mda-models/aissemble-test-data-delivery-pyspark-model/tests/features/environment.py
```diff
@@ -33,15 +36,6 @@ def before_all(context):
 
     print("Created spark session for tests...")
 
-    os.environ["S3Test_FS_PROVIDER"] = "s3"
```
ewilkins-csi, line 36: I: Moved this to where we're we're setting the PORT so it's all together.

test/test-mda-models/aissemble-test-data-delivery-pyspark-model/tests/features/environment.py
```diff
@@ -63,47 +57,104 @@ def after_scenario(context, scenario):
...
+    version = packaging.version.parse(version_str)
+    tag = version.base_version
+    if version.pre:
+        tag += "-" + "".join([str(x) for x in version.pre])
+    if version.is_devrelease:
+        tag += "-SNAPSHOT"
+    return tag
+
+
+def test_aws(container):
```
ewilkins-csi, line 145: I: I think this is less necessary now that we are confining the services started to just s3, but when we first tried this there was a connection refused in CI because it tried to authenticate before localstack was fully started.

jocobtt (approved):

JeffreyRoss (approved):

Issue #429, opened 2024-10-23
Feature: Build action should re-use the docker build cache


ewilkins-csi:
## Description
We save the m2 repository, the maven build cache, and the poetry cache to GitHub caches so we can re-use them across builds. We should also do this with the docker build cache at `~/.docker/cache`.  Note that we are already consuming over the 10GB limit for GH caches.  The docker build cache is expected to be over 5GB and so could cause issues with our caches being purged too aggressively.  There will be a [feature early next year](https://github.com/orgs/community/discussions/42506#discussioncomment-10388640) to allow open source projects to pay for increased cache limits. Further, there is an [action available](https://github.com/marketplace/actions/fast-actions-cache-for-s3) that is a one-to-one replacement for the official cache action that allows writing to S3.

However, on further discussion, we realized that our current pattern of segregating the maven and poetry caches based on the contents of the projects POM files is unnecessary and causing a lot of excess cache usage.  If we remove this segregation and just use a constant cache, our cache usage should drop to 6GB.  This may be enough that we can get by floating right at the limit. So for the first pass at this, we'll just trim back our other cache usage and see if it becomes an issue. If it does, we can explore the options listed above.

## DOD
Acceptance criteria required to realize the requested feature
- [x] the `build.yaml` workflow is updated to create/use a `docker-cache` GH cache to prevent full docker image rebuilds
- ~the `build.yaml` workflow is updated to save the m2 repo, maven build cache, and poetry cache to a constant cache name~
  - This is not possible, as GH actions [does not allow](https://github.com/actions/toolkit/issues/505) a cache key to be updated
- [x] the `build.yaml` workflow is updated to use the `install-dependencies` reusable action to install necessary build tools
  - Note, we added the `install-dependencies` action in #3, but failed to use it in the build workflow as well

## Test Strategy/Script
1. Examine the logs for the most recent build
2. Verify there are stages for restoring the Docker cache and saving the Docker cache
3. Verify the "Install dependencies" action is used instead of individual steps for installing Java/Python/Helm/etc.
4. Ensure a cache named `docker-cache-*` is in [GH caches](https://github.com/boozallen/aissemble/actions/caches)

ewilkins-csi:
**DOD/OTS** with @csun-cpointe

carter-cundiff:
Testing passed:
![image](https://github.com/user-attachments/assets/70d2bc13-8885-4a2b-b3a5-6554325ba18f)
[...]

PR #440, opened 2024-10-25
[#429] add docker cache to build workflow

ewilkins-csi:
Saves the docker cache configured in the `cacheTo` and `cacheFrom` options of the Fabric8 `docker-maven-plugin` to GH caches to speed image rebuilds.  Also uses the `install-dependencies` custom action to de-duplicate the required build tool installation logic.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,24 +49,15 @@ jobs:
               })
             }
             console.log("Clear completed")
-      - name: Login to Docker Hub
-        uses: docker/login-action@v3
+      - name: Install dependencies
+        uses: ./.github/actions/install_dependencies
         with:
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_TOKEN }}
-      - name: Install required packages
-        run: |
-          sudo apt-get update
-          sudo apt install -y build-essential libssl-dev zlib1g-dev libbz2-dev libreadline-dev libsqlite3-dev curl \
-              git libncursesw5-dev xz-utils tk-dev libxml2-dev libxmlsec1-dev libffi-dev liblzma-dev zip unzip \
-              libpython3.11
-      - name: Install Python
-        uses: gabrielfalcao/pyenv-action@v18
-        with:
-          default: 3.11.4
+          docker-username: ${{ secrets.DOCKERHUB_USERNAME }}
+          docker-token: ${{ secrets.DOCKERHUB_TOKEN }}
+      #NB: We restore/save cache manually so that we save the cache even if the build fails
       - name: Load m2 repository cache # Manually caching .m2 repo as the setup-java caching isn't falling back to older caches
         id: cached-m2-repo
-        uses: actions/cache@v4
+        uses: actions/cache/restore@v4
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -49,63 +49,43 @@ jobs:
...
-        uses: gabrielfalcao/pyenv-action@v18
-        with:
-          default: 3.11.4
+          docker-username: ${{ secrets.DOCKERHUB_USERNAME }}
+          docker-token: ${{ secrets.DOCKERHUB_TOKEN }}
+      #NB: We restore/save cache manually so that we save the cache even if the build fails
       - name: Load m2 repository cache # Manually caching .m2 repo as the setup-java caching isn't falling back to older caches
         id: cached-m2-repo
-        uses: actions/cache@v4
+        uses: actions/cache/restore@v4
```
ewilkins-csi, line 60: I: By using the `restore` granular action, we prevent the automated cache-save logic that doesn't meet our needs of saving even when the build fails. (Prior to this change we were basically always saving these caches twice.)

.github/workflows/build.yml
```diff
@@ -123,19 +103,24 @@ jobs:
       - name: Run Archetype Tests
         run: |
           ./mvnw -B clean install -Parchetype-test -pl :foundation-archetype
-      #NB: The following two explicit cache saves are necessary to ensure caches are saved on build failure,
-      # until https://github.com/actions/cache/issues/1315 is resolved
```
ewilkins-csi, line 127: I: This issue is unlikely to be resolved any time soon as there are underlying restrictions that prevent it.  Instead, we follow [the guidance](https://github.com/actions/cache/issues/1315#issuecomment-2163822166) of the action maintainers.

.github/workflows/build.yml
```diff
@@ -49,63 +49,45 @@ jobs:
...
+          restore-keys: |
+            docker-cache-
+      #NB: Not saving poetry cache on failure in case it's a failure caused by an in-flight python package release
       - name: Poetry cache
         id: cached-poetry
         uses: actions/cache@v4
         with:
           path: ~/.cache/pypoetry
-          key: poetry-cache-${{ hashFiles('**/pom.xml') }}
+          key: poetry-cache-${{ hashFiles('**/pyproject.toml') }}
```
ewilkins-csi, line 88: I: Use pyproject.toml so the cache is updated when the python dependencies change, even if the POMs have not changed.

.github/workflows/build.yml
```diff
@@ -49,63 +49,45 @@ jobs:
...
           restore-keys: |
             maven-build-cache-
-      - name: Install Poetry
-        uses: snok/install-poetry@v1
+      - name: Load docker build cache
+        id: cached-docker-build
+        uses: actions/cache/restore@v4
+        with:
+          path: ~/.docker/cache
+          key: docker-cache-${{ hashFiles('**/Dockerfile') }}
```
ewilkins-csi, line 79: I: This has some limitations, as it doesn't also include the files that are copied into the docker images (e.g. our python modules) but this should be close enough to the ideal case without a ton of complexity.

carter-cundiff (approved):

Issue #442, opened 2024-10-28
BUG: Delta Spark migrations fail for some YAML & POM files


ewilkins-csi:
## Description
The migration to update YAML files with the new `delta-spark` dependencies as part of the Java 17 upgrade does not gracefully handle YAML files that do not have maps/objects as their root. (I.e. a YAML file which is a list.)

## Steps to Reproduce
1. Create a new project from the 1.10.0-SNAPSHOT archetype
2. Add a yaml file that is a list, e.g.:
    ```yaml
    - name: test1
      desc: "A test item"
    - name: test2
      desc: "Another test item"
    ```
3. Run the build: `mvn clean install`

## Expected Behavior
The migration should simply skip the inapplicable YAML file and finish successfully.

## Actual Behavior
The build fails at the root project.

## Additional Context
Error:
```
Execution default of goal org.technologybrewery.baton:baton-maven-plugin:1.1.1:baton-migrate failed: Could not complete migrations!: class java.util.ArrayList cannot be cast to class java.util.Map (java.util.ArrayList and java.util.Map are in module java.base of loader 'bootstrap')
```
Stack trace:
```
Caused by: java.lang.ClassCastException: class java.util.ArrayList cannot be cast to class java.util.Map (java.util.ArrayList and java.util.Map are in module java.base of loader 'bootstrap')
    at com.boozallen.aissemble.upgrade.util.YamlUtils.loadYaml (YamlUtils.java:46)
    at com.boozallen.aissemble.upgrade.migration.v1_10_0.DeltaSparkYamlMigration.shouldExecuteOnFile (DeltaSparkYamlMigration.java:38)
    at org.technologybrewery.baton.AbstractMigration.execute (AbstractMigration.java:73)
    at org.technologybrewery.baton.BatonMojo.performMigration (BatonMojo.java:226)
    at org.technologybrewery.baton.BatonMojo.execute (BatonMojo.java:180)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:126)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2 (MojoExecutor.java:328)
```

## Definition of Done
 - Following the steps to reproduce causes the expected behavior
 - The NPE caused by a missing `dependenciesManagement` section during the `DeltaSparkPomMigration` no longer occurs

## Test Steps
1. Create a new project from the 1.10.0-SNAPSHOT archetype
[...]

PR #443, opened 2024-10-28
[#442] account for lists in `YamlUtils#loadYaml`

ewilkins-csi:
Also fixes a bug with the `hasDouble`, `hasInt`, and `hasBoolean` methods of `YamlObject`. Adds tests for `YamlUtils` including one to catch the root list issue. Finally, also updates the POM migration to account for POMs without dependency management sections.

### DIFF
--- a/foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/util/YamlUtils.java
+++ b/foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/util/YamlUtils.java
@@ -32,6 +32,7 @@ import org.yaml.snakeyaml.Yaml;
 import org.yaml.snakeyaml.DumperOptions;
 import org.yaml.snakeyaml.LoaderOptions;
 import org.yaml.snakeyaml.constructor.Constructor;
+import org.yaml.snakeyaml.error.YAMLException;
 import org.yaml.snakeyaml.introspector.Property;
 import org.yaml.snakeyaml.introspector.PropertyUtils;
 import org.yaml.snakeyaml.representer.Representer;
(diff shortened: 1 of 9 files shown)


foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/util/YamlUtils.java
```diff
@@ -78,23 +97,23 @@ public String getString(String... path) {
         }
 
         public boolean hasInt(String... path) {
-            return hasValue(int.class, path);
+            return hasValue(Integer.class, path);
```
ewilkins-csi, line 100: I: these were returning false when they shouldn't have as the map has the boxed version of the value so `isInstance` was failing.

foundation/foundation-upgrade/src/main/resources/migrations.json
```diff
@@ -73,7 +73,8 @@
         "implementation": "com.boozallen.aissemble.upgrade.migration.v1_10_0.DeltaSparkYamlMigration",
         "fileSets": [
           {
-            "includes": ["**/*.yaml"]
+            "includes": ["**/*.yaml"],
+            "excludes": ["**/target/**", "**/templates/**"]
```
ewilkins-csi, line 77: I: Not strictly necessary but this cuts down significantly on the number of output lines talking about un-parseable helm templates.

foundation/foundation-upgrade/src/test/java/com/boozallen/aissemble/upgrade/migration/utils/YamlUtilsSteps.java
```diff
@@ -0,0 +1,96 @@
...
+        String type = yaml.get(name).getClass().getSimpleName();
+        assertTrue("Property was not loaded as string: " + name + " (was " + type + ")", yaml.hasString(name));
+        assertEquals("Property value does not match for: " + name, value, yaml.getString(name));
+    }
+
+    @Then("the decimal value of the property {string} is {double}")
+    public void theDecimalValueOfThePropertyIs(String name, double value) {
+        String type = yaml.get(name).getClass().getSimpleName();
+        assertTrue("Property was not loaded as double: " + name + " (was " + type + ")", yaml.hasDouble(name));
+        assertEquals("Property value does not match for: " + name, value, yaml.getDouble(name), 0.01);
```
csun-cpointe, line 53: Q: `0.01`?

csun-cpointe (approved):

foundation/foundation-upgrade/src/test/java/com/boozallen/aissemble/upgrade/migration/utils/YamlUtilsSteps.java
ewilkins-csi: It's a `delta` param that they added to floating point assertions. Avoids the tricky floating point rounding issues where you get `0.000000000001` or some other minuscule difference just from how FP is represented. So this says treat the numbers as equal if they are within `0.01` of each other.

carter-cundiff (approved):

PR #446, opened 2024-10-30
#430 FIx data delivery steps not injecting smallrye when running test-data-delivery-spark-model

jaebchoi:
We have disabled some data delivery steps as those steps are not injecting smallrye reactive messaging correctly.

It seems like any step with asynchronous type, no inbound messaging , and messaging outbound is having issue with running step execution.

Issue https://github.com/boozallen/aissemble/issues/430

### DIFF
--- a/foundation/foundation-mda/src/main/resources/templates/data-delivery-spark/asynchronous.processor.base.java.vm
+++ b/foundation/foundation-mda/src/main/resources/templates/data-delivery-spark/asynchronous.processor.base.java.vm
@@ -26,7 +26,7 @@ import org.reactivestreams.Subscription;
 #if (${step.outbound})
 import org.eclipse.microprofile.reactive.messaging.Outgoing;
 #end
-#if ((${step.hasMessagingOutbound()} && ${step.hasNativeInbound()}))
+#if ((${step.hasMessagingOutbound()} && ${step.useEmitter()}))
 import org.eclipse.microprofile.reactive.messaging.Channel;
 import org.eclipse.microprofile.reactive.messaging.Emitter;
 #end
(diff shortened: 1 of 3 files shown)


foundation/foundation-mda/src/main/resources/templates/data-delivery-spark/asynchronous.processor.base.java.vm
```diff
@@ -26,7 +26,7 @@ import org.reactivestreams.Subscription;
 #if (${step.outbound})
 import org.eclipse.microprofile.reactive.messaging.Outgoing;
 #end
-#if ((${step.hasMessagingOutbound()} && ${step.hasNativeInbound()}))
+#if ((${step.hasMessagingOutbound()} && (${step.hasNativeInbound()} ||  ${step.hasVoidInbound()} )))
```
jaebchoi, line 29: I: I tried emitter instead of directly putting outgoing channel in method by referring this test [OutOnlyTestStep](https://github.com/boozallen/aissemble/blob/dev/extensions/extensions-data-delivery/extensions-data-delivery-spark/src/test/java/com/boozallen/aiops/data/delivery/messaging/pipeline/steps/OutOnlyTestStep.java) 


upon changing build seems to pass, also ran downstream project with problematic metamodel, seems to pass build as well as inspection of spark pipeline to make sure it succeeded. 


Screenshot 2024-10-30 at 9 55 58 AM
Screenshot 2024-10-30 at 9 56 11 AM

ewilkins-csi:
Do we know _when_ this started to fail?  Prior to 1.7, we were running the tests and this wasn't an issue.  I mainly ask because the changes made here are things I would have expected to be around for awhile. So curious what changed since 1.7 that would require the new checks.

foundation/foundation-mda/src/main/resources/templates/data-delivery-spark/asynchronous.processor.base.java.vm
```diff
@@ -137,7 +137,7 @@ public abstract class ${step.capitalizedName}Base extends AbstractPipelineStep {
     #end
 #end
 
-#if ($step.hasMessagingOutbound() && $step.hasNativeInbound())
+#if ($step.hasMessagingOutbound() && ($step.hasNativeInbound() || $step.hasVoidInbound()))
```
ewilkins-csi, line 140: A: Just to help with maintenance, we should pull this boolean term out into a variable like `useEmitter` so that if we need to change this calculation again we only do it in one place.

ewilkins-csi (approved):

Issue #434, opened 2024-10-24
Feature: Create baton migration for Quarkus Java 17 upgrade


carter-cundiff:
## Description
Follow on to #263, where we migrated our quarkus projects to java 17. This issue will focus on migrating downstream projects to be compatible with the new quarkus dependencies.

Questions:
- Archetype change: `<groupId>${groupId}</groupId>` -> `<groupId>${project.groupId}</groupId>` - worth migrating?
  - Wait until maven 4 upgrade
- What if version no longer matches the expected version?
  - Update regardless

## DOD
- [x] Migration to update downstream dependencies to include new version - run on every `pom.xml`:
  - [x] `org.scala-lang:scala-reflect:2.12.1` -> `${version.scala}`
  - [x] `org.apache.commons:commons-math3:3.2` -> `${version.commons.math}`
  - [x] `io.smallrye.reactive:smallrye-reactive-messaging-kafka` -> `${version.smallrye.reactive.messaging}`
  - [x] `junit:junit:4.13.2` -> `${version.junit}`
  - [x] `io.smallrye.reactive:smallrye-reactive-messaging-in-memory` -> `${version.smallrye.reactive.messaging}`
  - [x] `org.jboss.resteasy:resteasy-client` -> `${version.resteasy}`
  - [x] `org.jboss.resteasy:resteasy-jackson2-provider` -> `${version.resteasy}`
  - [x] `org.awaitility:awaitility` ->`${version.awaitility}`
  - [x] `io.vertx:vertx-core` -> `${version.vertx}`
  - [x] `io.smallrye.config:smallrye-config:2.10.0` -> `${version.smallrye.config}`
  - [x] `org.slf4j:slf4j-api:2.0.7` -> `${version.slf4j}`
  - [x] `org.slf4j:slf4j-simple:1.6.4` -> `${version.slf4j}`
   ```
   Scenario: Migrate all pom dependencies to the new version
   Scenario: Migrate pom dependencies in profiles and dependency management
   ```

- [x] Migration to update downstream pom to add the following dependencies - run on Spark pipeline `pom.xml`:
  - [x]  `javax.servlet:javax.servlet-api:${version.javax.servlet}` - run migration only on pipelines directory
   ```
   Scenario: Add javax servlet dependency spark pipeline pom 
   Scenario: Skip spark pipeline pom that already has the javax servlet dependency
   Scenario: Skip pom that does not have jar packaging
   ```

[...]

PR #448, opened 2024-10-30
#434 Create baton migration for Quarkus Java 17 upgrade

carter-cundiff:


DRAFT_RELEASE_NOTES.md
```diff
@@ -1,12 +1,13 @@
 # Major Additions
 
-## TBD
+## Java 17 Upgrade
+The aiSSEMBLE project is now built with Java 17. All aiSSEMBLE managed dependencies have been updated to a Java 17 compatible version and automatic migrations (detailed below) have been created to streamline this transition for downstream projects. It is recommended to review all non aiSSEMBLE managed dependencies and custom logic within your project, as this upgrade can break existing functionality due to deprecated classes and incompatible dependencies.
```
csun-cpointe, line 5: 👍

foundation/foundation-mda/src/main/resources/templates/cucumber.test.cdi.context.java.vm
```diff
@@ -15,7 +15,7 @@ import io.smallrye.reactive.messaging.memory.InMemoryConnector;
  *
  * Originally generated from: ${templateName} 
  */
-public class TestCdiContext extends PipelinesCdiContext {
+public class TestCdiContext implements CdiContext {
```
csun-cpointe, line 18: Q: Do we not need the `PipelineMessagingConfig` ?

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_10_0/SparkPipelineMessagingPomMigration.java
```diff
@@ -66,18 +66,6 @@ protected boolean performMigration(File pomFile) {
...
-    private boolean hasDependency(Model model, String groupId, String artifactId) {
-        return model.getDependencies()
-             .stream()
-             .anyMatch(dependency ->
-                 dependency.getGroupId().equals(groupId) &&
-                 dependency.getArtifactId().contains(artifactId)
-             );
-    }
-
     private String getDependencyContent(String ignore) {
```
csun-cpointe, line 69: Q: should the `ignore` param be removed?

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_10_0/SparkPipelineServletApiMigration.java
```diff
@@ -0,0 +1,65 @@
...
+        Model model = PomHelper.getLocationAnnotatedModel(pomFile);
+
+        this.detectAndSetIndent(pomFile);
+        PomModifications modifications = new PomModifications();
+        modifications.add(new Insertion(model.getLocation("dependencies" + END), 1, this::getDependencyContent));
+
+        return PomHelper.writeModifications(pomFile, modifications.finalizeMods());
+    }
+    
+    private String getDependencyContent(String ignore) {
```
csun-cpointe, line 56: same here

foundation/foundation-upgrade/src/test/java/com/boozallen/aissemble/upgrade/migration/v1_10_0/QuarkusBomMigrationSteps.java
```diff
@@ -0,0 +1,45 @@
...
+
+public class QuarkusBomMigrationSteps extends AbstractMigrationTest {
+    @Given("a POM that references the {string} Quarkus BOM")
+    public void a_pom_that_references_the_quarkus_bom(String quarkusBomType) {
+        testFile = getTestFile("v1_10_0/QuarkusBomMigration/migration/" + quarkusBomType + "-pom.xml");
+    }
+
+    @Given("a POM that does not contain a Quarkus BOM")
+    public void a_pom_that_does_not_contain_a_quarkus_bom() {
+        testFile = getTestFile("v1_10_0/QuarkusBomMigration/validation/quarkus-bom-pom.xml");
```
csun-cpointe, line 27: Q: This seems not right. It's validating the same file?

csun-cpointe (approved):

build-parent/pom.xml
```diff
@@ -68,6 +68,7 @@
         <version.plexus.util>3.5.1</version.plexus.util>
         <version.jackson.mapper.asl>1.9.3</version.jackson.mapper.asl>
         <version.exec.maven.plugin>3.4.1</version.exec.maven.plugin>
+        <version.commons.math>3.6.1</version.commons.math>
```
ewilkins-csi, line 71: A: We should pull this up to the root POM with the other `apache-commons-*` dependencies

foundation/foundation-mda/src/main/resources/templates/cucumber.test.cdi.context.java.vm
ewilkins-csi: It will be included automatically by virtue of the typical runtime CDI configuration.  This CDI context is only for adding test-specific on top of the runtime configuration.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/AbstractPomMigration.java
```diff
@@ -40,4 +53,54 @@ protected void detectAndSetIndent(File file) {
...
+        for (Profile profile : model.getProfiles()) {
+            matchingDependencies.addAll(this.getAllDependenciesByFilter(profile, filter));
+        }
+        return matchingDependencies;
+    }
+
+    /**
+     * Checks the dependency and dependency management section of a given pom model for any matches
+     */
+    protected List<Dependency> getAllDependenciesByFilter(ModelBase model, Predicate<? super Dependency> filter) {
```
ewilkins-csi, line 83: A:
```suggestion
    protected List<Dependency> getMatchingDependencies(ModelBase model, Predicate<? super Dependency> filter) {
```
and similarly `getMatchingDependenciesForProject`

ewilkins-csi, line 83: I'd also consider making this private since it's really just a helper to `getMatchingDependenciesForProject`.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_10_0/JavaPackageMigration.java
```diff
@@ -0,0 +1,89 @@
...
+
+import com.boozallen.aissemble.upgrade.migration.AbstractAissembleMigration;
+
+/**
+ * Updates the affected java classes from their old package name to their new package name 
+ * to ensure compatibility with the updated Java 17 dependencies 
+ */
+public class JavaPackageMigration extends AbstractAissembleMigration {
+    private static final Logger logger = LoggerFactory.getLogger(JavaPackageMigration.class);
+    private static final String[][] JAVA_PACKAGES = {
```
ewilkins-csi, line 30: A: I feel like this becomes easier to track in the actual logic as a `Map.of` instead of a list of lists.

foundation/foundation-upgrade/src/test/resources/specifications/v1_10_0/quarkus-bom-migration.feature
```diff
@@ -0,0 +1,16 @@
+Feature: Updates all references to the quarkus-bom and quarkus-universe-bom to use the new aissemble-quarkus-bom for managing Quarkus dependencies
```
ewilkins-csi, line 1: A: This isn't that important, especially as we aren't currently publishing test results, but generally it's good to keep the Feature name fairly short, like a title. You can expand on the purpose or detail of the feature in the description.  This makes the readout of the test a results a bit easier to navigate/understand in my experience.

foundation/foundation-upgrade/src/main/resources/migrations.json
```diff
@@ -77,6 +88,34 @@
...
+            "includes": ["pom.xml"]
+          }
+        ]
+      },
+      {
+        "name": "java-package-migration",
+        "implementation": "com.boozallen.aissemble.upgrade.migration.v1_10_0.JavaPackageMigration",
+        "fileSets": [
+          {
+            "includes": ["**/*.java"],
```
ewilkins-csi, line 106: A: We could have this run only on classes within active modules by restricting this fileset to only `src`. That way we don't grab all `.java` files in the root execution and instead catch each in the respective execution.
```suggestion
            "includes": ["src/**/*.java"],
```

I'm also realizing that I really could have done this with the YAML migration. I was worried about Java vs Python modules, but it still could have been achieved with just a couple of filesets.

foundation/foundation-upgrade/src/test/java/com/boozallen/aissemble/upgrade/migration/v1_10_0/SparkPipelineServletApiMigrationSteps.java
```diff
@@ -0,0 +1,52 @@
...
+import com.boozallen.aissemble.upgrade.migration.AbstractMigrationTest;
+
+import io.cucumber.java.en.Given;
+import io.cucumber.java.en.Then;
+import io.cucumber.java.en.When;
+
+public class SparkPipelineServletApiMigrationSteps extends AbstractMigrationTest {
+    @Given("a POM with {string} packaging type")
+    public void a_pom_with_packaging_type(String packaging) {
+        if (!packaging.equals("jar")) {
```
ewilkins-csi, line 22: Q: It doesn't really matter, but curious why you didn't follow the same pattern as the Quarkus migration and just use the clause parameter as part of the file name.

DRAFT_RELEASE_NOTES.md
```diff
@@ -1,12 +1,13 @@
...
```
ewilkins-csi, line 4:
```suggestion
The aiSSEMBLE project is now built with Java 17. All dependencies inherited from aiSSEMBLE have been updated to a Java 17 compatible version and automatic migrations (detailed below) have been created to streamline this transition for downstream projects. It is recommended to review noninherited dependencies and custom logic within your project, as this upgrade can break existing functionality due to deprecated classes and incompatible dependencies.
```

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_10_0/SparkPipelineMessagingPomMigration.java
carter-cundiff: The content producer methods for all the pom modification are required to take a string parameter. The parameter was initially meant to be the indent size, but we are setting the indent earlier in the migration and using our own instead. So we can safely ignore the string parameter, but still have to accept it as part of the method.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_10_0/SparkPipelineServletApiMigration.java
carter-cundiff: see above

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_10_0/SparkPipelineMessagingPomMigration.java
csun-cpointe: Gotcha. Thanks for the explanation.

DRAFT_RELEASE_NOTES.md
```diff
@@ -31,11 +32,15 @@ To reduce burden of upgrading aiSSEMBLE, the Baton project is used to automate t
...
-| log4j-maven-shade-plugin-migration                 | Updates the Maven Shade Plugin with the new Log4j dependency information                                                                                                     |
+| log4j-maven-shade-plugin-migration                 | Updates the Maven Shade Plugin with the new Log4j dependency information                                                                             |
+| quarkus-bom-migration                              | Updates all references to the `quarkus-bom` and `quarkus-universe-bom` to use the new `aissemble-quarkus-bom` for managing Quarkus dependencies                               |
+| pom-dependency-version-migration                   | Updates the pom dependencies previously managed by the aiSSEMBLE `bom-component` to include their necessary versions                                                                |
+| java-package-migration                             | Updates the affected java classes from their old package name to their new package name to ensure compatibility with the updated Java 17 dependencies                              |
```
ewilkins-csi, line 43: Q: Should we say either here or in the block at the top that this is largely Javax -> Jakarta?

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_10_0/JavaPackageMigration.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Override
+    protected boolean performMigration(File file) {
+        logger.info("Migrating file to update java class packages: {}", file.getAbsolutePath());
+        try {
+            String fileContent = Files.readString(file.toPath());
+
+            for (String[] packagePair : JAVA_PACKAGES) {
+                fileContent = fileContent.replace(packagePair[0], packagePair[1]);
```
ewilkins-csi, line 79: Q/A: We do have some sort of utility methods around replacing a string or pattern in a file. Is that functionality a good fit here? Or maybe it's too heavy handed?  Seems like we do this sort of thing enough that it would be a one liner.

foundation/foundation-upgrade/src/test/resources/specifications/v1_10_0/quarkus-bom-migration.feature
carter-cundiff: Good to know!

DRAFT_RELEASE_NOTES.md
carter-cundiff: Thinking we put it as part of the migration description, feels sort of out of place at the high level block

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_10_0/JavaPackageMigration.java
carter-cundiff: There is a baton `replaceInFile` method, but it writes the file out with each call. The way I'm currently doing it allows us to make all the necessary changes, then write out the file once at the end.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_10_0/PomDependencyVersionMigration.java
```diff
@@ -0,0 +1,147 @@
...
+            .toList();
+    }
+
+    /**
+     * Returns a list of any dependencies that need to be updated with a new version
+     */
+    private List<Dependency> getMatchingDependenciesWithDifferentVersionsForProject(Model model) {
+        List<Dependency> matchingDependencies = new ArrayList<>();
+
+        this.dependenciesToUpdate.stream()
```
ewilkins-csi, line 111: A: Why not invert this, and let the project dependency collection be the "outer" stream (and handled by the parent class method). Then `isMatchingDependencyWithDifferentVersion`, could be more like `dependencyRequiresUpdate`. Then _it_ streams the `dependenciesToUpdate` looking for a match.  That would also allow you to break out the different aspects of the filter, first comparing groupId, then artifactId, and lastly version.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_10_0/PomDependencyVersionMigration.java
```diff
@@ -0,0 +1,147 @@
...
+
+    /**
+     * Returns the updated version for the matching dependency
+     */
+    private String getUpdatedDependencyVersion(Dependency dependency) {
+        return this.dependenciesToUpdate.stream()
+            .filter(updatedDependency -> this.isMatchingDependency(updatedDependency, dependency))
+            .map(Dependency::getVersion)
+            .findFirst()
+            .orElse(null);
```
ewilkins-csi, line 145: S: Looks like the calling code assumes non-null. Should we use `orElseThrow` instead?

carter-cundiff: Yes that should throw an error if we don't find the version.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/AbstractPomMigration.java
carter-cundiff: Updated

carter-cundiff: Updated

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_10_0/JavaPackageMigration.java
carter-cundiff: Updated to use maps. Fun fact, `Map.of` only takes 10 entries max, while `Map.ofEntries` takes unlimited

foundation/foundation-upgrade/src/main/resources/migrations.json
carter-cundiff: Updated

DRAFT_RELEASE_NOTES.md
carter-cundiff: Updated

DRAFT_RELEASE_NOTES.md
carter-cundiff: Added to migration description

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_10_0/PomDependencyVersionMigration.java
```diff
@@ -0,0 +1,147 @@
...
+            "org.awaitility:awaitility:${version.awaitility}",
+            "io.vertx:vertx-core:${version.vertx}",
+            "io.smallrye.config:smallrye-config:${version.smallrye.config}",
+            "org.slf4j:slf4j-api:${version.slf4j}",
+            "org.slf4j:slf4j-simple:${version.slf4j}"
+        };
+
+        return Arrays.stream(dependencyStrings)
+            .map(dependencyString -> {
+                String[] dependencyAttributes = dependencyString.split(":");
```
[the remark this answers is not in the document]
carter-cundiff: Updated to use `dependency()` method and initialize at the attribute level similar to the JavaPackageMigration

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_10_0/PomDependencyVersionMigration.java
carter-cundiff: Updated to stream all dependencies initially, and stream dependenciesToUpdate in the helper method

build-parent/pom.xml
carter-cundiff: Well previously the commons was hardcoded in downstream projects like this:
```
        <dependency>
            <groupId>org.apache.commons</groupId>
            <artifactId>commons-math3</artifactId>
            <version>3.2</version>
        </dependency>
```
So rather than modify the entire migration logic to account for a to delete the version in this one case, figure it's fine to just use our new project `${version.commons.math}`

carter-cundiff: Logic actually wouldn't be too much more complicated, so could go either way on this. I lean towards adding redundant version `<version>${version.commons.math}</version>` so that it's more obvious to the user where the version is being set.

ewilkins-csi (approved):

build-parent/pom.xml
carter-cundiff: Went ahead and made the change to just remove the old version. Cleans up the baseline as well since we can use the managed dependency instead

Issue #451, opened 2024-11-01
BUG: Build fails when unneeded dependency is removed


ewilkins-csi:
## Description
The JBoss RESTEasy client version set in the bom-component is incompatible with some other libraries.  This isn't noticeable with a default project, because the `extensions-authzforce` dependency is always added to pipeline POMs even when it isn't needed. However, removing that unneeded dependency ends up rolling the REST Easy client back to an incompatible version.

## Steps to Reproduce
1. Create a downstream project with a Spark pipeline
2. Generate out the project until the pipeline POM is part of the build
3. Remove the `extensions-authzforce` dependency from the pipeline POM
4. Build the project

## Expected Behavior
The build should not fail because an unused dependency is removed.

## Actual Behavior
The build fails with the unneeded dependency removed.

## Definition of Done
 - Update JBoss REST Easy version in parent POM to compatible version
 - The bom-component is totally removed and the above steps to reproduce do not cause an `AbstractMethodError`
--- Stretch Goal ---
 - When alerting is enabled for a pipeline step, the `foundation-alerting` dependency is added to the pipeline POM

## Test Steps
1. Create a new project
```sh
mvn archetype:generate '-DarchetypeGroupId=com.boozallen.aissemble' \
                       '-DarchetypeArtifactId=foundation-archetype' \
                       '-DarchetypeVersion=1.10.0-SNAPSHOT' \
                       '-DgroupId=org.test' \
                       '-Dpackage=org.test' \
                       '-DprojectGitUrl=test.org/test.git' \
                       '-DprojectName=Test Project' \
                       '-DartifactId=final-451' \
&& cd final-451
```
2. Copy [SparkPipeline.json](https://github.com/user-attachments/files/17620771/SparkPipeline.json) into _final-451-pipeline-models/src/main/resources/pipelines_
3. Run the build once and follow manual actions
4. Edit _final-451-pipelines/spark-pipeline/pom.xml_ and remove the following dependency:
```xml
<dependency>
     <groupId>com.boozallen.aissemble</groupId>
     <artifactId>extensions-authzforce</artifactId>
</dependency>
```
5. Build the project again and verify the pipeline tests pass without errors:
```sh
08:50:19 [INFO] --- surefire:3.5.0:test (default-test) @ spark-pipeline ---
08:50:19 [INFO] Using auto detected provider org.apache.maven.surefire.junit4.JUnit4Provider
08:50:19 [INFO] 
08:50:19 [INFO] -------------------------------------------------------
[...]

PR #453, opened 2024-11-04
[#451] remove bom-component

ewilkins-csi:
Fixes a dependency version mismatch for pipelines when the `extensions-authzforce` dependency is removed by removing the bom-component dependency management, which has long been deprecated. This also updates the Spark pipeline MDA test POMs to align more with what they would be in a newly generated project.


build-parent/pom.xml
```diff
@@ -161,9 +161,6 @@
 
         <!-- OpenLineage java -->
         <version.open.lineage.java>1.23.0</version.open.lineage.java>
-
-        <!-- Used for pinning version specifically test-mda-model-->
-        <version.test.resteasy>3.15.6.Final</version.test.resteasy>
```
ewilkins-csi, line 166: I: No longer needed, as this was just a workaround for the version mismatch issue in the test-mda modules.

extensions/extensions-security/extensions-authzforce/pom.xml
```diff
@@ -111,40 +103,12 @@
...
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>info.cukes</groupId>
-            <artifactId>cucumber-java</artifactId>
-            <version>1.2.6</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>info.cukes</groupId>
```
ewilkins-csi, line 121: I: Just some general cleanup. We weren't using these dependencies.

foundation/foundation-mda/src/main/resources/templates/data-delivery-spark/pipeline.pom.xml.vm
```diff
@@ -166,21 +166,21 @@
...
-			<groupId>org.apache.sedona</groupId>
-			<artifactId>sedona-viz-3.0_2.12</artifactId>
-			<version>${version.sedona}</version>
-		</dependency>
-		<dependency>
-			<groupId>org.datasyslab</groupId>
-			<artifactId>geotools-wrapper</artifactId>
-			<version>${version.geotools.wrapper}</version>
-		</dependency>
+        <dependency>
```
ewilkins-csi, line 169: I: Just normalizing to spaces instead of tabs

pom.xml
```diff
@@ -103,6 +103,11 @@
                 <artifactId>commons-lang3</artifactId>
                 <version>3.17.0</version>
             </dependency>
+            <dependency>
+                <groupId>org.apache.commons</groupId>
+                <artifactId>commons-math3</artifactId>
+                <version>3.6.1</version>
```
ewilkins-csi, line 109: A: This should be removed since #448 already added this

test/test-mda-models/test-data-delivery-spark-model-basic/pom.xml
ewilkins-csi, line 1: I: As alluded to in the description, this aligns the dependencies of the basic Spark model test with those which would be output by MDA generation/`foundation-mda`. There are small tweaks like using `project.groupId` instead of the hard-coded `com.boozallen.aissemble`, but otherwise the list is identical. This same comment applies to the other test model POM.

carter-cundiff (approved):

Issue #373, opened 2024-09-26
BUG: Spark Infrastructure chart uses a different version of Spark than Spark Operator/Spark Worker


ewilkins-csi:
## Description
Spark Operator and Spark Worker deployments are both based on the `aissemble/spark` image.  However, the spark infrastructure chart just pulls in `apache/spark:3.5.1` directly instead of using the aissemble version. This results in Spark Operator and Spark Worker being built on/with Spark 3.4, but Spark Infrastructure running 3.5.  This is not causing any immediate issues, but as the two versions drift apart more incompatibilities could arise.

ewilkins-csi:
**DoD/OTS** with @carter-cundiff

carter-cundiff:
Testing passed:
Original version:
![image](https://github.com/user-attachments/assets/30fc8d0b-3e4f-4a7c-8be9-b3162dc76ba7)
![image](https://github.com/user-attachments/assets/3f4d23c1-bdfa-467c-8639-a40d76968717)
![image](https://github.com/user-attachments/assets/4659bb68-1953-48ba-9119-bb082b0793a7)

New version:
![image](https://github.com/user-attachments/assets/7ece43a2-ffb5-41ac-b704-c05efc8b0afd)
![image](https://github.com/user-attachments/assets/104e9e7a-dcba-401d-b70d-644ff4e7db91)
![image](https://github.com/user-attachments/assets/7c052fc0-8b15-4ea9-a396-71af0933a168)

PR #455, opened 2024-11-04
[#373] use .Chart.AppVersion for image tags in custom charts

ewilkins-csi:
We have a pattern where we use the maven replacer plugin and a templatized `values.yaml` file to inject the project version at build time into our charts. This is the only way to provide the aissemble version as the image tag to a subchart. However, in many of the cases we've applied this pattern, the chart or the subchart is actually a custom chart made by us. This means we can simply modify the chart's template files to default the tag to the Chart version, which is set dynamically at build time.  This is greatly preferred as it's much more simple and standard Helm, and it avoids build cache issues. (E.g., if the Quarkus chart is cached, only the templatized values file exists on rebuild causing errors in charts that depend on the Quarkus chart.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,6 +55,14 @@ jobs:
           docker-username: ${{ secrets.DOCKERHUB_USERNAME }}
           docker-token: ${{ secrets.DOCKERHUB_TOKEN }}
       #NB: We restore/save cache manually so that we save the cache even if the build fails
+      - name: Load docker build cache
+        id: cached-docker-build
+        uses: actions/cache/restore@v4
+        with:
+          path: ~/.docker/cache
+          key: docker-cache-${{ hashFiles('**/Dockerfile') }}
+          restore-keys: |
+            docker-cache-
       - name: Load m2 repository cache # Manually caching .m2 repo as the setup-java caching isn't falling back to older caches
         id: cached-m2-repo
         uses: actions/cache/restore@v4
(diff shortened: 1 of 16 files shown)


extensions/extensions-helm/aissemble-fastapi-chart/values-dev.yaml
ewilkins-csi, line 1: Not sure why this chart was here as it's not used and isn't helpful as an example.

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-thrift-server-chart/templates/deployment.yaml
```diff
@@ -38,15 +38,15 @@ spec:
       {{- if or (not (empty .Values.dependencies.packages)) (not (empty .Values.dependencies.jars)) }}
       initContainers:
         - name: "populate-thrift-service-jar-volume"
-          image: {{ .Values.deployment.image.repository }}:{{ .Values.deployment.image.tag }}
+          image: {{ .Values.deployment.image.repository }}:{{ .Values.deployment.image.tag | default .Chart.AppVersion }}
```
ewilkins-csi, line 41: I: This is the ingredient that allows us to avoid templatizing the values files for our custom charts.

.gitignore
```diff
@@ -64,9 +64,6 @@ docs/antora-aissemble-ui/public/
 **/charts/*
 Chart.lock
 extensions/extensions-helm/aissemble-spark-operator-chart/values.yaml
-extensions/extensions-helm/aissemble-fastapi-chart/values.yaml
```
ewilkins-csi, line 67: I: Each of these charts has the same pattern:
1. remove the values file from gitignore
2. rename values.template.yaml to values.yaml
3. set the tag to an empty string in values.yaml
4. remove the replacer plugin execution from the POM

carter-cundiff (approved):

.github/workflows/build.yml
```diff
@@ -55,6 +55,14 @@ jobs:
           docker-username: ${{ secrets.DOCKERHUB_USERNAME }}
           docker-token: ${{ secrets.DOCKERHUB_TOKEN }}
       #NB: We restore/save cache manually so that we save the cache even if the build fails
+      - name: Load docker build cache
```
ewilkins-csi, line 58: I: Just pushed this change up in light of the recent build starting from scratch. It looks like since the Docker cache is saved last, GH purges caches from oldest to newest, and that the docker cache is larger than the limit it's causing all of our caches to be cleared. (It's also possible someone manually deleted, but making this change just to be safe.)

We likely just need to switch to the s3 caching action so we have no size limit.

Issue #457, opened 2024-11-05
BUG: Build fails when using data records and messaging steps in Spark pipelines


ewilkins-csi:
## Description
There is a build error that occurs during project generation when there are semantic data record models and Spark pipelines that use those records in messaging steps.

## Steps to Reproduce
Clear, specific, and detailed steps taken to enable reproduction of the bug for investigation.
1. Create a project
2. Add a dictionary metamodel, a record metamodel, and a pipeline metamodel with Spark and at least one step with inbound/outbound type of `messaging` and a `recordType` set to the record model
3. Generate out the project following manual actions

## Expected Behavior
The project can be incepted without error.

## Actual Behavior
A build exception is thrown during the compile phase of the `*-data-records` module.

## Definition of Done
 - The POMs for combined data records, and the POM for core data records when using split generation are updated to include the requisite Kafka dependency, regardless of the pipeline
   - This will result in the dependency included when it isn't needed, but the extra dependency is relatively cheap, and avoids user confusion if they update their messaging steps _later_ to include the record

## Additional Context
Error:
```sh
14:49:13 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.13.0:compile (default-compile) on project semdata-messaging-bug-data-records: Compilation failure: Compilation failure: 
14:49:13 [ERROR] /Users/ewilkins/dev/aiops/init-tests/oneoff/semdata-messaging-bug/semdata-messaging-bug-shared/semdata-messaging-bug-data-records/src/generated/java/org/test/SimpleRecordDeserializerBase.java:[16,45] package org.apache.kafka.common.serialization does not exist
14:49:13 [ERROR] /Users/ewilkins/dev/aiops/init-tests/oneoff/semdata-messaging-bug/semdata-messaging-bug-shared/semdata-messaging-bug-data-records/src/generated/java/org/test/SimpleRecordDeserializerBase.java:[26,54] cannot find symbol
14:49:13 [ERROR]   symbol: class Deserializer
14:49:13 [ERROR] /Users/ewilkins/dev/aiops/init-tests/oneoff/semdata-messaging-bug/semdata-messaging-bug-shared/semdata-messaging-bug-data-records/src/generated/java/org/test/SimpleRecordSerializerBase.java:[16,45] package org.apache.kafka.common.serialization does not exist
[...]

PR #459, opened 2024-11-07
[#457] fix record messaging

ewilkins-csi:


foundation/foundation-archetype/test-project-archetype.sh
ewilkins-csi, line 1: I: majority of changes are tabs -> spaces

foundation/foundation-archetype/test-project-archetype.sh
```diff
@@ -99,72 +99,80 @@ echo "{
...
+        \"package\": \"${package}\",
+        \"type\": {
+          \"name\": \"data-flow\",
+          \"implementation\": \"data-delivery-pyspark\"
+        },
+        \"steps\": [
+          {
+            \"name\": \"Ingest\",
+            \"type\": \"synchronous\",
+            \"inbound\": {
```
ewilkins-csi, line 122: I: This `inbound` is new. This is the condition that causes the bug.

foundation/foundation-mda/src/main/resources/templates/data-delivery-data-records/data-delivery-combined-data-records-java.pom.xml.vm
```diff
@@ -31,5 +31,10 @@
             <groupId>com.boozallen.aissemble</groupId>
             <artifactId>extensions-data-delivery-spark</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.smallrye.reactive</groupId>
+            <artifactId>smallrye-reactive-messaging-kafka</artifactId>
```
ewilkins-csi, line 36: I: Key fix. Otherwise, the serde classes will be generated, but not have access to the Kafka classes that they use.  We could _conditionally_ add this dependency that's tied to the same condition that triggers the serde generation.  However:
 - it's not unlikely that the condition will be added after initial generation (requires updating a pipeline messaging step to use the record for inbound/outbound)
 - the fix isn't obvious if you do add record messaging
 - a single extra dependency is relatively cheap
 - it's very easy to detect and remove an unused dependency

extensions/extensions-encryption/aissemble-extensions-encryption-vault-python/tests/features/environment.py
```diff
@@ -52,6 +52,7 @@ def start_container(context, docker_image, feature) -> int:
         raise Exception("Vault failed to start")
     return extport
 
+
```
ewilkins-csi, line 55: I: This was automatically added by the formatter plugin

csun-cpointe (approved):

foundation/foundation-archetype/test-project-archetype.sh
```diff
@@ -99,72 +99,99 @@ echo "{
...
+        \"package\": \"${package}\",
+        \"type\": {
+          \"name\": \"data-flow\",
+          \"implementation\": \"data-delivery-spark\"
+        },
+        \"steps\": [
+          {
+            \"name\": \"Ingest\",
+            \"type\": \"synchronous\",
+            \"inbound\": {
```
ewilkins-csi, line 112: I: This step is new. `inbound` + `recordType` is the condition that causes the bug.

cwoods-cpointe (approved):

Issue #467, opened 2024-11-14
Feature: Document the setup of custom GitHub action runners that can execute the aiSSEMBLE build.


ewilkins-csi:
## Description
Our build has been failing recently due to disk pressure on the cluster that hosts our GHA runners.  During the process of expanding the disk size for the cluster, the [ARC](https://github.com/actions/actions-runner-controller) deployment responsible for provisioning runners on-demand for aiSSEMBLE stopped working.  It has been restored, but we should document the setup process and any caveats so that we can easily redeploy ARC for aiSSEMBLE if/when we need to.

## DOD
- [x] There is information in the `devops` directory of the repository that outlines ARC setup

## Test Strategy/Script
- Verify the README exists and describes the setup process
- Verify the latest build is passing

## References/Additional Context
We could create a chart for deploying ARC in a compatible way, but the focus of this ticket is largely to get the CI build working again so we can continue with normal operations.

ewilkins-csi:
**DoD/OTS** with @csun-cpointe

csun-cpointe:
Test passed!
**ARC ReadMe**
![Screenshot 2024-11-14 at 10 56 50 AM](https://github.com/user-attachments/assets/d30f6663-74c0-46f3-8fa9-815aac46d814)

**Latest Build Success**
![Screenshot 2024-11-14 at 10 57 28 AM](https://github.com/user-attachments/assets/c04d323c-473a-49a7-86c9-0ac217359653)

PR #468, opened 2024-11-14
[#467] document self-hosted runner setup

ewilkins-csi:
The documentation for setting up self-hosted runners that can handle the aissemble build via the Actions Runner Controller project has been added to `devops`. In addition, the `helm install --dry-run` IT has been removed and the build action has been adjusted to skip cache-restore actions if the trigger was a schedule.  The dry run was removed largely because it just requires too many permissions to function and isn't adding much more than our strict linting operation and unit tests. The cache restore skip was added as an alternative to the manual cache deletion logic we were doing in the build before. General cache cleanup is handled automatically by S3.

### DIFF
--- a/devops/ARC_README.md
+++ b/devops/ARC_README.md
@@ -1,0 +1,76 @@
+# Actions Runner Controller
+
+The [Actions Runner Controller](https://github.com/actions/actions-runner-controller) (ARC) is a project to ease the
+creation of self-hosted GitHub action runners.  The following content is largely in line with the information provided
+in the GitHub documentation but is preserved here as an audit trail in case the online information changes, since ARC is
+still pre-1.0.
+
+## Standing up ARC from scratch
+
+### Creating the Controller
+
+The controller only needs to be created once for the whole cluster.  The controller MUST be created before the runner
+sets and the versions must match between the two.  The controller is responsible for detecting new ARC CRD manifests and
+provisioning the appropriate resources for the runner-set, including the listener which recieves action run requests
+from GitHub.
+
+```sh
+helm install arc-controller oci://ghcr.io/actions/actions-runner-controller-charts/gha-runner-scale-set-controller \
+             --namespace gh-actions-controller \
+             --create-namespace \
+             -f controller-values.yaml
+```
+
+### Creating the Runner Scale Set
+
+Previously, we needed special permissions on the runner pod to execute `helm install --dry-run` as an integration test
+for our baseline charts. However, this requires pretty expansive permissions (cluster level + secrets retrieval). The
+dry-run IT wasn't really adding much over the simple strict linting approach so it's been dropped. The custom service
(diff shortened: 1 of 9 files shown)


devops/ARC_README.md
```diff
@@ -0,0 +1,67 @@
+# Standing up ARC from scratch
```
carter-cundiff, line 1: Q: What does ARC stand for? Might be helpful to mention it once.

devops/ARC_README.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Creating the Controller
+The controller only needs to be created once for the whole cluster.  The controller MUST be created before the runner
```
ewilkins-csi, line 4: A: This has duplicate content with the upgrade section. Needs to be cleaned up.

devops/runnerset-values.yaml
```diff
@@ -0,0 +1,7 @@
+containerMode:
+  type: "dind"
+githubConfigUrl: "https://github.com/boozallen/aissemble"
+maxRunners: 3
```
cwoods-cpointe, line 4: q: Have we tested with multiple simultaneous job executions?

carter-cundiff (approved):

devops/ARC_README.md
ewilkins-csi: [Actions Runner Controller](https://github.com/actions/actions-runner-controller). Added the info and link as an intro paragraph to the document.

devops/runnerset-values.yaml
ewilkins-csi: Yeah.  There _can_ be a conflict due to cache lockdown, but in general it's fine. This allows us to, say, do a release while running the build against `dev` and re-publishing the GH pages docs.

ewilkins-csi: And actually the default value for this is 5 and is the value we were using before. I actually decreased it here.

devops/ARC_README.md
```diff
@@ -0,0 +1,76 @@
...
+Previously, we needed special permissions on the runner pod to execute `helm install --dry-run` as an integration test
+for our baseline charts. However, this requires pretty expansive permissions (cluster level + secrets retrieval). The
+dry-run IT wasn't really adding much over the simple strict linting approach so it's been dropped. The custom service
+account setup has been left in place simply to demonstrate how this _would_ be done if we need permissions for some
+other use case in the future.
+
+>[!NOTE]
+>The namespace is currently hard-coded in the YAML file, so if you intend to deploy to a different namespace when
+>installing the runner-set, you must update the YAML file as well.  This is required because a RoleBinding object in
+>Kubernetes must specify the ServiceAccount namespace explicitly.
```
cpointe-ibllanos, line 35: Thank you for finding this nuance -- was wondering why my past attempts to parameterize the namespace had run into issues.

cpointe-ibllanos (approved):

Issue #422, opened 2024-10-18
BUG: Pyspark pipeline generates duplicated data records under pipeline with combine records


csun-cpointe:
## Description
When creating the pyspark pipeline project with combine records, the data records are generated under both `-pipelines/py-spark-pipeline` and the `-share` directories.

The same issue is also seen when generate the pyspark project with split data records.

## Steps to Reproduce
1. Create a new project on 1.10.0-SNAPSHOT.
    ```sh
    mvn archetype:generate '-DarchetypeGroupId=com.boozallen.aissemble' \
                           '-DarchetypeArtifactId=foundation-archetype' \
                           '-DarchetypeVersion=1.10.0-SNAPSHOT' \
                           '-DgroupId=org.test' \
                           '-Dpackage=org.test' \
                           '-DprojectGitUrl=test.org/test.git' \
                           '-DprojectName=Test combine records' \
                           '-DartifactId=test-combine-records' \
    && cd test-combine-records
    ```
2. Set your Java version to **_17_** if it is not currently
3. Unzip the [resources.zip](https://github.com/user-attachments/files/17437831/resources.zip) and replace the resources folder at the `-pipeline-models/src/main/` directory
4. Fully generate the project by running `mvn clean install` and following manual actions
5. Unzip the [krausening.zip](https://github.com/user-attachments/files/17437772/krausening.zip) and replace the krausening folder at the `-docker/test-combine-record-spark-worker-docker/src/main/resources` directory
7. Build the project without the cache and follow the last manual action.
    ```sh
    mvn clean install -Dmaven.build.cache.skipCache
    ```

## Expected Behavior
There should be no data records generated under the `-pipelines/py_spark_pipeline` directory

## Actual Behavior
The data records are generated under both `-pipelines/py_spark_pipeline` and `-shared` directories
Note: the same issues not be seen in the spark pipeline

## Definition of Done

- [x] Fix MDA to make sure data records should only generated under '-shared' directories.
- [x] make sure aissemble build is successful.
- [x] Make sure downstream project with pyspark pipeline can be stood up without any error.

## Testing Instructions

- [x] fetch latest dev branch of aissemble
- [x] do mvn clean install
- [x] Create a new project on 1.10.0-SNAPSHOT.
    ```
    mvn archetype:generate '-DarchetypeGroupId=com.boozallen.aissemble' \
                           '-DarchetypeArtifactId=foundation-archetype' \
[...]

PR #472, opened 2024-11-18
#422 BUG: Pyspark pipeline generates duplicated data records under pipeline with combine records

jaebchoi:
Issue 422 https://github.com/boozallen/aissemble/issues/422

### DIFF
--- a/docs/modules/ROOT/pages/semantic-data.adoc
+++ b/docs/modules/ROOT/pages/semantic-data.adoc
@@ -220,7 +220,8 @@ Additionally, update any `import` statements in your pipeline code to reflect th
 
 In some very targeted cases where the data model will never need to be re-used and there is only one pipeline that
 processes the data, the semantic data classes can be directly generated into the pipeline module. In general, this is
-the legacy behavior of aiSSEMBLE and should only be undertaken with due consideration. In order to do this, update the
+the legacy behavior of aiSSEMBLE and should only be undertaken with due consideration. In order to do this, first delete
+generated spark pipelines if necessary and update the
 POM file of the pipeline to replace the `<profile>data-delivery-spark-pipeline</profile>` with
 `<profile>data-delivery-spark</profile>` in the following location:
 
--- a/foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/generator/util/MavenUtil.java
+++ b/foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/generator/util/MavenUtil.java
@@ -214,6 +214,44 @@ public class MavenUtil {
             return packaging;
         }
     }
+
+    /**
+     * Reads the given POM file to determine what profile is used for pyspark data delivery pipeline
+     *
+     * @param context GenerationContext
+     * @return the profile for pyspark data delivery pipeline
+     */
+    public static String getPySparkDataDeliveryProfileName(GenerationContext context) {
+
+        File pomFile = new File(context.getProjectDirectory() + File.separator + "pom.xml");
+
+        try {
+            DocumentBuilderFactory dbFactory = DocumentBuilderFactory.newInstance();
+            DocumentBuilder dBuilder = dbFactory.newDocumentBuilder();
+
+            Document doc = dBuilder.parse(pomFile);
+            doc.getDocumentElement().normalize();
+            //Get all fields with tag <profile>
+            NodeList nodeList = doc.getElementsByTagName("profile");
+            for (int i = 0; i < nodeList.getLength(); i++){
+                Node node = nodeList.item(i);
+
+                if (node.getNodeType() == Node.ELEMENT_NODE){
+
+                    Element element = (Element) node;
+                    String field = element.getTextContent();
+                    if(field.equals("data-delivery-pyspark") || field.equals("data-delivery-pyspark-pipeline") )
+                    {
+                        return field;
+                    }
+                }
+            }
+        } catch (Exception e){
+            logger.error("Unable to find profileName", e);
+            throw new NoSuchElementException("No profile name identified within " + pomFile.getPath());
+        }
+        return null;
+    }
 }
 
 
--- a/foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/python/PythonStep.java
+++ b/foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/python/PythonStep.java
@@ -11,6 +11,7 @@ package com.boozallen.aiops.mda.metamodel.element.python;
  */
 
 import com.boozallen.aiops.mda.generator.util.PipelineUtils;
+import com.boozallen.aiops.mda.generator.util.PythonGeneratorUtils;
 import com.boozallen.aiops.mda.metamodel.element.BaseStepDecorator;
 import com.boozallen.aiops.mda.metamodel.element.Persist;
 import com.boozallen.aiops.mda.metamodel.element.Step;
@@ -28,6 +29,8 @@ import java.util.TreeSet;
  */
 public class PythonStep extends BaseStepDecorator {
 
+    private String profileName;
+    private String rootArtifactId;
     protected static final String DATAFRAME_TYPE = "pysparkDataFrame";
     private static final String NONE = "None";
     private static final String STRING = "str";
@@ -98,7 +101,7 @@ public class PythonStep extends BaseStepDecorator {
     public Set<String> getBaseImports() {
         getBaseSignature();
         addPersistImports();
-        return getImportsWithRelativePaths(false);
+        return getImports(false);
     }
 
     /**
@@ -110,7 +113,7 @@ public class PythonStep extends BaseStepDecorator {
     public Set<String> getImplImports() {
         getConcreteSignature();
 
-        return getImportsWithRelativePaths(true);
+        return getImports(true);
     }
 
     /**
@@ -120,6 +123,8 @@ public class PythonStep extends BaseStepDecorator {
      * By using relative (instead of absolute) imports to generated record/dictionary classes,
      * we simplify generation and mitigate the need to generate the encapsulating package
      * path into every import statement in order to form valid absolute imports.
+     * This Relative imports only applies for Old Monolith Generation.
+     * For New case of Semantic Data we simply fetch from shared directory.
      * <p> For example, the import statement to a metamodel defined record becomes
      * {@code from ..record.custom_record import CustomRecord} instead of
      * {@code from the_package_name.record.custom_record import CustomRecord}.
@@ -130,18 +135,25 @@ public class PythonStep extends BaseStepDecorator {
      * @return {@link #imports} that have been appropriately modified to include the correct
      * relative import paths for any references to metamodel-define records or dictionary types.
      */
-    private Set<String> getImportsWithRelativePaths(boolean isImplModule) {
-        Set<String> importsWithRelativePaths = new TreeSet<>();
+    private Set<String> getImports(boolean isImplModule) {
+        Set<String> importsSet = new TreeSet<>();
         for (String moduleImport : imports) {
             if (moduleImport.startsWith("from record.")
                     || moduleImport.startsWith("from dictionary.")) {
-                importsWithRelativePaths.add(moduleImport.replace("from ",
-                        isImplModule ? "from .." : "from ..."));
+                if(profileName.equals("data-delivery-pyspark"))
+                {
+                    importsSet.add(moduleImport.replace("from ",
+                            isImplModule ? "from .." : "from ..."));
+                }else if(profileName.equals("data-delivery-pyspark-pipeline")){
+                    String rootArtifact = rootArtifactId.replace("-", "_");
+                    importsSet.add(moduleImport.replace("from ",
+                            "from " + rootArtifact + "_data_records."));
+                }
             } else {
-                importsWithRelativePaths.add(moduleImport);
+                importsSet.add(moduleImport);
             }
         }
-        return importsWithRelativePaths;
+        return importsSet;
     }
 
     /**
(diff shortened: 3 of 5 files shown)


foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/generator/util/PythonGeneratorUtils.java
```diff
@@ -36,6 +37,7 @@ public static VelocityContext populateCommonPythonContext(VelocityContext veloci
         if (velocityContext == null) {
             velocityContext = new VelocityContext();
         }
+        rootArtifactId = generationContext.getRootArtifactId();
```
jaebchoi, line 40: I: This is done so that we can retrieve rootArtifactID in the metamodel package.

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/python/PythonStep.java
```diff
@@ -135,8 +136,9 @@ private Set<String> getImportsWithRelativePaths(boolean isImplModule) {
         for (String moduleImport : imports) {
             if (moduleImport.startsWith("from record.")
                     || moduleImport.startsWith("from dictionary.")) {
+                String rootArtifact = PythonGeneratorUtils.rootArtifactId.replace("-", "_");
                 importsWithRelativePaths.add(moduleImport.replace("from ",
-                        isImplModule ? "from .." : "from ..."));
+                        "from " + rootArtifact + "_data_records."));
```
jaebchoi, line 141: I: few of python imports had relative imports, because they did generated data records under pyspark directory.

Now that we deleted duplicated data record in pyspark, needed to point those into correct package.

foundation/foundation-mda/src/main/resources/templates/data-delivery-pyspark/pipeline.pom.xml.vm
```diff
@@ -24,7 +24,7 @@
                 <artifactId>fermenter-mda</artifactId>
                 <configuration>
                     <basePackage>${moduleArtifactIdPythonCase}</basePackage>
-                    <profile>data-delivery-pyspark</profile>
+                    <profile>data-delivery-pyspark-pipeline</profile>
                     <propertyVariables>
```
meliz19, line 28: What prompted this change?

jaebchoi: This is actually core fix to remove duplication. According to profile.json, data-delivery-pyspark profile had unnecessary profile references which led to duplication. Just importing data-delivery-pyspark-pipeline can be sufficient.

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/generator/util/PythonGeneratorUtils.java
```diff
@@ -20,6 +20,7 @@
  * across {@link org.technologybrewery.fermenter.mda.generator.Generator} hierarchies.
  */
 public final class PythonGeneratorUtils {
+    public static String rootArtifactId;
```
ewilkins-csi, line 23: S: This shared state seems like a bad idea, and doesn't seem cohesive with the class goals at all.  I think we're probably missing some parameter passing somewhere.

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/python/PythonStep.java
```diff
@@ -135,8 +136,9 @@ private Set<String> getImportsWithRelativePaths(boolean isImplModule) {
         for (String moduleImport : imports) {
             if (moduleImport.startsWith("from record.")
                     || moduleImport.startsWith("from dictionary.")) {
+                String rootArtifact = PythonGeneratorUtils.rootArtifactId.replace("-", "_");
                 importsWithRelativePaths.add(moduleImport.replace("from ",
```
ewilkins-csi, line 140: A: The name of this method/variable seems misleading now. These aren't relative imports.

jaebchoi: yup now it isrelative imports anymore, Ill update to something like getDataRecordsAndDictImports.

foundation/foundation-mda/src/main/resources/templates/data-delivery-pyspark/pipeline.pom.xml.vm
```diff
@@ -24,7 +24,7 @@
                 <artifactId>fermenter-mda</artifactId>
                 <configuration>
                     <basePackage>${moduleArtifactIdPythonCase}</basePackage>
-                    <profile>data-delivery-pyspark</profile>
+                    <profile>data-delivery-pyspark-pipeline</profile>
```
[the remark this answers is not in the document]
jaebchoi: Ill test to make sure, didn't realize this use case!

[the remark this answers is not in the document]
jaebchoi: Okay so old monolith generation did failed with my new impl, so I kept the old implementation to support both cases based on profile name of pipleline

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/generator/util/PythonGeneratorUtils.java
jaebchoi: updated to pass in artifactID instead of setting public static variables.

foundation/foundation-mda/src/main/resources/templates/data-delivery-pyspark/pipeline.pom.xml.vm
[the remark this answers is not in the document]
jaebchoi: Verified old monolith gen works

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/generator/util/MavenUtil.java
```diff
@@ -214,6 +214,45 @@ public String getPackaging() {
             return packaging;
         }
     }
+
```
jaebchoi, line 217: I: This is needed so that I can retrieve profile Name from maven pom so that I can handle Old Monolith Generation and new shared data records use cases.

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/python/PythonStep.java
```diff
@@ -130,18 +133,25 @@ public Set<String> getImplImports() {
      * @return {@link #imports} that have been appropriately modified to include the correct
      * relative import paths for any references to metamodel-define records or dictionary types.
      */
-    private Set<String> getImportsWithRelativePaths(boolean isImplModule) {
-        Set<String> importsWithRelativePaths = new TreeSet<>();
+    private Set<String> getImports(boolean isImplModule) {
```
jaebchoi, line 136: I: this is to retrieve all imports including records and dict so it would make sense to name it as just getImports.

docs/modules/ROOT/pages/semantic-data.adoc
```diff
@@ -220,7 +220,8 @@ Additionally, update any `import` statements in your pipeline code to reflect th
 
 In some very targeted cases where the data model will never need to be re-used and there is only one pipeline that
 processes the data, the semantic data classes can be directly generated into the pipeline module. In general, this is
-the legacy behavior of aiSSEMBLE and should only be undertaken with due consideration. In order to do this, update the
+the legacy behavior of aiSSEMBLE and should only be undertaken with due consideration. In order to do this, first delete
+generated spark pipelines if necessary and update the
```
jaebchoi, line 224: I: I added this because mvn clean build doesn't overwrite already generated files.

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/generator/util/MavenUtil.java
```diff
@@ -214,6 +214,45 @@ public String getPackaging() {
...
+            for (int i = 0; i < nodeList.getLength(); i++){
+                Node node = nodeList.item(i);
+
+                if (node.getNodeType() == Node.ELEMENT_NODE){
+
+                    Element element = (Element) node;
+                    String field = element.getTextContent();
+                    if(field.startsWith("data-delivery"))
+                    {
+                        return field;
```
csun-cpointe, line 246: Q: it seems we always return the first profile name starting with `data-delivery`, is it possible that we have multiple profile name starting with `data-delivery` and can we be more specific what profile name we are looking for?

jaebchoi: yup I missed that, this could cause issues when there're multiple profile that starts data-delivery by looking at the profiles.json
What I am seeking for is whether pom has data-delivery-pyspark or data-delivery-pyspark.
Upon investigating profiles.json, we have only two profiles that starts with data-delivery-pyspark so Ill update to start with data-delivery-pyspark.

In some rare edge case we add data-delivery-pyspark and data-delivery-pyspark-pipeline in a pom file, that's okay. There will be many duplication in code but we have cases that supports both, so just getting first profile starting data-delivery-pyspark is good.

csun-cpointe (approved):

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/generator/util/MavenUtil.java
```diff
@@ -214,6 +214,44 @@ public String getPackaging() {
...
+                    String field = element.getTextContent();
+                    if(field.equals("data-delivery-pyspark") || field.equals("data-delivery-pyspark-pipeline") )
+                    {
+                        return field;
+                    }
+                }
+            }
+        } catch (Exception e){
+            logger.error("Unable to find profileName", e);
+            throw new NoSuchElementException("No profile name identified within " + pomFile.getPath());
```
ewilkins-csi, line 251: A: We should not both log an exception and throw. That results in double-logging and cluttered log files that are hard to parse. We should just re-throw the exception in a wrapper that gives a specific message, and let higher code catch and log as appropriate. E.g.: `throw new FermenterException("Failed to locate pipeline generation Fermenter profile in " + pomFile.getPath(), e);`

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/generator/util/MavenUtil.java
```diff
@@ -214,6 +214,44 @@ public String getPackaging() {
...
+        }
+        return null;
```
ewilkins-csi, line 253: A: I don't think I agree with returning `null` here.  We cannot perform generation properly if we don't know whether data records are being included in the same source. I'd even prefer this method maybe return a boolean and be called `isRecordGenerationInPipelineModule`. Then return false if you don't find `data-delivery-pyspark` exactly. This would keep the code more flexible in case we add more/different pipeline-specific profiles in the future.

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/python/PythonStep.java
```diff
@@ -130,18 +135,25 @@ public Set<String> getImplImports() {
...
+        Set<String> importsSet = new TreeSet<>();
         for (String moduleImport : imports) {
             if (moduleImport.startsWith("from record.")
                     || moduleImport.startsWith("from dictionary.")) {
-                importsWithRelativePaths.add(moduleImport.replace("from ",
-                        isImplModule ? "from .." : "from ..."));
+                if(profileName.equals("data-delivery-pyspark"))
```
ewilkins-csi, line 143: A: Given we allow `profileName` to be `null` this is susceptible to a hard to trace NPE. In general, a good practice is to put the constant on the LHS of the equals just in case. E.g. `"data-delivery-pyspark".equals(profileName)`

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/python/PythonStep.java
```diff
@@ -130,18 +135,25 @@ public Set<String> getImplImports() {
...
+                {
+                    importsSet.add(moduleImport.replace("from ",
+                            isImplModule ? "from .." : "from ..."));
+                }else if(profileName.equals("data-delivery-pyspark-pipeline")){
+                    String rootArtifact = rootArtifactId.replace("-", "_");
+                    importsSet.add(moduleImport.replace("from ",
+                            "from " + rootArtifact + "_data_records."));
+                }
```
ewilkins-csi, line 151: A: else? We could consider throwing an exception in the final else block, again just to fail faster and help track down bugs by signaling they're true root faster.

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/python/PythonStep.java
```diff
@@ -130,18 +135,25 @@ public Set<String> getImplImports() {
...
             if (moduleImport.startsWith("from record.")
                     || moduleImport.startsWith("from dictionary.")) {
-                importsWithRelativePaths.add(moduleImport.replace("from ",
-                        isImplModule ? "from .." : "from ..."));
+                if(profileName.equals("data-delivery-pyspark"))
+                {
+                    importsSet.add(moduleImport.replace("from ",
+                            isImplModule ? "from .." : "from ..."));
+                }else if(profileName.equals("data-delivery-pyspark-pipeline")){
+                    String rootArtifact = rootArtifactId.replace("-", "_");
```
ewilkins-csi, line 148: A: I think there is a util method that will convert an artifact ID to a python package name.

Issue #469, opened 2024-11-14
Feature: Release 1.10.0


ewilkins-csi:
## Description
A clear and concise description of the new feature. It can include what the problem is and what you want to happen, and is preferably focused on the business context or how this impacts downstream users.

## DOD
Acceptance criteria required to realize the requested feature
- [ ] Using checkboxes helps to communicate progress

## Test Strategy/Script
How will this feature be verified?

## References/Additional Context
Add any other context, links, or screenshots about the feature request here.

carter-cundiff:
Patch fix for record bug: [record-fix.patch](https://github.com/user-attachments/files/17758549/record-fix.patch)

ewilkins-csi:
**OTS** with @csun-cpointe and @carter-cundiff

csun-cpointe:
Final test passed!
![Screenshot 2024-11-20 at 11 50 33 AM](https://github.com/user-attachments/assets/cdc0137f-5814-485e-b0f6-59f123d9c74b)

PR #473, opened 2024-11-18
[#469] fix breaking bugs before release

ewilkins-csi:
The following are a series of bugs that caused breakages in downstream projects. The fixes are relatively minor and straight-forward.

- Fix record templating when column name is defined
- Update Quarkus BOM to exclude Scala 2.13 dependencies
- Move sealed-secret from `foundation-mda` to `foundation-archetype`


bom/aissemble-quarkus-bom/pom.xml
```diff
@@ -28,8 +28,28 @@
                         <groupId>io.smallrye.reactive</groupId>
                         <artifactId>*</artifactId>
                     </exclusion>
+                    <exclusion>
+                        <groupId>org.scala-lang</groupId>
+                        <artifactId>*</artifactId>
+                    </exclusion>
                 </exclusions>
             </dependency>
+            <!-- Downgrade Scala 2.13 (from Quarkus) to Scala 2.12 -->
```
ewilkins-csi, line 37: I: aiSSEMBLE uses Scala 2.12 but Quarkus defaults to Scala 2.13

foundation/foundation-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
ewilkins-csi, line 1: I: The archetype definition had a lot of overlaps/redundancies that caused warnings.  For the Maven Wrapper configs, we were copying the `.mvn` directory twice in two different filesets, and the exclusions weren't necessary as they're only needed to remove specific files from a more general includes.  Finally, for the modules with `src` directories, including `**` also indicates the `.` directory itself causing a warning. The correct definition is `*/**`.

The archetype no longer generates warnings with these changes, but the resulting files are identical.

foundation/foundation-archetype/src/main/resources/archetype-resources/__rootArtifactId__-deploy/src/main/resources/templates/remote-auth-secret.yaml
```diff
@@ -0,0 +1,10 @@
+# Convert this to a SealedSecret using kubeseal and replace it here
```
ewilkins-csi, line 1: I: Moved this to the archetype so that projects wouldn't have to deal with it constantly regenerating if they a) didn't need it or b) renamed it to something more explicit to their project.

foundation/foundation-mda/src/main/resources/templates/data-delivery-data-records/spark.schema.base.java.vm
```diff
@@ -90,30 +92,29 @@ public abstract class ${record.capitalizedName}SchemaBase extends SparkSchema {
     public Dataset<Row> validateDataFrame(Dataset<Row> data) {
         Dataset<Row> dataWithValidations = data
             #foreach ($field in $record.fields)
-                #set ( $columnName = "#if($field.column)$field.column#{else}$field.upperSnakecaseName#end" )
                 #if (${field.isRequired()})
-                    .withColumn("${columnName}_IS_NOT_NULL", col(${columnName}_COLUMN).isNotNull())
```
ewilkins-csi, line 95: I: This is the original bug. When `field.column` is set to some name, the constant is referenced here as `col(my_col_name_COLUMN)` even though the constant is still `MY_FIELD_COLUMN`. To help ensure this mistake doesn't happen again, we just create a mapping of field to constant at the top of the file and use that mapping any time we want to reference the constant.

test/test-mda-models/test-data-delivery-spark-model/src/main/resources/records/RecordWithValidation.json
```diff
@@ -8,56 +8,64 @@
 			"type": {
 				"name": "stringWithValidation",
 				"package": "com.boozallen.aiops.mda.pattern.dictionary"
-			}
+			},
+			"column": "str_v8n"
```
ewilkins-csi, line 12: I: New test for column name bug

test/test-mda-models/aissemble-test-data-delivery-pyspark-model/src/aissemble_test_data_delivery_pyspark_model/resources/records/BinaryRecord.json
ewilkins-csi, line 1: I: A lot of the PySpark model deletions/renames were just to more closely align the test models between Spark and Pyspark. There are times when, due to the nature of the two implementations, the models must deviate. but largely we had distinctions without a difference.

carter-cundiff (approved):

cucumber-report-aggregator/pom.xml
```diff
@@ -49,6 +49,7 @@
             <plugin>
                 <groupId>net.masterthought</groupId>
                 <artifactId>maven-cucumber-reporting</artifactId>
+                <version>${version.cucumber.reporting.plugin}</version>
```
ewilkins-csi, line 52: I: Fixes warning about missing version definition

PR #476, opened 2024-11-19
[#469] fix encryption ITs for changed MDA model

ewilkins-csi:
Part of the encryption IT scenarios were checking that the right number of fields were retrieved by the `get_field_list` method.  I'm not sure that this makes sense as an encryption test.  Either the encryption method (which uses `get_field_list`) encrypts fields correctly, or it doesn't. I don't see a ton of value in testing that method specifically over the actual encryption tests which indirectly test `get_field_list`.

This changeset also splits off the vault scenarios into a separate feature, as those are the only real "IT" scenarios.  We can't just tag the scenarios directly because the feature-level tag is used to determine whether we should stand up the Vault image, and we can't base this logic at the scenario level because that would tear down and deploy Vault between each scenario, which is wasteful since we aren't changing the state of Vault between scenarios.

### DIFF
--- a/test/test-mda-models/aissemble-test-data-delivery-pyspark-model/tests/features/pyspark_data_encryption.feature
+++ b/test/test-mda-models/aissemble-test-data-delivery-pyspark-model/tests/features/pyspark_data_encryption.feature
@@ -1,22 +1,7 @@
-@pyspark_data_encryption @integration
+@pyspark_data_encryption
 Feature: Data encryption
     # Generated sample BDD specification/feature file - PLEASE ***DO*** MODIFY.
     # Originally generated from templates/behave.feature.vm.
-
-  Scenario: Get fields from native inbound collection with inbound record type (set([CustomData]))
-  Given a pipeline with native inbound collection and inbound record type
-  When field names are retrieved for the inbound record
-  Then a list of field names can be retrieved
-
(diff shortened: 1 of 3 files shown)


test/test-mda-models/aissemble-test-data-delivery-pyspark-model/tests/features/pyspark_data_encryption.feature
```diff
@@ -1,23 +1,8 @@
-@pyspark_data_encryption @integration
+@pyspark_data_encryption
 Feature: Data encryption
     # Generated sample BDD specification/feature file - PLEASE ***DO*** MODIFY.
     # Originally generated from templates/behave.feature.vm.
 
-  Scenario: Get fields from native inbound collection with inbound record type (set([CustomData]))
```
ewilkins-csi, line 6: I: Removed scenarios that I felt were redundant and overly fine-grained/brittle.  If there is a strong feeling that these are useful we can discuss keeping them and just fixing the assertion to account for the model changes.

cwoods-cpointe (approved):
Code changes look good but I cant comment on if the tests are redundant or not. Your reasoning is sound.

chang-annie (approved):
I think it makes sense to pull the get fields test out of encryption 👍

meliz19 (approved):

carter-cundiff (approved):

Issue #474, opened 2024-11-18
Feature: Publish aggregated Cucumber report from cucumber-report-aggregator


habibimoiz:
## Description
As part of [#447](https://github.com/boozallen/aissemble/issues/447) was to generate an aggregated report of the Cucumber tests for the entire aiSSEMBLE report. The next objective is to take the generated report and publish that to a public site such that the report can be accessible outside of aiSSEMBLE. Currently, the aggregated report is generated locally in the /target directory under /cucumber-reports/cucumber-html-reports/overview-features.html. Theoretically, this local HTML directory can be published.

### Approach 1: GitHub Pages (peaceiris)
Upon doing research, the [gh-actions-pages](https://github.com/peaceiris/actions-gh-pages?tab=readme-ov-file#getting-started) plugin from [peaceiris](https://github.com/peaceiris) appears to be a promising approach. This is a GitHub Action to deploy your static files to GitHub Pages. This deploy action can be combined simply and freely with [Static Site Generators](https://jamstack.org/generators/). (Hugo, MkDocs, Gatsby, mdBook, Next, Nuxt, and so on).

We can potentially use the [Deploy to Subdirectory destination_dir](https://github.com/peaceiris/actions-gh-pages?tab=readme-ov-file#%EF%B8%8F-deploy-to-subdirectory-destination_dir) option from their plugin to get the aggregated Cucumber report deployed to GitHub Pages.

### Approach 2: Explore the option on Publishing to Cucumber.io
Explore the ability to see if Cucumber allows an already generated report to be pushed to Cucumber.io; contact Cucumber support to see the possibility. 

### Approach 3: GitHub Pages (Jameslves)
Upon doing research, using the [github-pages-deploy-action](https://github.com/JamesIves/github-pages-deploy-action) plugin from [Jameslves](https://github.com/JamesIves) appears to be an alternate approach to [gh-actions-pages](https://github.com/peaceiris/actions-gh-pages?tab=readme-ov-file#getting-started) plugin. The same concerns exist with the use of this plugin. The only thing that stands out with this plugin however is the parallel use of our Antora docs with this. 

### Approach 4: Spin up new GitHub branch (i.e. `gh-pages`) and upload the aggregated cucumber report there
Spin up a new branch called `gh-pages`. The aggregated Cucumber report will live in here. From there, we will configure GitHub pages to read from this branch. By doing it this way, we can view both the Antora docs (aiSSEMBLE) docs in one GitHub Pages environment without any conflict.  

[...]

PR #490, opened 2024-12-03
#474 - Adding workflow step to push cucumber report to GitHub Pages branch. 

habibimoiz:
#474 - Adding workflow step to push cucumber report to GitHub Pages branch.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -100,6 +100,12 @@ jobs:
       - name: Run Archetype Tests
         run: |
           ./mvnw -B clean install -Parchetype-test -pl :foundation-archetype
+      - name: Publish aggregated Cucumber report
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./cucumber-report-aggregator/target/cucumber-reports/cucumber-html-reports
+          destination_dir: test-reports
       - name: Save docker build cache
         id: save-docker-build
         uses: runs-on/cache/save@v4


cpointe-ibllanos (approved):
Glad the peaceiris solution worked out!

.github/workflows/build.yml
```diff
@@ -100,6 +100,12 @@ jobs:
       - name: Run Archetype Tests
         run: |
           ./mvnw -B clean install -Parchetype-test -pl :foundation-archetype
+      - name: Publish aggregated Cucumber report
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./docs
```
epwilkins: C: This looks like the wrong path. Shouldn't it be the target directory under the report aggregator project?

.github/workflows/build.yml
```diff
@@ -100,6 +100,12 @@ jobs:
...
+          destination_dir: cucumber-html-reports
```
epwilkins: A: I think this will end up in the URL, so I'd prefer something like `test-reports`.

.github/workflows/build.yml
habibimoiz: Apologies, just updated it to the correct path. I also did a test run in CI against the branch the `gh-pages` branch did get updated.

.github/workflows/build.yml
habibimoiz: Updated the `destination_dir`.

epwilkins (approved):

Issue #484, opened 2024-11-21
Feature: integrate Universal Config store in the hive metastore service


jaebchoi:
## Description
Given that aiSSEMBLE support Universal config store service, It is great to make use of this service for consumer in order to prevent duplication in configuration.
First step is to try with one service called Hive Metastore service.

## DOD

- [x] Enable overriding hive metastoreServiceConfig.baseProperties in metastoreServiceConfig.properties 
- [x] Replace Hive Username with Universal config store.
- [x] Generate hive properties file on appropriate location (I.e. -deploy/src/main/resources/configuration)
- [x] Prompt potential consumers that we shall encrypt password using krausening with the link.
- [x] Create Downstream project with hive meta service and verify correct config is fetched from universal config service.
- [x] Add unit test to make sure yaml file contains appropriate properties values set and labels.( Helpful doc in Reference) 
- [x] Add baton migration for this change.
- [x] BDD scenario for migrating projects from legacy way to new way. 

## Stretch Goal
- [ ] Add Archetype test to ensure helm ignore file is correctly added https://maven.apache.org/archetype/maven-archetype-plugin/integration-test-mojo.html

## Test Strategy/Script

- Create a downstream project with baseline version 1.11.0-SNAPSHOT
```
mvn archetype:generate -U -DarchetypeGroupId=com.boozallen.aissemble \
  -DarchetypeArtifactId=foundation-archetype \
  -DarchetypeVersion=1.11.0-SNAPSHOT \
  -DgroupId=com.test \
  -DartifactId=test-hive1 \
  -DprojectGitUrl=https://github.com/{github-username}/test-hive1 \
  -DprojectName=test-hive1 \
  && cd test-hive1 
```

- Add the [SparkPipeline.json](https://github.com/user-attachments/files/16781488/SparkPipeline.json) file to the test-hive1-pipeline-models/src/main/resources/pipelines directory
- Run mvn clean install until all the manual actions are complete
- Once the manual actions are complete, run `mvn clean install -Dmaven.build.cache.skipCache=true` once to get any remaining manual actions
- Modify the helm templates for Argocd deployment:
       - Mac User: In the `-deploy/src/main/resources/apps/configuration-store/values-dev.yaml file` update the `volumePathOnNode` to be `/<pathToProject>/test-hive1-deploy/src/main/resources/configurations`
[...]

PR #494, opened 2024-12-05
#484 Feature: integrate Universal Config store in the hive metastore service

jaebchoi:
https://github.com/boozallen/aissemble/issues/484

### DIFF
--- a/DRAFT_RELEASE_NOTES.md
+++ b/DRAFT_RELEASE_NOTES.md
@@ -28,11 +28,12 @@ The following steps will upgrade your project to `1.11`. These instructions cons
 ## Automatic Upgrades
 To reduce burden of upgrading aiSSEMBLE, the Baton project is used to automate the migration of some files to the new version.  These migrations run automatically when you build your project, and are included by default when you update the `build-parent` version in your root POM.  Below is a description of all of the Baton migrations that are included with this version of aiSSEMBLE.
 
-| Migration Name                                     | Description                                                                                                                                                                  |
-|----------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| upgrade-tiltfile-aissemble-version-migration       | Updates the aiSSEMBLE version within your project's Tiltfile                                                                                                                 |
-| upgrade-v2-chart-files-aissemble-version-migration | Updates the Helm chart dependencies within your project's deployment resources (`<YOUR_PROJECT>-deploy/src/main/resources/apps/`) to use the latest version of the aiSSEMBLE |
-| upgrade-v1-chart-files-aissemble-version-migration | Updates the docker image tags within your project's deployment resources (`<YOUR_PROJECT>-deploy/src/main/resources/apps/`) to use the latest version of the aiSSEMBLE       |
+| Migration Name                                       | Description                                                                                                                                                                 |
+|------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| upgrade-tiltfile-aissemble-version-migration         | Updates the aiSSEMBLE version within your project's Tiltfile                                                                                                                |
+| upgrade-v2-chart-files-aissemble-version-migration   | Updates the Helm chart dependencies within your project's deployment resources (`<YOUR_PROJECT>-deploy/src/main/resources/apps/`) to use the latest version of the aiSSEMBLE |
+| upgrade-v1-chart-files-aissemble-version-migration   | Updates the docker image tags within your project's deployment resources (`<YOUR_PROJECT>-deploy/src/main/resources/apps/`) to use the latest version of the aiSSEMBLE      |
+| spark-infrastructure-universal-config-yaml-migration | Removes the default hive username (if present) from hive-metastore-service values.yaml so that it can be set by the configuration store service                             |
 
 To deactivate any of these migrations, add the following configuration to the `baton-maven-plugin` within your root `pom.xml`:
 
--- a/extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/templates/configmap.yaml
+++ b/extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/templates/configmap.yaml
@@ -2,20 +2,34 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: metastore-service-config
+  labels:
+    aissemble-configuration-store: {{ .Values.configMap.metadata.labels.configStore}}
 data:
   # Add all the default properties from the local values.yaml to the ConfigMap
   # Then check if there are any downstream properties and add them as well
   metastore-site.xml: |
+    {{- $basePropDict := dict }}
+    {{- $propDict := dict }}
+    {{- range $baseProperty := .Values.configMap.metastoreServiceConfig.baseProperties }}
+      {{- $basePropDict := set $basePropDict $baseProperty.name $baseProperty.value }}
+    {{- end }}
+    {{- if .Values.configMap.metastoreServiceConfig.properties }}
+      {{- range $property := .Values.configMap.metastoreServiceConfig.properties }}
+        {{- $propDict := set $propDict $property.name $property.value }}
+      {{- end }}
+    {{- end }}
+    
     <configuration>
-    {{- range $property := .Values.configMap.metastoreServiceConfig.baseProperties }}
+    {{- range $baseProperty := .Values.configMap.metastoreServiceConfig.baseProperties }}
+      {{- if and (hasKey $basePropDict $baseProperty.name) (not (hasKey $propDict $baseProperty.name))}}
       <property>
-        <name>{{ $property.name }}</name>
-        <value>{{ $property.value }}</value>
-
-      {{- if $property.description }}
-        <description>{{ $property.description }}</description>
+        <name>{{ $baseProperty.name }}</name>
+        <value>{{ $baseProperty.value }}</value>
+      {{- if $baseProperty.description }}
+        <description>{{ $baseProperty.description }}</description>
       {{- end }}
       </property>
+      {{- end }}
     {{- end }}
 
     {{- if .Values.configMap.metastoreServiceConfig.properties }}
@@ -23,7 +37,6 @@ data:
       <property>
         <name>{{ $property.name }}</name>
         <value>{{ $property.value }}</value>
-
         {{- if $property.description }}
         <description>{{ $property.description }}</description>
         {{- end }}
--- a/extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/values.yaml
+++ b/extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/values.yaml
@@ -15,11 +15,16 @@ image:
   dockerRepo: "ghcr.io/"
 
 mysql:
+  commonLabels:
+    aissemble-configuration-store: enabled
   enabled: true
   fullnameOverride: "hive-metastore-db"
   auth:
     # The schematool for the metastore service will create the database for us.
     database: "metastore"
+    # Note: Changing these values requires removal of the `hive-metastore-db-0` PVC, or manual modification of the
+    # persisted database.
+    username: $getConfigValue(groupName=spark-infrastructure;propertyName=metastore.db.username)
 
 hive:
   dbType: "mysql"
@@ -58,6 +63,9 @@ service:
 
 # hive-metastore-service Config Map
 configMap:
+  metadata:
+    labels:
+      configStore: enabled
   metastoreServiceConfig:
     baseProperties:
       - name: metastore.thrift.uris
@@ -72,4 +80,7 @@ configMap:
       - name: javax.jdo.option.ConnectionURL
         value: jdbc:mysql://hive-metastore-db:3306/metastore?createDatabaseIfNotExist=true&amp;allowPublicKeyRetrieval=true&amp;useSSL=false
         description: JDBC connect string for a JDBC metastore
+      - name: javax.jdo.option.ConnectionUserName
+        value: $getConfigValue(groupName=spark-infrastructure;propertyName=metastore.db.username)
+        description: Username to use against metastore database
     properties: {}
--- a/foundation/foundation-mda/src/main/resources/targets.json
+++ b/foundation/foundation-mda/src/main/resources/targets.json
@@ -2414,6 +2414,22 @@
     "overwritable": false
   },
   {
+    "name": "baseSparkInfrastructureProperties",
+    "templateName": "templates/deployment/spark-infrastructure/configurations/base/spark-infrastructure.properties.vm",
+    "outputFile": "configurations/base/spark-infrastructure.properties",
+    "generator": "com.boozallen.aiops.mda.generator.ModelAgnosticResourcesGenerator",
+    "metadataContext": "targeted",
+    "overwritable": false
+  },
+  {
+    "name": "envSparkInfrastructureProperties",
+    "templateName": "templates/deployment/spark-infrastructure/configurations/env/spark-infrastructure.properties.vm",
+    "outputFile": "configurations/env/spark-infrastructure.properties",
+    "generator": "com.boozallen.aiops.mda.generator.ModelAgnosticResourcesGenerator",
+    "metadataContext": "targeted",
+    "overwritable": false
+  },
+  {
     "name": "generatedPostActionInit",
     "templateName": "templates/python.init.py.vm",
     "outputFile": "post_action/__init__.py",
--- a/foundation/foundation-mda/src/test/java/com/boozallen/aiops/mda/generator/PropertiesStep.java
+++ b/foundation/foundation-mda/src/test/java/com/boozallen/aiops/mda/generator/PropertiesStep.java
@@ -1,0 +1,80 @@
+package com.boozallen.aiops.mda.generator;/*-
+ * #%L
+ * aiSSEMBLE::Foundation::MDA
+ * %%
+ * Copyright (C) 2021 Booz Allen
+ * %%
+ * This software package is licensed under the Booz Allen Public License. All Rights Reserved.
+ * #L%
+ */
+
+import com.boozallen.aiops.mda.metamodel.element.*;
+import io.cucumber.java.After;
+import io.cucumber.java.AfterStep;
+import io.cucumber.java.Before;
+import io.cucumber.java.Scenario;
+import io.cucumber.java.en.Given;
+import io.cucumber.java.en.Then;
+import io.cucumber.java.en.When;
+import org.apache.commons.io.FileUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.technologybrewery.fermenter.mda.GenerateSourcesHelper;
+import org.technologybrewery.fermenter.mda.element.ExpandedProfile;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.Assert.assertTrue;
+
+public class PropertiesStep extends AbstractModelInstanceSteps {
+    private static final Logger logger = LoggerFactory.getLogger(PropertiesStep.class);
+
+    @Before("@properties-generation")
+    public void setup(Scenario scenario) throws IOException {
+        this.scenario = scenario.getName();
+    }
+
+    @AfterStep("@properties-generation")
+    public void cleanup(Scenario scenario) throws IOException {
+        FileUtils.deleteDirectory(GENERATED_METADATA_DIRECTORY);
+    }
+
+    @Given("project called {string}")
+    public void project_called(String projectName) throws IOException {
+        createProject(projectName, "shared");
+    }
+
+    @Given("{string} pipeline is using {string}")
+    public void pipeline_using(String typeName, String implName) throws IOException {
+        createAndSavePipeline(unique("TestPipeline"), typeName, implName);
+    }
+
+    @When("the profile for {string} is generated")
+    public void profile_spark_infra_is_generated(String profileName) throws Exception {
+        readMetadata(projectName);
+        Map<String, ExpandedProfile> profiles = loadProfiles();
+        GenerateSourcesHelper.performSourceGeneration(profileName, profiles, this::createGenerationContext, (missingProfile, foundProfiles) -> {
+            throw new RuntimeException("Missing profile: " + missingProfile);
+        }, new Slf4jDelegate(logger), projectDir.toFile());
+    }
+
+    @Then("spark-infrastructure.properties file is generated in {string}")
+    public void properties_file_generated(String propertiesPath) {
+        Path properties = projectDir.resolve(propertiesPath);
+        assertTrue("File not created: " + properties, Files.exists(properties) && Files.isRegularFile(properties));
+    }
+
+    @Then("spark-infrastructure.properties file generated in {string}, {string} properties are set to {string}")
+    public void properties_set_values_correctly(String propertiesPath, String propertyName, String propertyValue) throws IOException {
+        Path properties = projectDir.resolve(propertiesPath);
+        List<String> lines = Files.readAllLines(properties);
+        String expectedPropertiesMap = propertyName + "=" + propertyValue;
+        assertTrue("Expected properties not found in " + propertiesPath, lines.contains(expectedPropertiesMap));
+    }
+
+
+}
--- a/foundation/foundation-upgrade/pom.xml
+++ b/foundation/foundation-upgrade/pom.xml
@@ -92,5 +92,10 @@
             <version>${version.cucumber.reporting.plugin}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.cucumber</groupId>
+            <artifactId>cucumber-java</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_11_0/SparkInfrastructureUniversalConfigYAMLMigration.java
+++ b/foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_11_0/SparkInfrastructureUniversalConfigYAMLMigration.java
@@ -1,0 +1,113 @@
+package com.boozallen.aissemble.upgrade.migration.v1_11_0;
+
+/*-
+ * #%L
+ * aiSSEMBLE::Foundation::Upgrade
+ * %%
+ * Copyright (C) 2021 Booz Allen
+ * %%
+ * This software package is licensed under the Booz Allen Public License. All Rights Reserved.
+ * #L%
+ */
+
+import com.boozallen.aissemble.upgrade.migration.AbstractAissembleMigration;
+import com.boozallen.aissemble.upgrade.util.YamlUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.yaml.snakeyaml.error.YAMLException;
+
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+
+/**
+ * This migration removes hive username in the hive-metastore-service values.yaml to use values from configuration store service only if values.yaml has default value.
+ */
+public class SparkInfrastructureUniversalConfigYAMLMigration extends AbstractAissembleMigration {
+    private static final Logger logger = LoggerFactory.getLogger(SparkInfrastructureUniversalConfigYAMLMigration.class);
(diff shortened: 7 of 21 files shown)


extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/values.yaml
```diff
@@ -15,11 +15,16 @@ image:
...
+  commonLabels:
+    aissemble-configuration-store: enabled
   enabled: true
   fullnameOverride: "hive-metastore-db"
   auth:
     # The schematool for the metastore service will create the database for us.
     database: "metastore"
+    # Note: Changing these values requires removal of the `hive-metastore-db-0` PVC, or manual modification of the
+    # persisted database.
+    username: $getConfigValue(groupName=hive;propertyName=HIVE_USERNAME)
```
carter-cundiff: Should update the helm chart README with these new values

foundation/foundation-mda/src/main/resources/templates/deployment/hive-metastore-service/hive.properties.vm
```diff
@@ -0,0 +1,4 @@
+
+# Password and other sensitive information shall be encrypted using krausening
```
carter-cundiff:
```suggestion
# Password and other sensitive information should be encrypted using krausening
```

foundation/foundation-mda/src/test/resources/specifications/data-records-generation.feature
```diff
@@ -15,7 +15,7 @@ Feature: Generating pipeline resources
       | type      | implementation        | packaging | dataRecordProfile                          |
       | data-flow | data-delivery-spark   | jar       | data-delivery-combined-data-records        |
       | data-flow | data-delivery-pyspark | habushu   | data-delivery-combined-data-records-python |
-
+
```
carter-cundiff: Can remove this change

foundation/foundation-mda/src/main/resources/targets.json
```diff
@@ -2413,6 +2413,22 @@
...
+    "overwritable": false
+  },
+  {
+    "name": "envHiveProperties",
+    "templateName": "templates/deployment/hive-metastore-service/hive.properties.vm",
+    "outputFile": "configurations/env/hive.properties",
+    "generator": "com.boozallen.aiops.mda.generator.ModelAgnosticResourcesGenerator",
+    "metadataContext": "targeted",
+    "overwritable": false
+  },
```
carter-cundiff, line 2431: Think we can just generate the base properties file. No point in having a env specific one that sets them to the exact same value. If the downstream uses wants them to be env specific, they can add that file.

foundation/foundation-upgrade/src/main/resources/migrations.json
```diff
@@ -52,6 +52,17 @@
...
+      {
+        "name": "hive-universal-config-yaml-migration",
+        "implementation": "com.boozallen.aissemble.upgrade.migration.v1_11_0.HiveUniversalConfigYAMLMigration",
+        "fileSets": [
+          {
+            "includes": [
+              "*-deploy/src/main/resources/apps/spark-infrastructure/values.yaml"
+            ]
+          }
+        ]
```
carter-cundiff, line 65: Should include this migration in the Automatic Upgrade section of the draft release notes

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_11_0/HiveUniversalConfigYAMLMigration.java
```diff
@@ -0,0 +1,110 @@
...
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+
+public class HiveUniversalConfigYAMLMigration extends AbstractAissembleMigration {
```
carter-cundiff: Recommend leaving a class level comment detailing what the migration aims to accomplish

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_11_0/HiveUniversalConfigYAMLMigration.java
```diff
@@ -0,0 +1,110 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(HiveUniversalConfigYAMLMigration.class);
+    private static final List<String> propertiesForConfigStore = Arrays.asList("javax.jdo.option.ConnectionUserName");
+    private static final List<String> mysqlAuthForConfigStore = Arrays.asList("username");
+    private static final HashSet<String> linesToRemove = new HashSet<>(Arrays.asList("username: hive", "- name: javax.jdo.option.ConnectionUserName" ,"value: hive", "description: Username to use against metastore database", "# NB: UCS"));
+    private static final String DEFAULT_HIVE_USERNAME_AND_PASSWORD = "hive";
+    @Override
+    protected boolean shouldExecuteOnFile(File file) {
+        if (file.getName().endsWith("values.yaml")) {
```
carter-cundiff: This check is redundant as baton is only going to select `values.yaml` files due to your configuration in the `migrations.json`

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_11_0/HiveUniversalConfigYAMLMigration.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Override
+    protected boolean performMigration(File file) {
+        try {
+            File tempFile = new File("tempValuesFile.yaml");
+            BufferedWriter writer = new BufferedWriter(new FileWriter(tempFile));
+            List<String> lines = Files.readAllLines(file.toPath());
+            for (int i = 0; i < lines.size(); i++) {
+                String line = lines.get(i).trim();
+                boolean noSkip = false;
```
carter-cundiff: Recommend changing this to be `skip`. Using `noSkip` is confusing as it creates a double negative

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_11_0/HiveUniversalConfigYAMLMigration.java
```diff
@@ -0,0 +1,110 @@
...
+                    List<YamlUtils.YamlObject> properties = yaml.getListOfObjects("configMap", "metastoreServiceConfig", "properties");
+                    for(YamlUtils.YamlObject property : properties)
+                    {
+                        if(propertiesForConfigStore.contains(property.get("name")) && !property.get("value").equals(DEFAULT_HIVE_USERNAME_AND_PASSWORD))
+                        {
+                            return false;
+                        }
+                    }
+                }
+                return true;
```
carter-cundiff: This method will still return true if the  `msql.auth` and `configMap.metastoreServiceConfig.properties` are present without the hive username attributes. Think it makes more sense to default the return to false and only return true when the default hive credentials are found.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_11_0/HiveUniversalConfigYAMLMigration.java
```diff
@@ -0,0 +1,110 @@
...
+
+            } catch (YAMLException e) {
+                if (logger.isDebugEnabled()) {
+                    logger.debug("Failed to parse YAML file, likely because it is a Helm template: {}", file.getPath(), e);
+                } else {
+                    logger.info("Failed to parse YAML file, likely because it is a Helm template: {}", file.getName());
```
carter-cundiff: This log message is not accurate as this migration should only be running against `values.yaml`.

foundation/foundation-upgrade/src/test/resources/specifications/v1_11_0/hive-uc-migration.feature
```diff
@@ -0,0 +1,20 @@
...
+  Given default values.yaml of spark infrastructure
+  When the 1.11.0 hive configuration migration executes
+  Then values.yaml is updated to remove hive username properties
+
+
+
+Scenario: Spark infrastructure file with modified hive credentials is not migrated
+Given values.yaml of spark infrastructure with hive username and password changed from "hive"
+When the 1.11.0 hive configuration migration executes
+Then hive configuration migration is skipped
```
carter-cundiff: Tab over statements under the scenario.

foundation/foundation-upgrade/src/test/resources/specifications/v1_11_0/hive-uc-migration.feature
```diff
@@ -0,0 +1,20 @@
...
+
+
+
+
```
carter-cundiff: Clean up excess whitespace

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/values.yaml
jaebchoi: updated!

foundation/foundation-mda/src/main/resources/targets.json
jaebchoi: Yup I tried without env but seems like I am getting error from config store when I stood up argoCD.
That was why I added env as placeholder and consumer can update freely.
Here are screenshot of error message without env.
Screenshot 2024-12-06 at 11 24 11 AM
Screenshot 2024-12-06 at 11 24 24 AM

foundation/foundation-upgrade/src/main/resources/migrations.json
jaebchoi: updated!

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_11_0/HiveUniversalConfigYAMLMigration.java
jaebchoi: Make sense. Removed the redundant check!

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_11_0/HiveUniversalConfigYAMLMigration.java
jaebchoi: Agreed, updated!

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_11_0/HiveUniversalConfigYAMLMigration.java
jaebchoi: updated log

foundation/foundation-mda/src/main/resources/targets.json
carter-cundiff: Does it work if you make the environment file blank? Think it could be useful make a generic envProperties template that just includes something like:
```
# Add environment specific overrides here. See the documentation for more details:
# https://boozallen.github.io/aissemble/aissemble/current/configuration-store.html#_environment_specific_overrides
```
That way as we build out more property files for the config store, we can reuse the generic one for our env property files.

DRAFT_RELEASE_NOTES.md
```diff
@@ -33,6 +33,7 @@ To reduce burden of upgrading aiSSEMBLE, the Baton project is used to automate t
 | upgrade-tiltfile-aissemble-version-migration       | Updates the aiSSEMBLE version within your project's Tiltfile                                                                                                                 |
 | upgrade-v2-chart-files-aissemble-version-migration | Updates the Helm chart dependencies within your project's deployment resources (`<YOUR_PROJECT>-deploy/src/main/resources/apps/`) to use the latest version of the aiSSEMBLE |
 | upgrade-v1-chart-files-aissemble-version-migration | Updates the docker image tags within your project's deployment resources (`<YOUR_PROJECT>-deploy/src/main/resources/apps/`) to use the latest version of the aiSSEMBLE       |
+| hive-universal-config-yaml-migration               | Updates hive-metastore-service values.yaml to removes hive username to use values from configuration store service only if hive username has default value.                  |
```
carter-cundiff:
```suggestion
| hive-universal-config-yaml-migration               | Removes the default hive username (if present) from hive-metastore-service values.yaml so that it can be set by the configuration store service         |
```

foundation/foundation-upgrade/pom.xml
```diff
@@ -92,5 +92,9 @@
             <version>${version.cucumber.reporting.plugin}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.cucumber</groupId>
+            <artifactId>cucumber-java</artifactId>
```
epwilkins, line 97: A: This should be `<scope>test</scope>`

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/templates/configmap.yaml
```diff
@@ -2,28 +2,41 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: metastore-service-config
+  labels:
+    aissemble-configuration-store: enabled
```
epwilkins: C: This needs to be templatized and set via `values` so it can be changed by downstream projects.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_11_0/HiveUniversalConfigYAMLMigration.java
```diff
@@ -0,0 +1,110 @@
...
+            if (yaml.hasList("configMap", "metastoreServiceConfig", "properties")) {
+                List<YamlUtils.YamlObject> properties = yaml.getListOfObjects("configMap", "metastoreServiceConfig", "properties");
+                for(YamlUtils.YamlObject property : properties)
+                {
+                    if(propertiesForConfigStore.contains(property.get("name")) && property.get("value").equals(DEFAULT_HIVE_USERNAME_AND_PASSWORD))
+                    {
+                        return true;
+                    }
+                   }
+            }
```
carter-cundiff: Think we want both of these conditions to be true in order to run the migration. So the return could be something like:
```
return hasMySqlAuthDefaultUsername && hasMetastoreServiceConfigDefaultUsername;
```
and you set those booleans in each of the inner most if blocks

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/values.yaml
```diff
@@ -15,11 +15,16 @@ image:
   dockerRepo: "ghcr.io/"
 
 mysql:
+  commonLabels:
```
epwilkins, line 18: A: Seems like we might want to be more precise and attach labels only to the resources that need it instead of using this `commonLables` field.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_11_0/HiveUniversalConfigYAMLMigration.java
```diff
@@ -0,0 +1,110 @@
...
+            File tempFile = new File("tempValuesFile.yaml");
+            BufferedWriter writer = new BufferedWriter(new FileWriter(tempFile));
+            List<String> lines = Files.readAllLines(file.toPath());
+            for (int i = 0; i < lines.size(); i++) {
+                String line = lines.get(i).trim();
+                boolean skip = true;
+                if(i >= 1 && lines.get(i-1).trim().equals("- name: javax.jdo.option.ConnectionPassword"))
+                {
+                    skip = false;
+                }
```
carter-cundiff: Why do we need this skip logic? Seems the `linesToRemove` below would be sufficient.

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/templates/configmap.yaml
```diff
@@ -2,28 +2,41 @@ apiVersion: v1
...
 data:
   # Add all the default properties from the local values.yaml to the ConfigMap
   # Then check if there are any downstream properties and add them as well
   metastore-site.xml: |
+    {{- $basePropDict := dict }}
```
epwilkins, line 11: S: Just use one dictionary. Write the base keys first, then write the props. That will automatically cause anything duplicate values in `properties` to take precedence over `baseProperties` without complex logic.

epwilkins: Actually I'm not really sure why we need these dictionaries at all. Can you not just do:
 `{{- if not hasProperty .Values.configMap.metastoreServiceConfig.properties $baseProperty.name }}`

foundation/foundation-mda/src/main/resources/templates/deployment/hive-metastore-service/hive.properties.vm
```diff
@@ -0,0 +1,4 @@
+
+# Password and other sensitive information should be encrypted using krausening
+# (See Here: https://github.com/TechnologyBrewery/krausening/tree/dev/krausening/#krausening-in-four-pints-leveraging-jasypt-for-encryptingdecrypting-properties)
+HIVE_USERNAME=hive
```
epwilkins: S: Properties files should generally follow Java package/variable naming conventions, i.e. `group.aSubGroup.propertyName`. Here I would suggest something like ~`hive.db.username`~ Actually since we're in `hive.properties`, maybe `database.username`/`jdo.username` or something like that.

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/values.yaml
jaebchoi: Yup! I wish I would inject label for specific to kind, but looking at mysql's logic all metadata's label logic is embedded to use only commonLabels. see here for example code snippet for mysql chart



```
{{- if eq .Values.architecture "replication" }}
apiVersion: {{ include "common.capabilities.statefulset.apiVersion" . }}
kind: StatefulSet
metadata:
  name: {{ include "mysql.secondary.fullname" . }}
  namespace: {{ include "common.names.namespace" . | quote }}
  labels: {{- include "common.labels.standard" ( dict "customLabels" .Values.commonLabels "context" $ ) | nindent 4 }}
    app.kubernetes.io/component: secondary
  {{- if .Values.commonAnnotations }}
  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
  {{- end }}
spec:
```

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_11_0/HiveUniversalConfigYAMLMigration.java
jaebchoi: Yup! Since ConnectionPassword's value also have same value of ConnectionUsername, without this logic, password's value will be wiped out. 


i.e. 

```
      <property>
        <name>javax.jdo.option.ConnectionUserName</name>
        <value>hive</value>
        <description>Username to use against metastore database</description>
      </property>
      <property>
        <name>javax.jdo.option.ConnectionPassword</name>
        <value>hive</value>
        <description>Password to use against metastore database</description>
      </property>
```

foundation/foundation-mda/src/test/java/com/boozallen/aiops/mda/generator/PropertiesStep.java
```diff
@@ -0,0 +1,76 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class PropertiesStep extends AbstractModelInstanceSteps {
+    private static final Logger logger = LoggerFactory.getLogger(PropertiesStep.class);
+
+    @Before("@properties-generation")
+    public void setup(Scenario scenario) throws IOException {
+        this.scenario = scenario.getName();
+        FileUtils.deleteDirectory(GENERATED_METADATA_DIRECTORY);
```
epwilkins, line 43: A: I would have expected this to be part of a teardown step not setup.  If this was necessary, it may mean we have some other test that isn't cleaning up after itself properly.

foundation/foundation-mda/src/main/resources/templates/deployment/hive-metastore-service/hive.properties.vm
epwilkins: Ok, I'm revisiting this again.  _Actually_, I think it's worth a little more thought toward what organization of properties makes sense in the long run.  This is configuring the Spark metastore. So it probably makes more sense to name the file `spark.properties` and the property `metastore.db.username`.  We just need to look at pre-existing config files that might be called `spark.properties` and make sure it all  works together.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_11_0/HiveUniversalConfigYAMLMigration.java
carter-cundiff: Let's make `skip` be `skipConnectionPassword` then so it's easier to understand and default that to `false`.

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/templates/configmap.yaml
jaebchoi: Sorry I'm bit new to helm syntax but from what I looked at the doc it has some limitation in creating variable and function so I had to go with dicts but open to idea and see if there's easy way to do it.

is hasProperty custom method? wasn't sure it was doable, even if it was custom methods,, shouldn't hasProperty need dict to accomplish the logic?

Im trying to accomplish 3 use case,
1. add baseProperty that doesn't exist in properties
2. add properties that doesn't exist in baseProperties
3. Add properties that both exist in baseProperties and properties ( take properties as precedence ) 

it's bit hard to accomplish contain logic in helm because property schema is list of dict.

foundation/foundation-mda/src/main/resources/templates/deployment/hive-metastore-service/hive.properties.vm
jaebchoi: looking at the aissemble repo and downstream project repo that uses Spark doesn't have config file called `spark.properties` or anything similar in that sense. 
I agree being spark. but since this is more of infrastructure. I think it also makes sense to name it as `spark-infrastructure.properties` unless anyone objects

foundation/foundation-mda/src/test/java/com/boozallen/aiops/mda/generator/PropertiesStep.java
jaebchoi: yup you are correct, that was legacy of my copy pasta. Thank you for pointing out! cleaning up..

foundation/foundation-mda/src/main/resources/targets.json
jaebchoi: yup it is working with empty properties file in env =) . Ill update to make this one empty.

foundation/foundation-mda/src/main/resources/templates/deployment/hive-metastore-service/hive.properties.vm
carter-cundiff: Would update the template locations within foundation mda to reflect the new file name, like move the `hive.properties.vm` from `hive-metastore-service` to `spark-infrastructure`

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_11_0/HiveUniversalConfigYAMLMigration.java
```diff
@@ -0,0 +1,111 @@
...
+
+            }
+            boolean hasMetastoreServiceConfigDefaultUsername = false;
+            if (yaml.hasList("configMap", "metastoreServiceConfig", "properties")) {
+                List<YamlUtils.YamlObject> properties = yaml.getListOfObjects("configMap", "metastoreServiceConfig", "properties");
+                for(YamlUtils.YamlObject property : properties)
+                {
+                    if(propertiesForConfigStore.contains(property.get("name")) && property.get("value").equals(DEFAULT_HIVE_USERNAME_AND_PASSWORD))
+                    {
+                        hasMetastoreServiceConfigDefaultUsername = true;
```
carter-cundiff: Since this is the only case in the for loop, you can `break` once it's true.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_11_0/SparkInfrastructureUniversalConfigYAMLMigration.java
```diff
@@ -0,0 +1,112 @@
...
+                return false;
+            }
+            boolean hasMySqlAuthDefaultUsername = false;
+            if (yaml.hasObject("mysql", "auth")) {
+                YamlUtils.YamlObject mysqlAuth = yaml.getObject("mysql", "auth");
+                for(String key : mysqlAuth.keySet())
+                {
+                    if(mysqlAuthForConfigStore.contains(key) && mysqlAuth.get(key).equals(DEFAULT_METASTORE_DB_USERNAME))
+                    {
+                        hasMySqlAuthDefaultUsername = true;
```
carter-cundiff, line 54: Can `break` here as well

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/v1_11_0/SparkInfrastructureUniversalConfigYAMLMigration.java
```diff
@@ -0,0 +1,112 @@
...
+                boolean skipConnectionPassword = false;
+                if(i >= 1 && lines.get(i-1).trim().equals("- name: javax.jdo.option.ConnectionPassword"))
+                {
+                    skipConnectionPassword = true;
+                }
+                if(linesToRemove.contains(line) && !skipConnectionPassword)
+                {
+                    continue;
+                }
+                writer.write(lines.get(i) + System.getProperty("line.separator"));
```
carter-cundiff, line 104: I believe this logic will break in the event the password appears before the username:
```
        - name: javax.jdo.option.ConnectionPassword
          value: hive
          description: Password to use against metastore database
        - name: javax.jdo.option.ConnectionUserName
          value: hive
          description: Username to use against metastore database
```

Think it could be resolved by updating the logic to the following:
```
                if(linesToRemove.contains(line))
                {
                    if(!skipConnectionPassword) {
                        continue;
                   } else {
                       skipConnectionPassword = false;
                   }
                }
```

jaebchoi: Hmm not sure my logic fails with the case you mentioned,

what I am checking for is if name is javax.jdo.option.ConnectionPassword (which is previous line i-1)we set skipConnectionPassword to true and only skip writing file if skipConnectionPassword is true and contains value: hive

but after that iteration skipConnectionPassword defaults back to false on every line read so no matter password comes first and set skipConnectionPassword  to true, that value will not carry over on next iteration. so I think my logic still stand
 
 I did try unit test with password coming first and it still passes

carter-cundiff: Oh you're right, didn't realize the boolean resets each iteration

carter-cundiff (approved):

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/templates/configmap.yaml
jaebchoi: I can def cut follow up ticket if this requires refactoring given that this ticket is past DPC1!

jaebchoi: added https://jira.boozallencsn.com/browse/AIOPS-3838 as a follow up and will be working this

Issue #513, opened 2024-12-19
Feature:Remove in-flight changes that prevent immediate release


epwilkins:
## Description
In our recent work on path to production alignment and baking in the universal config store, we've drifted away from our goal of always being ready to cut a release.  We want to get back to a good state and move any changes that requires further work to fully polish to feature branches so we can respond more quickly to demand.

## DOD
- [x] Path to production alignment changes that prevent deployment via Tilt or introduce files to support local ArgoCD deployment are moved to a feature branch
  - [x] New infrastructure module is removed
  - [x] New deploy shell script it removed
  - [x] Other changes are analyzed and removed as appropriate
  - [x] Remove any migrations specific to the removed changes
- [x] Universal config work that will force projects to setup the config store are moved to a feature branch
  - [x] Default usage of injection hooks in spark-infrastructure charts is removed
  - [x] Other changes are analyzed and removed as appropriate
  - [x] Remove any migrations specific to the removed changes
- [x] Ensure release notes do not include any instructions/information about changes that have been moved to feature branches
- [x] Ensure Antora docs do not contain information for changes moved to feature branches
- [x] For tickets that introduced the changes moved to feature branches, ensure a paper trail is left leading back to this ticket

## Test Strategy/Script
1. Create a new project
```sh
mvn archetype:generate '-DarchetypeGroupId=com.boozallen.aissemble' \
                       '-DarchetypeArtifactId=foundation-archetype' \
                       '-DarchetypeVersion=1.11.0-SNAPSHOT' \
                       '-DgroupId=org.test' \
                       '-Dpackage=org.test' \
                       '-DprojectGitUrl=test.org/test.git' \
                       '-DprojectName=Test Project' \
                       '-DartifactId=final-513'
```
2. Verify there is no _final-513-infrastructure_ module or a _deploy.sh_ script
3. Add [SparkPipeline.json](https://github.com/user-attachments/files/18203992/SparkPipeline.json) to _final-513-pipeline-models/src/main/resources/pipelines_
4. Build the project and resolve manual actions until none are printed
5. Run the build again with the cache disabled and apply final manual actions: `mvn clean install -Dmaven.build.cache.skipCache`
6. Deploy the project with `tilt up`
7. Verify all resource start successfully
8. Run the **spark-pipeline** resource and verify it completes successfully

epwilkins:
**DoD** with @nartieri

csun-cpointe:
OTS completed!

[...]

PR #515, opened 2024-12-20
[#513] remove noRdAdmin profile

epwilkins:
The TestContainers now respects the docker context so an explicit profile is no longer needed. Also fixes the healthcheck URL for drift-detection tests so the ITs don't wait the full timeout every time.

Ideally, we would use docker socket detection logic with `file.exists` and `file.missing` profile activations to update the `dockerHost` configuration or `docker.host` property of the docker-maven-plugin as indicated by the build error message when DOCKER_HOST is unset. However, because that plugin creates it's own `config.json` with only auth settings and always passes in `--config` during a buildx build, the builder will always look for the docker socket at /var/run/docker.sock. We will need to update the docker-maven-plugin source code to respect `dockerHost` during this config creation.

### DIFF
--- a/DRAFT_RELEASE_NOTES.md
+++ b/DRAFT_RELEASE_NOTES.md
@@ -18,7 +18,15 @@ _Note: instructions for adapting to these changes are outlined in the upgrade in
 
 
 # Known Issues
-_There are no known issues with the 1.11 release._
+
+## Docker Module Build Failures
+When using a Docker daemon that does not reside in `/var/run` (e.g. running Rancher Desktop without admin privileges) the docker-maven-plugin will fail to build with the message below. To work around this failure, set the `DOCKER_HOST` variable to the location of the daemon socket file. For example, to make the docker-maven-plugin work with Rancher Desktop, run `export DOCKER_HOST=unix://$HOME/.rd/docker.sock`.
+
+```shell
+[ERROR] Failed to execute goal org.technologybrewery.fabric8:docker-maven-plugin:0.45-tb-0.1.0:build (default-build) on project final-513-spark-worker-docker:
+   Execution default-build of goal org.technologybrewery.fabric8:docker-maven-plugin:0.45-tb-0.1.0:build failed: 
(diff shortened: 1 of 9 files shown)


DRAFT_RELEASE_NOTES.md
```diff
@@ -18,7 +18,15 @@ _Note: instructions for adapting to these changes are outlined in the upgrade in
 
 
 # Known Issues
-_There are no known issues with the 1.11 release._
+
+## Docker Module Build Failures
```
epwilkins, line 22: I: Putting this in known issues to keep pressure on us to smooth this out.

build-parent/pom.xml
```diff
@@ -1132,6 +1132,45 @@ To suppress enforce-helm-version rule, you must add following plugin to the root
                 <hide.manual.actions>true</hide.manual.actions>
             </properties>
         </profile>
+
+        <!-- Docker host configuration by detecting common docker socket locations -->
+        <profile>
+            <!-- Rancher Desktop on Linux/Mac -->
```
ewilkins-csi, line 1138: I: It would be trivially easy to include Docker Desktop as well, but since we strongly recommend Rancher Desktop I didn't add it.

foundation/foundation-drift-detection/aissemble-foundation-drift-detection-client-python/tests/features/environment.py
```diff
@@ -38,7 +38,7 @@ def before_all(context):
     )
 
     # call to health check to verify Drift Detection Services are started
-    health_check_url = "http://localhost:8084/q/health/live"
+    health_check_url = "http://localhost:8084/invoke-drift/health-check"
```
epwilkins, line 41: I: The previous healthcheck URL is not a valid endpoint in our deployment. This was causing the ITs to wait the full timeout and eventually print an error before continuing on to successfully execute the actual tests.

cwoods-cpointe (approved):

PR #517, opened 2024-12-20
[#498] perform config injection on update

### LINKED ISSUE #498: Feature: Configuration Store injection should happen on update
## Description
Currently, the configuration store only injects values on object creation.  This means if an object is _updated_ with an injection point, it is not processed.  This especially causes issues in ArgoCD, where changing an non-injected field will lead to previously injected fields being replaced with their injection call.

This actually causes a pretty severe issue with `StatefulSets` and `DaemonSets`. Those objects [create revision](https://shivering-isles.com/til/2024/01/statefulset-revisions) objects called `ControllerRevisions` to track state. Without update functionality, the newly created revision gets injection performed but the actual workload object does not. This causes the set to see a difference between the current state and the recorded revision, so it redeploys a new revision to track the new state. But again, this newly created revision gets injection and does not match, leading to an infinite loop of creating `ControllerRevisions`.  We should ensure that after this feature is implemented this no longer happens.

## DOD
- [x] `MutatingWebhookConfiguration` for Config Store includes the `UPDATE` operation

## Test Strategy/Script
1. Deploy the config store helm chart
```sh
helm install aissemble-config oci://ghcr.io/boozallen/aissemble-configuration-store-chart --version 1.11.0-SNAPSHOT --set 'aissemble-quarkus-chart.deployment.env[0].name=BASE_PROPERTY' --set 'aissemble-quarkus-chart.deployment.env[0].value=/tmp'
```
2. Verify the webhook config includes `UPDATE` under `webhooks.rules.operations`. (You might need to wait a couple of minutes for the webhook config to be created)
```sh
kubectl get mutatingwebhookconfiguration boozallen.aissemble-configuration-store.webhook -o yaml
```
3. Uninstall the chart
```sh
helm uninstall aissemble-config
```

epwilkins:
Updates the webhook configuration to also route resource updates to the config store for config injection.  Also:
 - Remove unneeded files form devops
 - Improve logging when base property location is not set
 - Prevent crashing if optional environments overrides aren't set


devops/clusterrole-configmap-manager-default.yml
epwilkins, line 1: I: These were never meant to be checked in and were an active development artifact.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/ConfigStoreInit.java
```diff
@@ -93,31 +93,32 @@ public void init() {
     }
 
     private void pullConfigs() throws IOException {
```
epwilkins, line 95: I: Small refactor to de-duplicate some code, improve log output when BASE_PROPERTY isn't set and ensure landing directories for extensions are always created even if the ENVIRONMENT_PROPERTY isn't set. This prevents the service from crashing if env overrides aren't provided.

foundation/foundation-upgrade/pom.xml
```diff
@@ -92,10 +92,5 @@
             <version>${version.cucumber.reporting.plugin}</version>
             <scope>test</scope>
         </dependency>
-        <dependency>
```
epwilkins, line 95: I: resolve POM warning about duplicately declared dependency.

cwoods-cpointe (approved):

csun-cpointe (approved):

Issue #497, opened 2024-12-06
Feature: Configuration Store should be able to inject into data fields of Secrets


epwilkins:
## Description
Any value in the `data` field of a Kubernetes `Secret` are required to be base64 encoded.  This means the standard injection string of `$getConfig(group, property)` must be base64 encoded before being submitted to a cluster. The Config Store injection service should recognize when a Secret is being processed and decode its data values to ensure full injection.

## DOD
- [x] Update Webhook to comply with Secret injection by supporting base64 encoding
- [x] add unit test if appropriate.
- [x] Verify Downstream project with argoCD can be stood up locally without any error
- [x] The configuration store webhook can inject plain data field.
- [x] The configuration store webhook can inject a data field in an opaque Secret

## Test Strategy/Script

- Pull 497-test branch (This issue is merged into feature branch Feature/Universal-Config and need some deploy script in feature/path2production-alignment for easily testing so this 497-test branch is just Universal-Config branch with deploy script added.  )

- Create a downstream project with baseline version 1.11.0-SNAPSHOT
```
mvn archetype:generate -U -DarchetypeGroupId=com.boozallen.aissemble \
  -DarchetypeArtifactId=foundation-archetype \
  -DarchetypeVersion=1.11.0-SNAPSHOT \
  -DgroupId=com.test \
  -DartifactId=test-497 \
  -DprojectGitUrl=https://github.com/{your-github-username}/test-497 \
  -DprojectName=test-497 \
  && cd test-497 
```

Replace {your-github-username} with your github username

- Add the [SparkPipeline.json](https://github.com/user-attachments/files/16781488/SparkPipeline.json) file to the test-497-pipeline-models/src/main/resources/pipelines directory
- Run mvn clean install until all the manual actions are complete
- Once the manual actions are complete, run `mvn clean install -Dmaven.build.cache.skipCache=true` once to get any remaining manual actions
- Modify the helm templates for Argocd deployment:
       - Mac User: In the `-deploy/src/main/resources/apps/configuration-store/values-dev.yaml file` update the `volumePathOnNode` to be `/<pathToProject>/test-496-deploy/src/main/resources/configurations`
       - Window User: In the `-deploy/src/main/resources/values-dev.yaml file` update the `volumePathOnNode` to be `/mnt/c/Users/YOUR_USER/PATH/TO/test-496-deploy/src/main/resources/configurations`

- Modify Hive metastore chart
[...]

PR #522, opened 2025-01-06
#497 Feature: Configuration Store should be able to inject into data fields of Secrets

jaebchoi:
Any value in the data field of a Kubernetes Secret are required to be base64 encoded. This means the standard injection string of $getConfig(group, property) must be base64 encoded before being submitted to a cluster. The Config Store injection service should recognize when a Secret is being processed and decode its data values to ensure full injection.

### DIFF
--- a/extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/README.md
+++ b/extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/README.md
@@ -45,11 +45,17 @@ helm install hive-metastore-service oci://ghcr.io/boozallen/aissemble-hive-metas
 See [the official bitnami documentation](https://github.com/bitnami/charts/tree/main/bitnami/mysql) for full
 configuration options.
 
-| Property         | Default             |
-|------------------|---------------------|
-| fullnameOverride | "hive-metastore-db" |
-| auth.database    | "metastore"         |
-| auth.username    | "hive"              |
+| Property         | Default                                                                            |
+|------------------|------------------------------------------------------------------------------------|
+| fullnameOverride | "hive-metastore-db"                                                                |
+| auth.database    | "metastore"                                                                        |
+| auth.username    | $getConfigValue(groupName=spark-infrastructure;propertyName=metastore.db.username) |
+| auth.password    | $getConfigValue(groupName=spark-infrastructure;propertyName=metastore.db.password) |
+
+**Note**: 
+By Default, username and passwords are being injected from [Universal Configuration Store](https://boozallen.github.io/aissemble/aissemble/current/configuration-store.html).
+Configuration Store will look for $getConfigValue(...) and find properties file to inject values from corresponding groupName and propertyName.
+
 
 # Migration from aiSSEMBLE v1 Helm Charts
 
--- a/extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/values.yaml
+++ b/extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/values.yaml
@@ -25,6 +25,7 @@ mysql:
     # Note: Changing these values requires removal of the `hive-metastore-db-0` PVC, or manual modification of the
     # persisted database.
     username: $getConfigValue(groupName=spark-infrastructure;propertyName=metastore.db.username)
+    password: $getConfigValue(groupName=spark-infrastructure;propertyName=metastore.db.password)
 
 hive:
   dbType: "mysql"
--- a/foundation/foundation-configuration-store/pom.xml
+++ b/foundation/foundation-configuration-store/pom.xml
@@ -165,6 +165,11 @@
             <artifactId>jakarta.ws.rs-api</artifactId>
             <version>${version.jakarta.wr.rs}</version>
         </dependency>
+        <dependency>
+            <groupId>com.tdunning</groupId>
+            <artifactId>json</artifactId>
+            <version>1.8</version>
+        </dependency>
 
         <!-- quarkus dependencies -->
         <dependency>
--- a/foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
+++ b/foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
@@ -9,17 +9,20 @@ package com.boozallen.aissemble.configuration.mutating.webhook;
  * This software package is licensed under the Booz Allen Public License. All Rights Reserved.
  * #L%
  */
- 
+
 import com.boozallen.aissemble.configuration.store.ConfigLoader;
 import com.boozallen.aissemble.configuration.store.Property;
 import com.boozallen.aissemble.configuration.store.PropertyKey;
 import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.json.JsonReadFeature;
+import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.flipkart.zjsonpatch.JsonDiff;
 import io.fabric8.kubernetes.api.model.StatusBuilder;
 import org.apache.commons.lang3.StringUtils;
+import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -36,6 +39,7 @@ import jakarta.ws.rs.Produces;
 import jakarta.ws.rs.POST;
 import jakarta.ws.rs.Path;
 import jakarta.ws.rs.core.MediaType;
+
 import java.util.Base64;
 import java.util.HashMap;
 import java.util.Map;
@@ -58,7 +62,7 @@ public class ConfigMutatingWebhook {
 
     @Inject
     public ConfigLoader configLoader;
-    
+
     @POST
     @Path("/process")
     @Consumes(MediaType.APPLICATION_JSON)
@@ -83,9 +87,18 @@ public class ConfigMutatingWebhook {
         AdmissionResponse response;
         try {
             String json = mapper.writeValueAsString(request.getObject());
+
+            //This scans to see if there's any config need to be injected and replace corresponding values and saves into updatedJson
             String updatedJson = replaceConfigStoreKeysWithConfigValue(json);
 
-            if (!updatedJson.isEmpty()) {
+            if(request.getKind().getKind().equals("Secret")) {
+                //if replaceConfigStoreKeysWithConfigValue doesn't have any injection it would return empty, so we need original json to scan.
+                String secretJson = updatedJson.isEmpty() ? json: updatedJson;
+                boolean plainTextInjectionUpdated = !updatedJson.isEmpty();
+                updatedJson = replaceSecretConfigStoreKeyWithConfigValue(secretJson, mapper, plainTextInjectionUpdated);
+            }
+
+            if (!updatedJson.isEmpty()){
                 JsonNode patch = JsonDiff.asJson(mapper.readTree(json), mapper.readTree(updatedJson));
                 response = new AdmissionResponseBuilder()
                         .withAllowed(true)
@@ -94,7 +107,7 @@ public class ConfigMutatingWebhook {
                         .withPatchType("JSONPatch")
                         .build();
                 logger.info("The kubernetes resource request has been updated.");
-            } else {
+            }else {
                 response = new AdmissionResponseBuilder()
                         .withAllowed(true)
                         .withUid(request.getUid())
@@ -129,9 +142,58 @@ public class ConfigMutatingWebhook {
         }
     }
 
+
+    /**
+     * This method is to inject values in the data fields for any encoded Configvalues from the properties file and save it as based64 encoded.
+     *
+     * @param json json string to scan
+     * @param mapper json object mapper
+     * @param plainTextInjectionUpdated boolean that plain text injection is updated.
+     * @return updatedJSON  If no injection, return empty.
+     */
+    private String replaceSecretConfigStoreKeyWithConfigValue(String json, ObjectMapper mapper, boolean plainTextInjectionUpdated) throws JsonProcessingException {
+        String updatedJSON = json;
+        JSONObject jsonObject = new JSONObject(json);
+        //If there's no data key, no injection can be done.
+        if(!jsonObject.has("data")) {
+            return plainTextInjectionUpdated ? json : "";
+        }
+        String dataString = jsonObject.getString("data");
+        Map<String, String> secretDataMap = mapper.readValue(dataString, new TypeReference<>() {});
+
+        //key being base64 val that has getConfigValue value being actual value to inject from property.
+        Map<String, String> injectionMap = new HashMap<>();
+
+        for(String secretDataVal : secretDataMap.values()) {
+            String decodedStr = new String(Base64.getDecoder().decode(secretDataVal));
+
+            if(decodedStr.startsWith(CONFIG_STORE_INJECT_START)) {
+                //trim out unnecessary strings
+                String[] injectionKeyValuePair = decodedStr.substring(CONFIG_STORE_INJECT_START.length(),decodedStr.length()-1).split(PROPERTY_KEY_DELIMITER);
+                Map<String, String> injectionKeyValueMap = convertToKeyMap(injectionKeyValuePair);
+                String propertyValue = getProperty(injectionKeyValueMap.get(GROUP_NAME), injectionKeyValueMap.get(PROPERTY_NAME));
+                injectionMap.put(secretDataVal, new String(Base64.getEncoder().encode(propertyValue.getBytes())));
+            }
+        }
+
+        for(Map.Entry<String, String> injection: injectionMap.entrySet()) {
+            updatedJSON = updatedJSON.replace(injection.getKey(), injection.getValue());
+        }
+
+        //We still need to carry out plain text injection differences.
+        if(plainTextInjectionUpdated) {
+            return updatedJSON.equals(json) ? json : updatedJSON;
+        }
+
+        //If there is no change in injection in both secret and plain text, return empty. If there are any change, return updated JSON.
+        return updatedJSON.equals(json) ? "" : updatedJSON;
+
+    }
+
     private String replaceConfigStoreKeysWithConfigValue(String json) {
         StringBuilder updatedJson = new StringBuilder();
         String[] data = json.split(Pattern.quote(CONFIG_STORE_INJECT_START));
+
         // if the config store key is found
         if (data.length > 1) {
             updatedJson.append(data[0]);
@@ -161,5 +223,4 @@ public class ConfigMutatingWebhook {
 
         return map;
     }
- }
- 
+}
--- a/foundation/foundation-configuration-store/src/test/java/com/boozallen/aissemble/configuration/mutating/webhook/MutatingWebhookSteps.java
+++ b/foundation/foundation-configuration-store/src/test/java/com/boozallen/aissemble/configuration/mutating/webhook/MutatingWebhookSteps.java
@@ -23,7 +23,10 @@ import io.fabric8.kubernetes.api.model.ConfigMap;
 import io.fabric8.kubernetes.api.model.ConfigMapBuilder;
 import io.fabric8.kubernetes.api.model.GroupVersionKind;
 import io.fabric8.kubernetes.api.model.GroupVersionResource;
+import io.fabric8.kubernetes.api.model.KubernetesResource;
 import io.fabric8.kubernetes.api.model.ObjectMeta;
+import io.fabric8.kubernetes.api.model.Secret;
+import io.fabric8.kubernetes.api.model.SecretBuilder;
 import io.fabric8.kubernetes.api.model.admission.v1.AdmissionRequest;
 import io.fabric8.kubernetes.api.model.admission.v1.AdmissionReview;
 import io.fabric8.kubernetes.api.model.admission.v1.AdmissionReviewBuilder;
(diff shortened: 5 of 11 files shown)


foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -129,9 +148,63 @@ private String getProperty(String groupName,String propertyName) {
         }
     }
 
+    //This method is to inject values in the data fields for any encoded Configvalues from the properties file and save it as based64 encoded.
+    private String replaceSecretConfigStoreKeyWithConfigValue(String json, ObjectMapper mapper) throws JsonProcessingException {
```
jaebchoi: I: As feature stated, this logic is only applicable for data fields of Secrets

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -129,9 +148,63 @@ private String getProperty(String groupName,String propertyName) {
...
+        }else{
+            updatedJSON = json;
+        }
+
+        JSONObject jsonObject = new JSONObject(json);
+        if(!jsonObject.has("data"))
+        {
+            return updatedJSON;
+        }
+        String dataString = jsonObject.getString("data");
```
jaebchoi, line 161: I: in case of plain text, they just pull out raw json string and replace any values starting with $getConfigValue(... 
But for encoded case, it never works because encoded value always change as group or values of config values change and you can't really tell which substring is base64-ed or not off of raw json string, so you would need to basically change it to JSON Object, and go through all fields of data and check if value is base64 encoded, and decode to see if it starts with $getConfigValue(...

foundation/foundation-configuration-store/pom.xml
```diff
@@ -207,5 +212,11 @@
             <version>${version.quarkus.cucumber}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.tdunning</groupId>
+            <artifactId>json</artifactId>
+            <version>1.8</version>
+            <scope>compile</scope>
```
cwoods-cpointe: a: compile is default so you dont need to specify it directly. And if its not test scope then we should move it out of the unit test section

foundation/foundation-configuration-store/pom.xml
```diff
@@ -165,6 +165,11 @@
             <artifactId>jakarta.ws.rs-api</artifactId>
             <version>${version.jakarta.wr.rs}</version>
         </dependency>
+        <dependency>
+            <groupId>commons-codec</groupId>
+            <artifactId>commons-codec</artifactId>
+            <version>1.17.1</version>
```
cwoods-cpointe: a: Might be nice to have these versions be properties in the base pom

foundation/foundation-mda/src/main/resources/templates/deployment/spark-infrastructure/v2/spark.infrastructure.values.yaml.vm
```diff
@@ -52,7 +52,6 @@ aissemble-hive-metastore-service-chart:
       # persisted database.
       rootPassword: hive
       replicationPassword: hive
-      password: hive
```
cwoods-cpointe, line 55: q: Is it worth adding the rootPassword and replicationPassword as well?

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -84,17 +89,31 @@ public AdmissionReview validate(AdmissionReview admissionReviewRequest) {
...
             String json = mapper.writeValueAsString(request.getObject());
             String updatedJson = replaceConfigStoreKeysWithConfigValue(json);
+            String updatedSecretJson = "";
 
-            if (!updatedJson.isEmpty()) {
+            if(request.getKind().getKind().equals("Secret")) {
+                updatedSecretJson = replaceSecretConfigStoreKeyWithConfigValue(json, mapper);
+            }
+
+            if (request.getKind().getKind().equals("Secret") && !StringUtils.difference(updatedSecretJson, json).isEmpty()) {
```
meliz19: This seems redundant with line 94. I know nested if statements aren't the best, so I don't really have a suggestion, but I'm curious if others have a take on this?

foundation/foundation-mda/src/main/resources/templates/deployment/spark-infrastructure/v2/spark.infrastructure.values.yaml.vm
jaebchoi: We can discuss further, but upon standing up hive db, password was only thing that uses to gain access to mysql db. 
If I remember correctly during convo with baseline team,  we would inject this password only.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
cwoods-cpointe: Yeah inside is good. Id also use StringUtils.equals instead of .difference if all we care about is if they are equal or not. Might be worth having a comment on what it means to have the updatedSecretJson and json be different

jaebchoi: Word, Ill update to use .equals and additional comments!

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -129,9 +151,63 @@ private String getProperty(String groupName,String propertyName) {
...
+        String plainJsonInjection = replaceConfigStoreKeysWithConfigValue(json);
+
+        String updatedJSON = null;
+
+        if(!plainJsonInjection.isEmpty())
+        {
+            updatedJSON = plainJsonInjection;
+        }else{
+            updatedJSON = json;
+        }
```
cwoods-cpointe: a: if else can be simplified with a ternary operator

foundation/foundation-configuration-store/src/test/java/com/boozallen/aissemble/configuration/mutating/webhook/MutatingWebhookSteps.java
```diff
@@ -93,11 +127,33 @@ public void theProcessedKubernetesResourceIsReturned() throws JsonProcessingExce
         assertNotNull(responsePatch);
     }
 
+    @Then("the Secret patch is returned")
+    public void theProcessedKubernetesResourceIsReturnedForSecret() throws JsonProcessingException {
```
epwilkins: A: This looks identical to `theProcessedKubernetesResourceIsReturned`. Can we not just re-use that step?

foundation/foundation-configuration-store/src/test/java/com/boozallen/aissemble/configuration/mutating/webhook/MutatingWebhookSteps.java
```diff
@@ -93,11 +127,33 @@ public void theProcessedKubernetesResourceIsReturned() throws JsonProcessingExce
...
     @Then("the ConfigMap patch contains the injected value")
     public void theConfigMapDefinitionContainsTheInjectedValue() throws JsonProcessingException {
         JsonNode jsonPatch = objectMapper.readTree(new String(Base64.getDecoder().decode(responsePatch)));
         assertEquals("The response patch does not contain the injected value", expectedInjectedValue, jsonPatch.findValue("value").asText());
     }
+
+    @Then("the Secret patch contains the encoded injected value")
+    public void theSecretDefinitionContainsTheInjectedValue() throws JsonProcessingException {
+        JsonNode jsonPatch = objectMapper.readTree(new String(Base64.getDecoder().decode(responsePatch)));
+        assertEquals("The response patch does not contain the injected value", new String(Base64.getEncoder().encode(expectedInjectedSecretValue.getBytes())), jsonPatch.findValue("value").asText());
```
epwilkins: A: The logic gets a bit lost in the length of this line. It'd be easier to see what we were asserting if we pulled out some local variables. Especially for the expected value.

foundation/foundation-configuration-store/src/test/java/com/boozallen/aissemble/configuration/mutating/webhook/MutatingWebhookSteps.java
```diff
@@ -119,4 +175,26 @@ private AdmissionReview createAdmissionReviewRequest() {
             .withRequest(request)
             .build();
     }
+
+    /**
+     * Create an {@link AdmissionReviewRequest} for testing
+     */
+    private AdmissionReview createAdmissionReviewRequestForSecret() {
```
epwilkins: Q/A: Why not just have `createAdmissionReviewRequest` take the object and kind as parameters?

foundation/foundation-configuration-store/src/test/java/com/boozallen/aissemble/configuration/mutating/webhook/MutatingWebhookSteps.java
```diff
@@ -79,6 +98,21 @@ public void aKubernetesResourceRequestIsMade() throws JsonProcessingException {
                 .then();
     }
 
+    @When("a kubernetes resource request is made to create a Secret")
+    public void aKubernetesResourceRequestIsMadeForSecret() throws JsonProcessingException {
```
epwilkins, line 127: A: Again it seems like we could re-use the existing step as the logic seems identical.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -129,9 +151,63 @@ private String getProperty(String groupName,String propertyName) {
...
+    //This method is to inject values in the data fields for any encoded Configvalues from the properties file and save it as based64 encoded.
+    private String replaceSecretConfigStoreKeyWithConfigValue(String json, ObjectMapper mapper) throws JsonProcessingException {
+
+        //inject any plain text configuration store values first.
+        String plainJsonInjection = replaceConfigStoreKeysWithConfigValue(json);
+
+        String updatedJSON = null;
+
+        if(!plainJsonInjection.isEmpty())
+        {
```
epwilkins: A: This doesn't match the standard code styling of the project. It should be possible to setup your IDE to use this styling automatically for this project.
```suggestion
        if (!plainJsonInjection.isEmpty()) {
```

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -129,9 +151,63 @@ private String getProperty(String groupName,String propertyName) {
         }
     }
 
+    //This method is to inject values in the data fields for any encoded Configvalues from the properties file and save it as based64 encoded.
+    private String replaceSecretConfigStoreKeyWithConfigValue(String json, ObjectMapper mapper) throws JsonProcessingException {
+
+        //inject any plain text configuration store values first.
+        String plainJsonInjection = replaceConfigStoreKeysWithConfigValue(json);
```
epwilkins: S: Isn't this redundant processing? We'd have already replaced plaintext values on line 93. We just need to check the base64 encoded fields at this stage.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -129,9 +151,63 @@ private String getProperty(String groupName,String propertyName) {
...
+                    String value = getProperty(keyMap.get(GROUP_NAME), keyMap.get(PROPERTY_NAME));
+                    injectionMap.put(val, new String(Base64.encodeBase64(value.getBytes())));
+                }
+            }
+
+        }
+
+        for(String injection: injectionMap.keySet())
+        {
+            updatedJSON = updatedJSON.replace(injection, injectionMap.get(injection));
```
cwoods-cpointe: a: If we are using both they key and value then we can update the for to loop over the entrySet

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -83,18 +88,35 @@ public AdmissionReview validate(AdmissionReview admissionReviewRequest) {
...
             String updatedJson = replaceConfigStoreKeysWithConfigValue(json);
+            //This is only applicable for Secret that scans both encoded values and plain text value and replace corresponding values and saves into updatedSecretJson
+            String updatedSecretJson = "";
 
-            if (!updatedJson.isEmpty()) {
+            if(request.getKind().getKind().equals("Secret")) {
+                updatedSecretJson = replaceSecretConfigStoreKeyWithConfigValue(json, mapper);
+            }
+
+            if (request.getKind().getKind().equals("Secret") && !StringUtils.equals(updatedSecretJson, json)) {
```
epwilkins: Q: Why do we have special handling here? This looks identical to the next else-if block just with a different string variable.  Can we not just use `updatedJson`? I'd think we would first process the plain text updates, and then append any encoded updates to the end of that. If there are neither, `updatedJson` would be empty. If there's either or both, then the logic here for sending the patch response is the same.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -129,9 +151,63 @@ private String getProperty(String groupName,String propertyName) {
...
+        //inject any plain text configuration store values first.
+        String plainJsonInjection = replaceConfigStoreKeysWithConfigValue(json);
+
+        String updatedJSON = null;
+
+        if(!plainJsonInjection.isEmpty())
+        {
+            updatedJSON = plainJsonInjection;
+        }else{
+            updatedJSON = json;
```
epwilkins: C: This is a patch. If we include the entire original JSON, we're saying that the whole thing has to be patched. We should only include fields for which we have changes.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -129,9 +151,63 @@ private String getProperty(String groupName,String propertyName) {
...
+            {
+                String decodedStr = new String(Base64.decodeBase64(val), StandardCharsets.UTF_8);
+
+                if(decodedStr.startsWith(CONFIG_STORE_INJECT_START))
+                {
+                    //trim out unnecessary strings
+                    String[] key = decodedStr.substring(CONFIG_STORE_INJECT_START.length(),decodedStr.length()-1).split(PROPERTY_KEY_DELIMITER);
+                    Map<String, String> keyMap = convertToKeyMap(key);
+                    String value = getProperty(keyMap.get(GROUP_NAME), keyMap.get(PROPERTY_NAME));
+                    injectionMap.put(val, new String(Base64.encodeBase64(value.getBytes())));
```
cwoods-cpointe: a: Can we update the generic key and value names to be more descriptive. Hard to follow with the dataMap val being use for the injectionMap key

epwilkins: I: Here is where you could just append the new base64 encoded value (post injection) to the `updatedJson` string.

foundation/foundation-configuration-store/pom.xml
```diff
@@ -165,6 +165,16 @@
             <artifactId>jakarta.ws.rs-api</artifactId>
             <version>${version.jakarta.wr.rs}</version>
         </dependency>
+        <dependency>
+            <groupId>commons-codec</groupId>
+            <artifactId>commons-codec</artifactId>
```
epwilkins: Q/S: Why do we need to introduce this dependency instead of using Java's [built-in](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Base64.html) `Base64` class?

foundation/foundation-configuration-store/pom.xml
```diff
@@ -165,6 +165,16 @@
...
+            <version>${version.common.codec}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.tdunning</groupId>
+            <artifactId>json</artifactId>
```
epwilkins, line 170: Q/S: Similarly, what functionality are we getting out of this package that we don't get from Jackson or zjsonpatch?

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/values.yaml
```diff
@@ -25,6 +25,7 @@ mysql:
     # Note: Changing these values requires removal of the `hive-metastore-db-0` PVC, or manual modification of the
     # persisted database.
     username: $getConfigValue(groupName=spark-infrastructure;propertyName=metastore.db.username)
+    password: $getConfigValue(groupName=spark-infrastructure;propertyName=metastore.db.password)
```
epwilkins, line 28: C: This needs to be updated in the README and release notes.

foundation/foundation-configuration-store/src/test/java/com/boozallen/aissemble/configuration/mutating/webhook/MutatingWebhookSteps.java
jaebchoi: Agreed, Ill update to reuse the steps!

foundation/foundation-configuration-store/src/test/java/com/boozallen/aissemble/configuration/mutating/webhook/MutatingWebhookSteps.java
jaebchoi: true, I can update to add param and reuse!

foundation/foundation-configuration-store/pom.xml
jaebchoi: [commons-codec](https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/binary/Base64.html#isBase64(java.lang.String)) have isBase64 to differentiate whether some string has valid base64 format or not, I can do regex but this way seems much cleaner

foundation/foundation-configuration-store/pom.xml
jaebchoi: JSON org lib can easily convert to JSON object and [has](https://stleary.github.io/JSON-java/index.html) method can provide whether json string have certain key or not. Not sure if Jackson have same functionality

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
jaebchoi: It seems to be current logic for replaceConfigStoreKeysWithConfigValue is if there's change it would return whole updated json and if no change it would return empty. is that intended?

foundation/foundation-configuration-store/pom.xml
epwilkins: Why do you need to determine whether it's base64 or not? We know any value inside of the `data` map of a `Secret` must be base64 encoded. I think we can assume it is.  If that's somehow violated it would result in an exception and a 500 return, which would be correct.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
epwilkins: I see. The `JsonDiff` class is used to create a patch before attaching to the admission response.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
jaebchoi: agreed, updated to just use updatedJson and removed unnecessary checks

foundation/foundation-configuration-store/pom.xml
jaebchoi: Fair point, updated to remove common-codec and its usage!

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/README.md
```diff
@@ -50,6 +50,7 @@ configuration options.
 | fullnameOverride | "hive-metastore-db" |
 | auth.database    | "metastore"         |
 | auth.username    | "hive"              |
+| auth.password    | "hive"              |
```
epwilkins: S: This is inaccurate. We should put the actual value, and a footnote explaining that the value is being injected from the Universal Configuration store by default. We should link the words "Universal Configuration" to the relevant [tech docs page](https://boozallen.github.io/aissemble/aissemble/current/configuration-store.html).

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -129,9 +142,63 @@ private String getProperty(String groupName,String propertyName) {
...
+     * @param plainTextInjectionUpdated boolean that plain text injection is updated.
+     * @return updatedJSON  If no injection, return empty.
+     */
+    private String replaceSecretConfigStoreKeyWithConfigValue(String json, ObjectMapper mapper, boolean plainTextInjectionUpdated) throws JsonProcessingException {
+        String updatedJSON = json;
+        JSONObject jsonObject = new JSONObject(json);
+        //If there's no data key, no injection can be done.
+        if(!jsonObject.has("data"))
+        {
+            return "";
```
epwilkins: C: This is incorrect if `plainTextInjectionUpdated` is true. Might consider adding a unit test for a Secret that has both plaintext and encoded injection.

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/README.md
jaebchoi: hmm actual value being encoded values? or $getConfigValue(..?

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
jaebchoi: Missed that, good call!

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/README.md
epwilkins: I'm on the fence with what to put in the table. On the one hand `"$getConfigValue(...)"` feels more accurate, but on the other hand it feels a bit arcane if you're not familiar with Universal Config. I think it probably makes sense to put the real value we're setting in `values.yaml` (i.e. `$getConfig`) and just make sure the footnote is explanatory enough.

epwilkins (approved):

PR #531, opened 2025-01-14
#511 fix the generated python step data record import statement

### LINKED ISSUE #511: BUG: When creating a project with both a Java and Python data delivery pipeline, as well as a custom record, the record import statement in the ingest_base.py has the incorrect name for the data records folder.
## Description
Found during regression test script development. When creating a project with both a Java and Python data delivery pipeline, as well as a custom record, the record import statement in the `ingest_base.py` file has the incorrect name for the data records folder.

## Steps to Reproduce
1. Navigate to the regression test script.
2. Follow the steps for Getting Started, Instantiating the Regression Test Project, and Adding Pipelines
3. On the second build utilizing mvn clean install, you will encounter an error:
`[ERROR] src/sync_py_spark_pipeline/generated/step/ingest_base.py:19:0: E0401: Unable to import 'regression_test_data_records.record.py_person' (import-error)`

## Expected Behavior
The build should pass and the proper import name should exist in the `ingest_base.py`.

## Actual Behavior
The build fails.

## Test Steps
Follow the `Steps to Reproduce` and verify you can successfully complete the build

csun-cpointe:


foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/generator/TargetedPipelineStepPythonGenerator.java
```diff
@@ -67,4 +76,55 @@ protected boolean shouldGenerateStep(Step step, GenerationContext generationCont
...
+                Node execution = executions.item(i);
+                if (execution.getNodeType() == Node.ELEMENT_NODE){
+                    configuration = getDataRecordConfiguration(execution);
+                    if (configuration != null) {
+                        break;
+                    }
+                }
+            }
+
+            if (configuration != null && configuration.getTextContent().contains("separate")) {
```
epwilkins: A: I think it would be more straight-forward to check for the module within `modules`. This is probably a sensible check anyway, as it would allow us to verify that the module is actually a part of the project, which is not guaranteed just because of the precense of a fermenter execution.

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/generator/TargetedPipelineStepPythonGenerator.java
```diff
@@ -67,4 +76,55 @@ protected boolean shouldGenerateStep(Step step, GenerationContext generationCont
...
     }
 
+    /**
+     * Get the appropriate data record package from the given context
+     * @param context the generation context
+     * @return data record package name
+     */
+    private String getPythonDataRecordPackage(GenerationContext context) {
+        File rootDir = context.getExecutionRootDirectory();
+        String sharedDir = rootDir.getPath() + File.separator + rootDir.getName() + "-shared";
```
epwilkins: A: Consider using `getSharedModuleName`

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/generator/TargetedPipelineStepPythonGenerator.java
csun-cpointe: Yeah.. that was actually my original approach but given the multiple piplelines and changes of split, there could be more module names for validation so instead, I directly go for the profile name

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/python/PythonStep.java
```diff
@@ -145,9 +145,8 @@ private Set<String> getImports(boolean isImplModule) {
                     importsSet.add(moduleImport.replace("from ",
                             isImplModule ? "from .." : "from ..."));
                 }else{
-                    String rootArtifact = PipelineUtils.deriveLowerSnakeCaseNameFromHyphenatedString(rootArtifactId);
                     importsSet.add(moduleImport.replace("from ",
-                            "from " + rootArtifact + "_data_records."));
+                            "from " + PipelineUtils.deriveLowerSnakeCaseNameFromHyphenatedString(pythonDataRecordPackage) + "."));
```
epwilkins, line 149: C: What if the data records are generated in-line with the pipeline?

epwilkins: Ah never mind. I see we have the if check above this. We only execute this logic if it's not in-line generation.

csun-cpointe: the line approach is already covered by another if block

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/generator/TargetedPipelineStepPythonGenerator.java
```diff
@@ -67,4 +76,55 @@ protected boolean shouldGenerateStep(Step step, GenerationContext generationCont
         return true;
     }
 
+    /**
+     * Get the appropriate data record package from the given context
+     * @param context the generation context
+     * @return data record package name
+     */
+    private String getPythonDataRecordPackage(GenerationContext context) {
```
epwilkins, line 85: Q/A: Could this also make the determination about whether the data records are in-line with the pipeline, so that the `PythonStep` logic could be simplified?

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/generator/TargetedPipelineStepPythonGenerator.java
epwilkins: I'm not sure I understand. Wouldn't you just check for the the `DataRecordModule.CORE` name within modules, then check for `DataRecordModule.COMBINED` if core was not present?  Why does the number of modules in `*-shared` matter?

foundation/foundation-mda/src/test/java/com/boozallen/aiops/mda/generator/SparkApplicationStep.java
```diff
@@ -131,4 +159,99 @@ protected void saveFileStores(List<FileStore> fileStores) throws IOException  {
         this.pipeline.setFileStores(fileStores);
         savePipelineToFile(pipeline);
     }
+
+    private void AddPipeline(boolean isPySparkPipeline) throws Exception {
```
epwilkins: A:
```suggestion
    private void addPipeline(boolean isPySparkPipeline) throws Exception {
```

epwilkins (approved):

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/generator/TargetedPipelineStepPythonGenerator.java
csun-cpointe: Discussed offline. To be more clear, will use the full profile name,`aissemble-data-records-separate-module` to check in case of profile name changes.

PR #537, opened 2025-01-17
[#525] update libraries on Spark images

### LINKED ISSUE #525: Feature: Update libraries on Spark image
## Description
There are a few out of date JARs on the Spark image that would benefit from upgrading. We should patch in the latest versions that are compatible with the existing versions.

## DOD
- [x] Updated libraries are pulled in during the Spark image build
  - [x] Create automated tests to verify that Spark can be started after the updates
- [x] The Spark images function as before

## Test Strategy/Script
1. Create a project from the archetype:
    ```sh
    mvn archetype:generate '-DarchetypeGroupId=com.boozallen.aissemble'\
                           '-DarchetypeArtifactId=foundation-archetype'\
                           '-DarchetypeVersion=1.11.0-SNAPSHOT'\
                           '-DgroupId=org.test'\
                           '-Dpackage=org.test'\
                           '-DprojectGitUrl=test.org/test.git'\
                           '-DprojectName=Test Project'\
                           '-DartifactId=final-525'
    && cd final-525
    ```
2. Add a [Spark](https://github.com/user-attachments/files/18456683/SparkPipeline.json) and [Pyspark](https://github.com/user-attachments/files/18456682/PysparkPipeline.json) pipeline metamodel to _final-525-pipeline-models/src/main/resources/pipelines_
3. Fully generate the project until there are no manual actions
4. **OTS Only**:
    1. Set the **repository** values in the `spark-infrastructure` _Chart.yaml_ to the respective local chart
    2. Set the image tag to 1.11.0-SNAPSHOT for each deployment in `spark-infrastructure`
    3. Turn off the helm chart v2 Baton migration in the root _pom.xml_
    4. Update _final-525-pipelines/pyspark-pipeline/pyproject.toml_ to set `aissemble-extensions-data-delivery-spark` to the local python module
    5. Rebuild the project5. Deploy the project: `tilt up`6. Ensure the _hive-metastore-service_, _spark-thrift-server_, and _spark-history-server_ start successfully
    - _**Note:** the history server will throw an exception about the TAR codec, this is pre-existing_
7. Run the _spark-pipeline_ and ensure it completes successfully
8. Run the _pyspark-pipeline_ and ensure it completes successfully
9. Navigate to the Spark History UI (usually at http://localhost:18080) and ensure the two pipeline runs are registered

epwilkins:
Primarily, this changeset hardens the Spark images by upgrading libraries used by Spark to their highest compatible version. This included upgrading Spark from 3.5.2 to 3.5.4 (the latest version), and modifying the history and Thrift server charts to not hardcode the version of the Ivy jar used to load libraries at deploy time.

This changeset also includes a fix to the Helm version enforcment logic to properly detect/allow for non-standard release version (e.g. release candidate versions).

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -79,7 +79,7 @@
         <version.javax.servlet>3.1.0</version.javax.servlet>
 
         <!-- Spark Default Dependencies.  See `spark-*` profiles below for alternative sets -->
-        <version.spark>3.5.2</version.spark>
+        <version.spark>3.5.4</version.spark>
         <version.scala>2.12.20</version.scala>
         <version.scala.minor>2.12</version.scala.minor>
         <version.delta>3.2.1</version.delta>
@@ -1038,33 +1038,6 @@ To suppress enforce-helm-version rule, you must add following plugin to the root
                                 <addPypiRepoAsPackageSources>false</addPypiRepoAsPackageSources>
                                 <useDevRepository>true</useDevRepository>
                             </configuration>
-                        </plugin>
-                        <plugin>
-                            <groupId>${group.fabric8.plugin}</groupId>
-                            <artifactId>docker-maven-plugin</artifactId>
-                            <version>${version.fabric8.docker.maven.plugin}</version>
-                            <executions>
-                                <execution>
-                                    <id>default-build</id>
-                                    <phase>package</phase>
-                                    <goals>
-                                        <goal>build</goal>
-                                    </goals>
-                                    <configuration>
-                                        <!-- Deploy will build all platforms, so skip build phase in this case: -->
-                                        <buildArchiveOnly>true</buildArchiveOnly>
-                                        <images>
-                                            <image>
-                                                <build>
-                                                    <buildx>
-                                                        <builderName>maven</builderName>
-                                                    </buildx>
-                                                </build>
-                                            </image>
-                                        </images>
-                                    </configuration>
-                                </execution>
-                            </executions>
                         </plugin>
                     </plugins>
                 </pluginManagement>
--- a/build-support/aissemble-enforcer-extension/src/main/java/com/boozallen/aissemble/maven/enforcer/helper/HelmVersionHelper.java
+++ b/build-support/aissemble-enforcer-extension/src/main/java/com/boozallen/aissemble/maven/enforcer/helper/HelmVersionHelper.java
@@ -31,7 +31,7 @@ public class HelmVersionHelper {
     private static final Logger logger = LoggerFactory.getLogger(HelmVersionHelper.class);
 
     private static final String HELM_COMMAND = "helm";
-    private static final String EXTRACT_VERSION_REGEX = "\"v((\\d\\.?)+)\"";
+    private static final String EXTRACT_VERSION_REGEX = "Version:\"v(.*?)\"";
 
     private final File workingDirectory;
 
--- a/extensions/extensions-docker/aissemble-spark-infrastructure/pom.xml
+++ b/extensions/extensions-docker/aissemble-spark-infrastructure/pom.xml
@@ -24,7 +24,7 @@
             <groupId>com.boozallen.aissemble</groupId>
             <artifactId>aissemble-spark</artifactId>
             <version>${project.version}</version>
-            <type>pom</type>
+            <type>docker-build</type>
         </dependency>
         <dependency>
             <groupId>org.apache.hadoop</groupId>
@@ -70,6 +70,64 @@
                         </image>
                     </images>
                 </configuration>
+                <executions>
+                    <execution>
+                        <id>test-image</id>
+                        <phase>integration-test</phase>
+                        <goals>
+                            <goal>start</goal>
+                        </goals>
+                        <configuration>
+                            <showLogs>true</showLogs>
+                            <images>
+                                <image>
+                                    <run>
+                                        <containerNamePattern>${project.artifactId}-test</containerNamePattern>
+                                        <!-- If autoRemove is true and wait is specified, the build will fail. See https://github.com/fabric8io/docker-maven-plugin/issues/1622 -->
+                                        <autoRemove>false</autoRemove>
+                                        <entrypoint>sh -c /tmp/test/image-test.sh</entrypoint>
+                                        <wait>
+                                            <time>60000</time>
+                                            <http>
+                                                <!-- History server check -->
+                                                <url>http://localhost:18080</url>
+                                            </http>
+                                            <!-- Thrift server check -->
+                                            <log>.*HiveThriftServer2 started.*</log>
+                                        </wait>
+                                        <volumes>
+                                            <bind>
+                                                <volume>${project.basedir}/src/test/resources:/tmp/test</volume>
+                                            </bind>
+                                        </volumes>
+                                    </run>
+                                </image>
+                            </images>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>exec-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>test-image-cleanup</id>
+                        <phase>post-integration-test</phase>
+                        <goals>
+                            <goal>exec</goal>
+                        </goals>
+                        <configuration>
+                            <executable>docker</executable>
+                            <arguments>
+                                <argument>rm</argument>
+                                <argument>-f</argument>
+                                <argument>${project.artifactId}-test</argument>
+                                <argument>hive-metastore-db</argument>
+                            </arguments>
+                        </configuration>
+                    </execution>
+                </executions>
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
--- a/extensions/extensions-docker/aissemble-spark-operator/src/main/resources/docker/Dockerfile
+++ b/extensions/extensions-docker/aissemble-spark-operator/src/main/resources/docker/Dockerfile
@@ -18,7 +18,7 @@ ARG VERSION_AISSEMBLE
 
 FROM docker.io/kubeflow/spark-operator:v1beta2-1.6.2-3.5.0 AS builder
 
-# We would be able to use the kubeflow image directly, except that it is on Spark 3.5 instead of 3.4
+# Use our image as a base to ensure Spark version alignment and take advantage of image hardening
 FROM ${DOCKER_BASELINE_REPO_ID}boozallen/aissemble-spark:${VERSION_AISSEMBLE}
 
 LABEL org.opencontainers.image.source="https://github.com/boozallen/aissemble"
--- a/extensions/extensions-docker/aissemble-spark/src/main/resources/docker/Dockerfile
+++ b/extensions/extensions-docker/aissemble-spark/src/main/resources/docker/Dockerfile
@@ -4,13 +4,16 @@ ARG SPARK_VERSION
 ARG SCALA_VERSION
 FROM docker.io/apache/spark:${SPARK_VERSION}-scala${SCALA_VERSION}-java17-python3-ubuntu
 
+#pull var from global scope to build stage
+ARG SPARK_VERSION
+ARG PYTHON_VERSION=3.11
+
 LABEL org.opencontainers.image.source="https://github.com/boozallen/aissemble"
 
 USER root
 
-# Configures the desired version of Python to install
-ARG PYTHON_VERSION=3.11
 # Setup Spark home directory
+RUN usermod -d /opt/spark spark
 RUN mkdir $SPARK_HOME/checkpoint && \
   mkdir $SPARK_HOME/krausening && \
   mkdir $SPARK_HOME/warehouse && \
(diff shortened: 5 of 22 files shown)


build-support/aissemble-enforcer-extension/src/main/java/com/boozallen/aissemble/maven/enforcer/helper/HelmVersionHelper.java
```diff
@@ -31,7 +31,7 @@ public class HelmVersionHelper {
     private static final Logger logger = LoggerFactory.getLogger(HelmVersionHelper.class);
 
     private static final String HELM_COMMAND = "helm";
-    private static final String EXTRACT_VERSION_REGEX = "\"v((\\d\\.?)+)\"";
+    private static final String EXTRACT_VERSION_REGEX = "Version:\"v(.*?)\"";
```
epwilkins, line 34: I: This allows us to capture the version regardless of its format, including `3.17.0-rc1`

extensions/extensions-docker/aissemble-spark-infrastructure/pom.xml
```diff
@@ -70,6 +70,59 @@
                         </image>
                     </images>
                 </configuration>
+                <executions>
+                    <execution>
+                        <id>test-image</id>
+                        <phase>integration-test</phase>
```
epwilkins, line 76: I: I noticed this image isn't actually used in the new spark-infrastructure chart. There is an argument to make that we should just remove it. However, I reasoned that the new chart hasn't been used in projects _enough_ yet to determine if the absence of this image is an oversight or reasonable streamlining.

extensions/extensions-docker/aissemble-spark-operator/src/main/resources/docker/Dockerfile
```diff
@@ -18,7 +18,7 @@ ARG VERSION_AISSEMBLE
 
 FROM docker.io/kubeflow/spark-operator:v1beta2-1.6.2-3.5.0 AS builder
 
-# We would be able to use the kubeflow image directly, except that it is on Spark 3.5 instead of 3.4
+# Use our image as a base to ensure Spark version alignment and take advantage of image hardening
```
epwilkins, line 21: I: The original reasoning is no longer valid, however based on how the kubeflow image is managed, it's still valid to use our image to ensure version alignment. The more recent versions of the kubeflow image drop the spark version from the tag all together.

extensions/extensions-docker/aissemble-spark/src/main/resources/scripts/setup.sh
```diff
@@ -0,0 +1,113 @@
...
+  # Following approach mentioned in https://github.com/pypa/pip/issues/10458
+  echo "\
+from setuptools import setup
+
+setup(
+    name='pyspark',
+    version='$2',
+    description='A dummy package representing the provided PySpark installation',
+)" > "$1/python/setup.py"
+  python3 -m pip install "$1/python"
```
epwilkins, line 102: I: This is a little bit of sleight-of-hand to ensure that when a PySpark pipeline is installed on the image, it recognizes that PySpark is already available and does not duplicatively download/install it.

extensions/extensions-docker/aissemble-spark/src/main/resources/scripts/setup.sh
```diff
@@ -0,0 +1,113 @@
...
+# #%L
+# aiSSEMBLE::Extensions::Docker::Spark
+# %%
+# Copyright (C) 2021 Booz Allen
+# %%
+# This software package is licensed under the Booz Allen Public License. All Rights Reserved.
+# #L%
+###
+
+update_maven_jars() {
```
epwilkins, line 17: A: Document args for each function

extensions/extensions-helm/aissemble-spark-application-chart/tests/deployment_test.yaml
```diff
@@ -11,7 +11,7 @@ tests:
           value: "placeholder"
       - equal:
           path: spec.sparkVersion
-          value: 3.5.2
+          value: 3.5.4
```
epwilkins, line 14: A: Consider if there is a way we can avoid hard-coding this and instead just assert that it is equal to the `version.spark` property from the POM.

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-spark-history-chart/templates/deployment.yaml
```diff
@@ -59,7 +59,8 @@ spec:
           imagePullPolicy: {{ .Values.deployment.image.imagePullPolicy | default "IfNotPresent" }}
           # Uses the Ivy jar packaged with spark to download dependency packages (and their transitive dependencies) to /tmp/jars/...
           # Specifically downloads the binaries-- Does not download supplemental classifiers, ie sources
-          command: ["java", "-Divy.cache.dir=/tmp/", "-Divy.home=/tmp/", "-jar", "/opt/spark/jars/ivy-2.5.1.jar", "-confs", "default", "-retrieve", "/tmp/jars/[artifact].[ext]", {{ include "deps.packages.ivy" . }}]
+          command: ["sh", "-c"]
+          args: ["java -Divy.cache.dir=/tmp/ -Divy.home=/tmp/ -cp '/opt/spark/jars/*' org.apache.ivy.Main -confs default -retrieve /tmp/jars/[artifact].[ext] {{ include "deps.packages.ivy" . }}"]
```
epwilkins, line 63: I: It was necessary to compact this into one string so that the quoted `'/opt/spark/jars*'` argument would be passed correctly. Otherwise it was not processed successfully resulting in a `ClassNotFound` error

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-spark-history-chart/tests/deployment_test.yaml
```diff
@@ -288,38 +288,22 @@ tests:
         - some_jar.jar
         - another_jar.jar
     asserts:
-# See https://github.com/helm-unittest/helm-unittest/issues/342
-#      - contains:
-#          path: spec.template.spec.initContainers[1].command
-#          any: true
```
epwilkins, line 294: I: Left a comment on the GH issue, but `any` isn't required for this use case and due to a (possible) bug actually results in a failure. Omitting it works just fine and would be the intended usage.

build-parent/pom.xml
```diff
@@ -1039,33 +1039,6 @@ To suppress enforce-helm-version rule, you must add following plugin to the root
                                 <useDevRepository>true</useDevRepository>
                             </configuration>
                         </plugin>
-                        <plugin>
-                            <groupId>${group.fabric8.plugin}</groupId>
-                            <artifactId>docker-maven-plugin</artifactId>
-                            <version>${version.fabric8.docker.maven.plugin}</version>
```
epwilkins, line 1045: I: This is likely the most controversial change.  We did this so that the native image was not built twice, once during `install` and again during `deploy`.  This was a nice optimization that cut our build times down significantly when it was first introduced.  However, it precludes automated testing of the image as was introduced in this commit.  Further, I _believe_ that with some of our other optimizations and build stability improvements the build time lengthening again should not be terribly intrusive.  Even if this were to double our build time, the build would still be finished by 5 am ET.

extensions/extensions-docker/aissemble-spark-infrastructure/pom.xml
```diff
@@ -24,7 +24,7 @@
             <groupId>com.boozallen.aissemble</groupId>
             <artifactId>aissemble-spark</artifactId>
             <version>${project.version}</version>
-            <type>pom</type>
+            <type>docker-build</type>
```
csun-cpointe, line 27: Q: is this change made to match the aissemble-spark/pom.xml file packaging definition

extensions/extensions-docker/aissemble-spark/src/main/resources/docker/Dockerfile
```diff
@@ -28,23 +31,31 @@ RUN apt-get update -y && apt-get install --assume-yes \
...
+  && ln -s /usr/bin/python${PYTHON_VERSION} /usr/bin/python \
+  && python -m pip install --upgrade cryptography
 
 # Pyspark uses `python3` to execute pyspark pipelines. This links out latest python install to that command.
 RUN ln -sf /usr/bin/python3.11 /usr/bin/python3
 
+# Update gosu from Ubuntu 24 distribution (compatible with but not available for Ubuntu 22)
+RUN curl -SsLO https://launchpad.net/ubuntu/+archive/primary/+files/gosu_1.17-1ubuntu0.24.04.2_amd64.deb \
+  && apt-get install ./gosu_1.17-1ubuntu0.24.04.2_amd64.deb \
+  && rm gosu_1.17-1ubuntu0.24.04.2_amd64.deb
```
csun-cpointe: Q: is these amd64 specific? will it support arm64 env?

extensions/extensions-docker/aissemble-spark/src/main/resources/docker/Dockerfile
```diff
@@ -28,23 +31,31 @@ RUN apt-get update -y && apt-get install --assume-yes \
   curl \
   python${PYTHON_VERSION} \
   python${PYTHON_VERSION}-dev \
-#TODO is distutils needed?
   python${PYTHON_VERSION}-distutils \
-#Patch for CVE-2023-4863: upgrade libwebp7 to latest
+#Patch for CVE-2023-4863: upgrade libwebp7 to 1.3.2+
```
csun-cpointe, line 35: Q: is the latest version 1.3.2+? it doesn't seem we specify 1.3.2+ specifically

csun-cpointe (approved):

build-parent/pom.xml
carter-cundiff: Since we only have tests for 2 images at this time, could we just configure those images to have `<buildArchiveOnly>false</buildArchiveOnly>`  while keeping the others on the existing functionality?

extensions/extensions-docker/aissemble-spark/src/main/resources/docker/Dockerfile
epwilkins: I only added the version number in to help with recognizing when we could remove this patch. The version that address the CVE is 1.3.2, and once the base image we're including has that version or higher, we can remove this fix.  Without the number I had to go track it down to see if it was still necessary.

extensions/extensions-docker/aissemble-spark/src/main/resources/docker/Dockerfile
epwilkins: Good point. I'll make this dynamic based on architecture.

extensions/extensions-docker/aissemble-spark-infrastructure/pom.xml
epwilkins: This fixes a build cache issue. When we use type `pom`, the build cache will only rebuild this module if there are changes to the actual POM of `aissemble-spark`.  By more honestly defining the dependency in this way, a rebuild is triggered if _anything_ is changed in `aissemble-spark`.

csun-cpointe: ah.. right.. the calculation of cache checksum ignores the dependency with pom type. Great fix. Thanks.

extensions/extensions-docker/aissemble-spark-infrastructure/pom.xml
```diff
@@ -70,6 +70,59 @@
                         </image>
                     </images>
                 </configuration>
+                <executions>
+                    <execution>
+                        <id>test-image</id>
+                        <phase>integration-test</phase>
+                        <goals>
+                            <goal>start</goal>
```
carter-cundiff, line 78: Very cool, didn't know fabric8 had this goal!

extensions/extensions-docker/aissemble-spark-infrastructure/pom.xml
```diff
@@ -70,6 +70,59 @@
...
+                        <configuration>
+                            <showLogs>true</showLogs>
+                            <images>
+                                <image>
+                                    <run>
+                                        <containerNamePattern>${project.artifactId}-test</containerNamePattern>
+                                        <!-- If autoRemove is true and wait is specified, the build will fail. See https://github.com/fabric8io/docker-maven-plugin/issues/1622 -->
+                                        <autoRemove>false</autoRemove>
+                                        <entrypoint>/opt/spark/test/image-test.sh</entrypoint>
+                                        <wait>
```
carter-cundiff, line 89: Q: So is the idea that all these wait conditions have to be met and that considered a passing docker image?

extensions/extensions-docker/aissemble-spark-infrastructure/pom.xml
```diff
@@ -70,6 +70,59 @@
...
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>exec-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>test-image-cleanup</id>
```
carter-cundiff, line 115: Q: Any reason we can't use the fabric8 stop goal here instead?

extensions/extensions-docker/aissemble-spark-infrastructure/src/main/resources/docker/Dockerfile
```diff
@@ -10,6 +10,8 @@ RUN curl -L https://github.com/delta-io/connectors/releases/download/v${DELTA_HI
 ARG JARS_DIR
 ADD ${JARS_DIR}/* $SPARK_HOME/jars/
 
+COPY --chmod=755 src/test/resources/image-test.sh $SPARK_HOME/test/image-test.sh
```
carter-cundiff: This might be overkill so feel free to ignore. Since we're building this image twice (once in build and once in deploy), I think it would be possible to conditionally copy this test script over only in the build phase. Then in the deploy phase we don't copy it and it's not in our final image. See [SO post](https://stackoverflow.com/a/54245466) on conditionally copying using build args.

extensions/extensions-docker/aissemble-spark/src/main/resources/docker/Dockerfile
```diff
@@ -28,23 +31,29 @@ RUN apt-get update -y && apt-get install --assume-yes \
...
   && apt-get upgrade -y libwebp7 \
   && rm -rf /var/lib/apt/lists/* \
   && apt-get clean \
-  && ln -s /usr/bin/python${PYTHON_VERSION} /usr/bin/python
+  && ln -s /usr/bin/python${PYTHON_VERSION} /usr/bin/python \
+  && python -m pip install --upgrade cryptography
```
carter-cundiff, line 41: Q: Is this a CVE patch? Might be helpful to leave a comment

extensions/extensions-docker/aissemble-spark-infrastructure/pom.xml
epwilkins: Correct

extensions/extensions-docker/aissemble-spark-infrastructure/pom.xml
epwilkins: We could use `stop`, however the key concern is to remove the container, not simply stop it.  Unfortunately there is a bug with `autoRemove` and wait conditions, so we can't just use that. Further, the `remove` goal (as far as I can tell) only removes images and not containers.

extensions/extensions-docker/aissemble-spark-infrastructure/src/main/resources/docker/Dockerfile
epwilkins: I like the thought, but I'm thinking [volume mounts](https://dmp.fabric8.io/#start-volumes) would be a bit more straightforward.

extensions/extensions-docker/aissemble-spark/src/main/resources/docker/Dockerfile
epwilkins: Good idea

extensions/extensions-docker/aissemble-spark/src/main/resources/docker/Dockerfile
```diff
@@ -28,23 +31,29 @@ RUN apt-get update -y && apt-get install --assume-yes \
...
 
 # Pyspark uses `python3` to execute pyspark pipelines. This links out latest python install to that command.
 RUN ln -sf /usr/bin/python3.11 /usr/bin/python3
 
+# Update gosu (based on https://github.com/tianon/gosu/blob/master/INSTALL.md)
+COPY --from=docker.io/tianon/gosu:1.17-debian /gosu /usr/sbin/gosu
```
carter-cundiff, line 48: A: Comment for why we're updating gosu would be helpful

extensions/extensions-docker/aissemble-spark-infrastructure/pom.xml
carter-cundiff: Looks like the stop goal should also remove the container. It has a `keepContainer` config option which defaults to false.

carter-cundiff (approved):

build-parent/pom.xml
epwilkins: I wanted to circle back to document the real impact of this change. After 4 full nightly builds, it looks like this change has added about 6 minutes to the total build time (going from ~2h3m to ~2h9m)

Issue #533, opened 2025-01-15
Feature: Support Configuration Store values to use properties to inject values


jaebchoi:
## Description
The way we configured Configstore properties file seems to be more specific to one application and specifying volumePathOnNode for config store may be complicated on different machines (especially WSL).

We need additional ways to be more suitable across all machines and be able to abstract out properties so that we can reuse properties across different applications (not just spark-infrastructure). 

## Definition Of Done
- [x] Rename spark-infrastructure.properties to spark.properties
- [x] Support of updating configuration-store values file to add baseproperties with filename associated with properties key/values pair
```
  name: configuration-store-config
  baseConfigFiles:
    spark.properties: |-
      metastore.db.username=hive
      metastore.db.password=hive
  envConfigFiles:
    spark.properties: |-
      metastore.db.username=hive
      metastore.db.password=hive

```
- [x] update to support Configuration store logic to use properties pattern in addition to configurationVolume
- [x] Update README and antora doc of configuration Store based on this change.
- [x] Make sure configuration store can still work in local ArgoCD and can inject configurations correctly.

## Test Strategy/Script

- Pull `test533-with-deploy-script` branch (This includes  deploy script in feature/path2production-alignment for easily testing since this feature uses mutation webhook )
- Run mvn clean install

- Create a downstream project with baseline version 1.11.0-SNAPSHOT
```
mvn archetype:generate -U -DarchetypeGroupId=com.boozallen.aissemble \
  -DarchetypeArtifactId=foundation-archetype \
  -DarchetypeVersion=1.11.0-SNAPSHOT \
  -DgroupId=com.test \
  -DartifactId=test-533 \
  -DprojectGitUrl=https://github.com/{your-github-username}/test-533 \
  -DprojectName=test-533 \
  && cd test-533 
```

Replace {your-github-username} with your github username

- Add the [SparkPipeline.json](https://github.com/user-attachments/files/16781488/SparkPipeline.json) file to the test-533-pipeline-models/src/main/resources/pipelines directory
- Run mvn clean install until all the manual actions are complete
- Once the manual actions are complete, run `mvn clean install -Dmaven.build.cache.skipCache=true` once to get any remaining manual actions
- Modify `-deploy/src/main/resources/apps/spark-infrastructure/Charts.yaml' update following
```
  - name: aissemble-hive-metastore-service-chart
    version: 1.11.0-SNAPSHOT
    repository: oci://ghcr.io/jaebchoi
```
[...]

PR #541, opened 2025-01-22
#533 Feature: Support Configuration Store values to use properties to inject values 

jaebchoi:
Issue: https://github.com/boozallen/aissemble/issues/533


docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -54,6 +54,22 @@ auth-token=auth-secret
 
 NOTE: The `topic` property is overwritten when the environment specific value `ciTopic`.
 
+The Configuration Store also support simplified way to provide config file instead of creating additional .properties file.
```
carter-cundiff:
```suggestion
The Configuration Store also support simplified way to provide configurations instead of mounting properties files.
```

carter-cundiff: This should be moved to the bottom of the "Mounting Properties Files" section of this doc

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -54,6 +54,22 @@ auth-token=auth-secret
...
+----
+configMap:
+  name: configuration-store-quarkus-config
+  baseSparkProperties:
+    - connector=smallrye-kafka
+    - topic=baseTopic
+  envSparkProperties:
+    - connector=smallrye-kafka
+    - auth-token=auth-secret
+----
```
carter-cundiff: This should be moved to the helm chart README. Could create a new section title "Defining Properties Files"

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -54,6 +54,22 @@ auth-token=auth-secret
 
 NOTE: The `topic` property is overwritten when the environment specific value `ciTopic`.
 
+The Configuration Store also support simplified way to provide config file instead of creating additional .properties file.
+
+Under values.yaml, specifying base and environment specific properties can act as configuration properties file.
```
carter-cundiff:
```suggestion
Within the Helm chart, one can directly define properties files with their included configurations.
```

Then we should hyperlink part of the above text "define properties files" to the new section in the helm chart readme.

extensions/extensions-helm/aissemble-configuration-store-chart/values.yaml
```diff
@@ -29,13 +29,26 @@ aissemble-quarkus-chart:
...
+        configMap:
+          name: configuration-store-quarkus-config
+    baseVolumeMounts:
+      - name: config
+        mountPath: "/configurations/base/spark.properties"
+        subPath: spark-base-properties
+    envVolumeMounts:
+      - name: config
+        mountPath: "/configurations/env/spark.properties"
+        subPath: spark-env-properties
```
carter-cundiff: C: The way this is designed the user can only overwrite `spark.properties` from the values file. We want them to be able to define any properties files from the values.yaml. So something like this should be valid:
```
base:
  custom1.properties: |-
       property: value 
  custom2.properties: |-
       property: value 
env:
  custom1.properties: |-
       property: value 
```

foundation/foundation-mda/src/main/resources/targets.json
```diff
@@ -2398,17 +2398,17 @@
...
     "generator": "com.boozallen.aiops.mda.generator.ModelAgnosticResourcesGenerator",
     "metadataContext": "targeted",
     "overwritable": false
   },
   {
-    "name": "envSparkInfrastructureProperties",
-    "templateName": "templates/deployment/spark-infrastructure/configurations/env/spark-infrastructure.properties.vm",
-    "outputFile": "configurations/env/spark-infrastructure.properties",
+    "name": "envSparkProperties",
+    "templateName": "templates/deployment/spark-infrastructure/configurations/env/spark.properties.vm",
```
carter-cundiff, line 2410: Don't think we need this file at all anymore since the new way will be for the user to specify their configs through the helm chart values.yaml.

extensions/extensions-helm/aissemble-quarkus-chart/templates/configmap.yaml
```diff
@@ -7,4 +7,16 @@ data:
     {{- range $property := .Values.configMap.supplementalQuarkusConfig }}
       {{ $property }}
     {{- end }}
-    {{- template "toggleable.configuration.configMap" . }}
\ No newline at end of file
+    {{- template "toggleable.configuration.configMap" . }}
+  {{ if .Values.configMap.baseSparkProperties }}
```
epwilkins: S: I think we should split this into its own config map instead of just appending it to the Quarkus config.

extensions/extensions-helm/aissemble-configuration-store-chart/values.yaml
epwilkins: I'd also point out that these types of connected properties should be mostly handled in the templates instead of the values file. This allows them to be more easily managed by users.

epwilkins:
Q/C: I thought we talked about merging this to dev?

extensions/extensions-helm/aissemble-configuration-store-chart/values.yaml
jaebchoi: I see, then it seems like we can create separate deployment.yaml instead of manipulating quarkus chart correct?

extensions/extensions-helm/aissemble-quarkus-chart/templates/configmap.yaml
jaebchoi: sounds good!

jaebchoi:
> Q/C: I thought we talked about merging this to dev?

Ah yes,, I'll make changes and change it to dev! (might need make new PR)

extensions/extensions-helm/aissemble-configuration-store-chart/values.yaml
carter-cundiff: Probably makes more sense to call it `configmap.yaml`, but yes a separate one within the configuration store helm chart templates.

Oh I see, you need to change the functionality of the `deployment.yaml`, let me think about it.

carter-cundiff: I think you'll want to make the quarkus chart `deployment.yaml` configurable so it can be disabled. So something at the top of the template that defaults to true, but when `deployment.enabled: false` is set then the template does not get generated out. 

Then in the config store helm chart, you can set `aissemble-quarkus-chart.deployment.enabled: false` and use you own custom `deployment.yaml`

jaebchoi: Hmm I see, if we just write own custom deployment.yaml it doesn't get overridden?

carter-cundiff: You can try, but I think you'll end up with duplicate deployment objects in your cluster

epwilkins: I'm not convinced this is the right answer, but one option to also explore would be to make the deployment.yaml file in Quarkus use the [template rendering function](https://pkg.go.dev/text/template#hdr-Nested_template_definitions). That way the values.yaml file could be itself template code that is interpreted. I haven't looked at this specific case to see if it would pan out, but it's something to think about.

epwilkins:
> Ah yes,, I'll make changes and change it to dev! (might need make new PR)

If you click _**Edit**_ next to the PR title, you can change the merge target.

extensions/extensions-helm/aissemble-configuration-store-chart/values.yaml
jaebchoi: Tried custom deployment.yaml approach, seems doable but messy in terms of values.file  since you would take out aissemble-quarkus-chart: key for deployment value and we need another boolean for turning on/off aissemble-quarkus-chart

quarkus have _helpers.tpl where you can define specific template code, they make use of this for togglable config, I think I'll take a shot at this approach

jaebchoi: UPDATE: per meeting with Emily, we decided to split up baseConfigMap and envConfigMap and use configMap name to mount whole file into designated directory.

foundation/foundation-mda/src/main/resources/targets.json
jaebchoi: correct me if i am wrrong we still wanna support this functionality, https://bah-ssp.slack.com/archives/C065RBGCT18/p1736968402284669

jaebchoi: oh we keep  volumePathOnNode function I think for now good to remove, I think

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/values.yaml
```diff
@@ -72,4 +79,7 @@ configMap:
       - name: javax.jdo.option.ConnectionURL
         value: jdbc:mysql://hive-metastore-db:3306/metastore?createDatabaseIfNotExist=true&amp;allowPublicKeyRetrieval=true&amp;useSSL=false
         description: JDBC connect string for a JDBC metastore
+      - name: javax.jdo.option.ConnectionUserName
+        value: $getConfigValue(groupName=spark;propertyName=metastore.db.username)
```
epwilkins, line 83: C: We can't merge default usage of config store to dev until we better flesh out path to production alignment

extensions/extensions-helm/aissemble-configuration-store-chart/README.md
```diff
@@ -86,3 +100,25 @@ When disabling the webhook, ensure the `configMap.supplementalQuarkusConfig`, `d
...
 
 \* When creation of the service account/certificate secret is disabled, the name must be overridden with an existing resource on the cluster to be there replacement. The secret should include the following keys with their respective base64 encoded values: `ca.crt`, `tls.crt`, `tls.key`. Note the `ca.crt` is the Certificate Authority used to validating the webhook's server certificate represented by `tls.crt` and `tls.key`. Additionally, the service account must have at the minimum permissions to create and destroy both `jobs` and `mutatingwebhookconfigurations` kubernetes resources.
+
+## Defining Properties Files
+
+The Configuration Store support simplified way to provide configurations instead of mounting properties files.
+The user can define its configuration properties within values.yaml file as below:
+
+```yaml
+baseconfigMap:
```
epwilkins, line 110: S: I suggested we re-org this a bit and make it easier to turn on/off simple configuration.  Maybe something like:
```yaml
configMapProperties:
  enable: true
  baseFiles:
    spark.properties: |-
      connector=smallrye-kafka
      topic=baseTopic
  envFiles:
    spark.properties: |-
      topic=envTopic
```

extensions/extensions-helm/aissemble-configuration-store-chart/templates/baseConfigmap.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
```
epwilkins, line 1: S: With the proposal above, we would then surround this entire template in `{{- if .Values.configMapProperties.enabled -}}`

extensions/extensions-helm/aissemble-quarkus-chart/templates/deployment.yaml
```diff
@@ -24,6 +24,14 @@ spec:
         {{- if .Values.deployment.supplementalVolumes }}
           {{- toYaml .Values.deployment.supplementalVolumes | nindent 8}}
         {{- end }}
+        {{- if eq .Values.configMap.name "configuration-store-quarkus-config" }}
```
epwilkins, line 27: C/Q: I don't understand this logic. Why would the attachment of the config store properties config maps be reliant on the name of the quarkus config map?

foundation/foundation-mda/src/main/resources/targets.json
carter-cundiff: Yes we want to still support it. However we should utilize the new simplified functionality as our default, which means we no longed need to include this old way.

So a user could choose to use the old way and it should work, but out of the box we'll be using the new way.

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/values.yaml
jaebchoi: True, Ill get it fixed with cherry pick commit.

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -71,6 +71,11 @@ secondary location for properties files that house the environment specific conf
 
 NOTE: Multiple configuration properties files can be stored at either location.
 
+The Configuration Store also support simplified way to provide configurations instead of mounting properties files.
```
epwilkins, line 74: A: I think we need to make the choices here more obvious. It might make sense use separate sections. Maybe something like:
```adoc
==== Properties Files from Config Maps
The simplest way to mount properties files onto the Configuration Store deployment 
is to use ConfigMaps.  To support this case, the configuration store Helm chart 
supports [directly specifying base and environment properties] under values.yaml.

==== Properties Files from Volumes
_existing content_
```

extensions/extensions-helm/aissemble-quarkus-chart/templates/deployment.yaml
jaebchoi: I needed some sort of check to make sure this only applies with configstore service since quarkus chart is also used in other services

jaebchoi: I can use app name to make sure this is only for configuration store instead of configmap name, seems better choice

  app:
    name: aissemble-configuration-store

extensions/extensions-helm/aissemble-configuration-store-chart/README.md
```diff
@@ -20,20 +20,34 @@ helm install configuration-store oci://ghcr.io/boozallen/aissemble-configuration
...
+
+## Configuration Store Properties
+The following properties are used injected in configMap to mount configuration properties file in the downstream project.
+
+| Property                                     | Description                                              | Required Override | Default                         |
+|----------------------------------------------|----------------------------------------------------------|-------------------|---------------------------------|
+| baseconfigMap.name                           | The name of base Configuration properties configMap      | No                | base-configuration-store-config |
+| baseconfigMap.configFiles                    | The list of base config properties files to mount        | No                | spark.properties:               |
+| envconfigMap.name                            | The name of base Configuration properties configMap      | No                | env-configuration-store-config  |
+| envconfigMap.configFiles                     | The list of environment config properties files to mount | No*               | spark.properties:               |
```
epwilkins, line 49: C: Add unit tests for these options, and ensure they can be turned off and direct volume mounting used instead.

docs/modules/ROOT/pages/configuration-store.adoc
carter-cundiff, line 74:
```suggestion
The Configuration Store also supports a simplified way of providing configurations without mounting properties files.
```

docs/modules/ROOT/pages/configuration-store.adoc
```diff
@@ -71,6 +71,11 @@ secondary location for properties files that house the environment specific conf
...
+
+Under values.yaml, https://github.com/boozallen/aissemble/blob/dev/extensions/extensions-helm/aissemble-configuration-store-chart/README.md[specifying base and environment specific properties,role=external,window=_blank] can act as configuration properties file
```
carter-cundiff, line 76: Merge these two sentences into one:

The Configuration Store also supports a simplified way of providing configurations without mounting properties files by https://github.com/boozallen/aissemble/blob/dev/extensions/extensions-helm/aissemble-configuration-store-chart/README.md[defining properties files,role=external,window=_blank] as part of the Helm chart's values.

docs/modules/ROOT/pages/configuration-store.adoc
carter-cundiff: Building on this:

```adoc
==== Properties Files from Config Maps
The simplest way to mount properties files onto the Configuration Store deployment 
is to use ConfigMaps.  To support this case, one can provide configurations by [defining properties files]
as part of the Helm chart's values.
```

Like to avoid using `values.yaml` cause it could also be there `values-dev.yaml` or `values-ci.yaml` so best to keep it generic.

carter-cundiff: Also when changing the header names, make sure to update the hyperlinks in the above `Deploying the Configuration Store` section

extensions/extensions-helm/aissemble-configuration-store-chart/README.md
```diff
@@ -20,20 +20,34 @@ helm install configuration-store oci://ghcr.io/boozallen/aissemble-configuration
...
+| configStore.baseVolumeMounts                                | The ConfigStore base volume mounts for the pod                                                                                                                                                                                                                               | No                | `/etc/webhook/cert=certs`                                                                                                                                                                                                       |
+
+## Configuration Store Properties
```
carter-cundiff, line 41: We need to reorganize this README a little bit. We now have three sections that all have different headers but could contain `Configuration Store Properties`.

Thinking we do this structure:
```
## Configuration Store Properties
### Persistent Volume Properties
#### Defining File Locations

### Config Map Properties
#### Defining Properties Files

### Mutating Webhook Properties
```

where `### Environment Variables` -> `#### Defining File Locations` and `## Custom Properties` -> `### Mutating Webhook Properties`

carter-cundiff: The [environment variables](https://github.com/boozallen/aissemble/blob/dev/extensions/extensions-helm/aissemble-configuration-store-chart/README.md#environment-variables) hyperlink in the config store adoc that will need to be updated as well

extensions/extensions-helm/aissemble-quarkus-chart/templates/deployment.yaml
carter-cundiff: This would still result in the template breaking if a downstream project changes the name `aissemble-configuration-store`

jaebchoi: hmm true,, chart name can also be changed.. I wonder what can make check to differentiate configstore

foundation/foundation-mda/src/main/resources/templates/deployment/spark-infrastructure/v2/spark.infrastructure.values.yaml.vm
```diff
@@ -39,17 +39,15 @@ aissemble-hive-metastore-service-chart:
     auth:
       # Note: Changing these values requires removal of the `hive-metastore-db-0` PVC, or manual modification of the
       # persisted database.
-      username: hive
       rootPassword: hive
       replicationPassword: hive
-      password: hive
 
   configMap:
+    metadata:
```
carter-cundiff, line 46: Also gonna need a baton migration for adding this. Can probably include it as part of the existing `SparkInfrastructureUniversalConfigYAMLMigration` since it's the same file and conditions.

carter-cundiff: Or is this no longer needed? There's also a `configMap.configStore` attribute that's been added.

jaebchoi: sorry @carter-cundiff  actual PR should be [here](https://github.com/boozallen/aissemble/pull/545) as we don't want feature branch merged. closing

Issue #547, opened 2025-01-28
Feature: SLACK: Port secret configuration store injection support


jaebchoi:
## Description
Currently, we have #497 secret configuration store injection work committed in [feature branch](https://github.com/boozallen/aissemble/commit/1cb143c7be9402f4e8091d02db356b9ded8ac009).
This universal Config integration feature work can't be merged to dev until Path to Production work is aligned.
In order to minimize amount of commit of feature branch, we can port this work into dev as this is safe to do so today.

## DOD
Acceptance criteria required to realize the requested feature
- [x] Support mutation webhook to make sure it can inject secret values correctly, but don't turn on configstore and inject secret values yet
- [x] Make sure there's no regression in local ArgoCD deployment.

## Test Strategy/Script

- Pull test547-with-deploy-script branch (This includes deploy script in feature/path2production-alignment for easily testing since this feature uses mutation webhook )

- Run mvn clean install

- Create a downstream project with baseline version 1.11.0-SNAPSHOT

```
mvn archetype:generate -U -DarchetypeGroupId=com.boozallen.aissemble \
  -DarchetypeArtifactId=foundation-archetype \
  -DarchetypeVersion=1.11.0-SNAPSHOT \
  -DgroupId=com.test \
  -DartifactId=test-test-547 \
  -DprojectGitUrl=https://github.com/{your-github-username}/test-547 \
  -DprojectName=test-547 \
  && cd test-547 
```
** Replace {your-github-username} with your github username

- Add the [SparkPipeline.json](https://github.com/user-attachments/files/16781488/SparkPipeline.json) file to the test-547-pipeline-models/src/main/resources/pipelines directory
- Run mvn clean install until all the manual actions are complete
- Once the manual actions are complete, run `mvn clean install -Dmaven.build.cache.skipCache=true `once to get any remaining manual actions

- Create a repo for the project created

- create a new test-547 repository from the github website

- At the test-547 root directory run below commands
```
git init 
git add .
git commit -m "init test-547 base"
git branch -M main
git remote add origin https://github.com/<username>/<repo_name>.git
git push -u origin main
```
- on the root of your downstream project,. make deploy script executable permission
`chmod +x deploy.sh `

- change port fowarding of argocd-server to 30080
i.e. go to Rancher desktop-> Click Port fowarding on left section
NOTE: we’re port forwarding to avoid bug with NodePort in RD on ARM Mac

- Run following to stand up argoCD
`./deploy.sh up`

- Go to localhost:30080

[...]

PR #550, opened 2025-01-29
#547 SLACK: Port secret configuration store injection support 

jaebchoi:
Issue : https://github.com/boozallen/aissemble/issues/547

### DIFF
--- a/foundation/foundation-configuration-store/pom.xml
+++ b/foundation/foundation-configuration-store/pom.xml
@@ -165,6 +165,11 @@
             <artifactId>jakarta.ws.rs-api</artifactId>
             <version>${version.jakarta.wr.rs}</version>
         </dependency>
+        <dependency>
+            <groupId>com.tdunning</groupId>
+            <artifactId>json</artifactId>
+            <version>1.8</version>
+        </dependency>
 
         <!-- quarkus dependencies -->
         <dependency>
--- a/foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
+++ b/foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
@@ -9,17 +9,20 @@ package com.boozallen.aissemble.configuration.mutating.webhook;
  * This software package is licensed under the Booz Allen Public License. All Rights Reserved.
  * #L%
  */
- 
+
 import com.boozallen.aissemble.configuration.store.ConfigLoader;
 import com.boozallen.aissemble.configuration.store.Property;
 import com.boozallen.aissemble.configuration.store.PropertyKey;
 import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.json.JsonReadFeature;
+import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.flipkart.zjsonpatch.JsonDiff;
 import io.fabric8.kubernetes.api.model.StatusBuilder;
 import org.apache.commons.lang3.StringUtils;
+import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -36,6 +39,7 @@ import jakarta.ws.rs.Produces;
 import jakarta.ws.rs.POST;
 import jakarta.ws.rs.Path;
 import jakarta.ws.rs.core.MediaType;
+
 import java.util.Base64;
 import java.util.HashMap;
 import java.util.Map;
@@ -58,7 +62,7 @@ public class ConfigMutatingWebhook {
 
     @Inject
     public ConfigLoader configLoader;
-    
+
     @POST
     @Path("/process")
     @Consumes(MediaType.APPLICATION_JSON)
@@ -83,9 +87,19 @@ public class ConfigMutatingWebhook {
         AdmissionResponse response;
         try {
             String json = mapper.writeValueAsString(request.getObject());
+
+            //This scans to see if there's any config need to be injected and replace corresponding values and saves into updatedJson
             String updatedJson = replaceConfigStoreKeysWithConfigValue(json);
 
-            if (!updatedJson.isEmpty()) {
+            JSONObject jsonObject = new JSONObject(json);
+            if(request.getKind().getKind().equals("Secret") && jsonObject.has("data") ) {
+                //if replaceConfigStoreKeysWithConfigValue doesn't have any injection it would return empty, so we need original json to scan.
+                boolean hasPlainTextInjection = !updatedJson.isEmpty();
+                String secretJson = hasPlainTextInjection ? updatedJson: json;
+                updatedJson = replaceSecretConfigStoreKeyWithConfigValue(secretJson, mapper, hasPlainTextInjection);
+            }
+
+            if (!updatedJson.isEmpty()){
                 JsonNode patch = JsonDiff.asJson(mapper.readTree(json), mapper.readTree(updatedJson));
                 response = new AdmissionResponseBuilder()
                         .withAllowed(true)
@@ -94,7 +108,7 @@ public class ConfigMutatingWebhook {
                         .withPatchType("JSONPatch")
                         .build();
                 logger.info("The kubernetes resource request has been updated.");
-            } else {
+            }else {
                 response = new AdmissionResponseBuilder()
                         .withAllowed(true)
                         .withUid(request.getUid())
@@ -129,9 +143,51 @@ public class ConfigMutatingWebhook {
         }
     }
 
+
+    /**
+     * This method is to inject values in the data fields for any encoded Configvalues from the properties file and save it as based64 encoded.
+     *
+     * @param json json string to scan
+     * @param mapper json object mapper
+     * @param hasPlainTextInjection boolean that plain text injection is updated.
+     * @return updatedJSON  If no injection, return empty.
+     */
+    private String replaceSecretConfigStoreKeyWithConfigValue(String json, ObjectMapper mapper, boolean hasPlainTextInjection) throws JsonProcessingException {
+        String updatedJSON = json;
+        JSONObject jsonObject = new JSONObject(json);
+        String dataString = jsonObject.getString("data");
+        Map<String, String> secretDataMap = mapper.readValue(dataString, new TypeReference<>() {});
+
+        //key being base64 value that contains getConfigValue(), value being base64 value injected with the actual property.
+        Map<String, String> injectionMap = new HashMap<>();
+        for(String secretDataVal : secretDataMap.values()) {
+            String decodedStr = new String(Base64.getDecoder().decode(secretDataVal));
+            String updatedDecodedStr = replaceConfigStoreKeysWithConfigValue(decodedStr);
+            if (!updatedDecodedStr.isEmpty()) {
+                injectionMap.put(secretDataVal, new String(Base64.getEncoder().encode(updatedDecodedStr.getBytes())));
+            }
+        }
+
+        for(Map.Entry<String, String> injection: injectionMap.entrySet()) {
+            updatedJSON = updatedJSON.replace(injection.getKey(), injection.getValue());
+        }
+
+        if (updatedJSON.equals(json)) {
+            if (hasPlainTextInjection) {
+                return json; // no change in secret text, return the json with plain text injections
+            } else {
+                return "";  // no change in both secret and plain text, return empty
+            }
+        } else {
+            return updatedJSON;
+        }
+
+    }
+
     private String replaceConfigStoreKeysWithConfigValue(String json) {
         StringBuilder updatedJson = new StringBuilder();
         String[] data = json.split(Pattern.quote(CONFIG_STORE_INJECT_START));
+
         // if the config store key is found
         if (data.length > 1) {
             updatedJson.append(data[0]);
@@ -161,5 +217,4 @@ public class ConfigMutatingWebhook {
 
         return map;
     }
- }
- 
+}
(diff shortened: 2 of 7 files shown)


foundation/foundation-configuration-store/pom.xml
```diff
@@ -165,6 +165,11 @@
             <artifactId>jakarta.ws.rs-api</artifactId>
             <version>${version.jakarta.wr.rs}</version>
         </dependency>
+        <dependency>
+            <groupId>com.tdunning</groupId>
+            <artifactId>json</artifactId>
+            <version>1.8</version>
+        </dependency>
```
carter-cundiff, line 172: Q: Is this dependency needed? Don't see any imports from it in the java code.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -129,9 +142,58 @@ private String getProperty(String groupName,String propertyName) {
...
+     * @param json json string to scan
+     * @param mapper json object mapper
+     * @param plainTextInjectionUpdated boolean that plain text injection is updated.
+     * @return updatedJSON  If no injection, return empty.
+     */
+    private String replaceSecretConfigStoreKeyWithConfigValue(String json, ObjectMapper mapper, boolean plainTextInjectionUpdated) throws JsonProcessingException {
+        String updatedJSON = json;
+        JSONObject jsonObject = new JSONObject(json);
+        //If there's no data key, no injection can be done.
+        if(!jsonObject.has("data")) {
```
carter-cundiff: Think we could move this logic to up before the method call. So something like this:
```
if (request.getKind().getKind().equals("Secret") && jsonObject.has("data")) {
```

foundation/foundation-configuration-store/pom.xml
jaebchoi: Yup it is needed for` import org.json.JSONObject;` in the ConfigMutatingWebhook.java
We had to convert json string to json object to retrieve key value pair for Data

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -83,9 +87,18 @@ public AdmissionReview validate(AdmissionReview admissionReviewRequest) {
...
             String json = mapper.writeValueAsString(request.getObject());
+
+            //This scans to see if there's any config need to be injected and replace corresponding values and saves into updatedJson
             String updatedJson = replaceConfigStoreKeysWithConfigValue(json);
 
-            if (!updatedJson.isEmpty()) {
+            if(request.getKind().getKind().equals("Secret")) {
+                //if replaceConfigStoreKeysWithConfigValue doesn't have any injection it would return empty, so we need original json to scan.
+                String secretJson = updatedJson.isEmpty() ? json: updatedJson;
+                boolean plainTextInjectionUpdated = !updatedJson.isEmpty();
```
carter-cundiff:
```suggestion
                boolean hasPlainTextInjection = !updatedJson.isEmpty();
```

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -83,9 +87,18 @@ public AdmissionReview validate(AdmissionReview admissionReviewRequest) {
...
         try {
             String json = mapper.writeValueAsString(request.getObject());
+
+            //This scans to see if there's any config need to be injected and replace corresponding values and saves into updatedJson
             String updatedJson = replaceConfigStoreKeysWithConfigValue(json);
 
-            if (!updatedJson.isEmpty()) {
+            if(request.getKind().getKind().equals("Secret")) {
+                //if replaceConfigStoreKeysWithConfigValue doesn't have any injection it would return empty, so we need original json to scan.
+                String secretJson = updatedJson.isEmpty() ? json: updatedJson;
```
carter-cundiff:
```suggestion
                String secretJson = hasPlainTextInjection ? updatedJson: json;
```

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -129,9 +142,58 @@ private String getProperty(String groupName,String propertyName) {
...
+
+    /**
+     * This method is to inject values in the data fields for any encoded Configvalues from the properties file and save it as based64 encoded.
+     *
+     * @param json json string to scan
+     * @param mapper json object mapper
+     * @param plainTextInjectionUpdated boolean that plain text injection is updated.
+     * @return updatedJSON  If no injection, return empty.
+     */
+    private String replaceSecretConfigStoreKeyWithConfigValue(String json, ObjectMapper mapper, boolean plainTextInjectionUpdated) throws JsonProcessingException {
```
carter-cundiff:
```suggestion
    private String replaceSecretConfigStoreKeyWithConfigValue(String json, ObjectMapper mapper, boolean hasPlainTextInjection) throws JsonProcessingException {
```

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -129,9 +142,58 @@ private String getProperty(String groupName,String propertyName) {
...
+        String updatedJSON = json;
+        JSONObject jsonObject = new JSONObject(json);
+        //If there's no data key, no injection can be done.
+        if(!jsonObject.has("data")) {
+            return plainTextInjectionUpdated ? json : "";
+        }
+        String dataString = jsonObject.getString("data");
+        Map<String, String> secretDataMap = mapper.readValue(dataString, new TypeReference<>() {});
+
+        //key being base64 val that has getConfigValue value being actual value to inject from property.
```
carter-cundiff:
```suggestion
        //key being base64 value that contains getConfigValue(), value being base64 value injected with the actual property.
```

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -129,9 +142,58 @@ private String getProperty(String groupName,String propertyName) {
...
+        Map<String, String> injectionMap = new HashMap<>();
+
+        for(String secretDataVal : secretDataMap.values()) {
+            String decodedStr = new String(Base64.getDecoder().decode(secretDataVal));
+
+            if(decodedStr.startsWith(CONFIG_STORE_INJECT_START)) {
```
carter-cundiff: Think we could use contains here. We want to support injection where ever it is in the base64 string, not just at the beginning.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -129,9 +142,58 @@ private String getProperty(String groupName,String propertyName) {
...
+            return plainTextInjectionUpdated ? json : "";
+        }
+        String dataString = jsonObject.getString("data");
+        Map<String, String> secretDataMap = mapper.readValue(dataString, new TypeReference<>() {});
+
+        //key being base64 val that has getConfigValue value being actual value to inject from property.
+        Map<String, String> injectionMap = new HashMap<>();
+
+        for(String secretDataVal : secretDataMap.values()) {
+            String decodedStr = new String(Base64.getDecoder().decode(secretDataVal));
```
carter-cundiff, line 164: So after we decode the string, why can't we just reuse the injection logic from `replaceConfigStoreKeysWithConfigValue`? No need to have two different ways of injecting our config values into a string.

carter-cundiff: Something like this:
```
String updatedDecodedStr = replaceConfigStoreKeysWithConfigValue(decodedStr);

if (!updatedDecodedStr.isEmpty()) {
  // Reencode and add back to JSON
}

File: foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java (hunk 7 above)
Comment (jaebchoi) [reply foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java]: interesting, so if we have something like this:
`test : test-$getConfigValue(groupName={spark};propertyName={testprop})-123`

and property is
`testprop=testing`

then resulting value would be
`
test: test-testing-123`

correct?

File: foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java (hunk 8 above)
Comment (jaebchoi) [reply foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java:164]: I could but though it would be faster to just create Map and replace values instead of iterating whole data string splitted with ConfigStoreInjectionStart. 

If this seems hard to read I can def replicate logic with replaceConfigStoreKeysWithConfigValue

File: foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java (hunk 7 above)
Comment (carter-cundiff) [reply foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java]: Yes. Why someone would do that idk, but it seems simple enough to support it without any negative consequences.

File: foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java (hunk 8 above)
Comment (jaebchoi) [reply foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java:164]: on another thought I think it's simpler, Ill update that!

File: foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java [hunk 9]
```diff
@@ -83,9 +87,19 @@ public AdmissionReview validate(AdmissionReview admissionReviewRequest) {
         AdmissionResponse response;
         try {
             String json = mapper.writeValueAsString(request.getObject());
+
+            //This scans to see if there's any config need to be injected and replace corresponding values and saves into updatedJson
             String updatedJson = replaceConfigStoreKeysWithConfigValue(json);
 
-            if (!updatedJson.isEmpty()) {
+            JSONObject jsonObject = !updatedJson.isEmpty() ? new JSONObject(updatedJson) : new JSONObject(json);
```
carter-cundiff:
```suggestion
            JSONObject jsonObject = new JSONObject(json);
```
We're just interested in whether the `data` field exists, this shouldn't be affected by injection.

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
carter-cundiff: Still needs to be updated

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
jaebchoi: if updatedJson have nothing to inject for plaintext, it would return empty string, not original JSON in line 92, in which could result in issue trying to instantiate JSON object of empty string. (tested in BDD)

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
jaebchoi: def missed,updated!

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
```diff
@@ -129,9 +143,49 @@ private String getProperty(String groupName,String propertyName) {
...
+            updatedJSON = updatedJSON.replace(injection.getKey(), injection.getValue());
+        }
+
+        //We still need to carry out plain text injection differences.
+        if(hasPlainTextInjection) {
+            return updatedJSON.equals(json) ? json : updatedJSON;
+        }
+
+        //If there is no change in injection in both secret and plain text, return empty. If there are any change, return updated JSON.
+        return updatedJSON.equals(json) ? "" : updatedJSON;
```
carter-cundiff: Think this return section can be made more readable:
```Java
if (updatedJSON.equals(json)) {
  if (hasPlainTextInjection) {
    return json; // no change in secret text, return the json with plain text injections
  } else {
    return "";  // no change in both secret and plain text, return empty
  }
} else {
  return updatedJSON;
}

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
carter-cundiff: This would be using the original Json object so it should never be empty. We don't need to check the updated json for the data field.

carter-cundiff (approved):

foundation/foundation-configuration-store/src/main/java/com/boozallen/aissemble/configuration/mutating/webhook/ConfigMutatingWebhook.java
jaebchoi: you're 100% correct, didn't look this throughly updated!

PR #551, opened 2025-01-29
[#540] update jars/libs in hive image

### LINKED ISSUE #540: Feature: Remove unneeded JARs from Hive image
## Description
We recently moved from using a Hive standalone JAR to using the official Hive image as the base for our `aissemble-hive-service` image. The new image actually includes a lot more JARs than are actually necessary for the standalone metastore. Unfortunately, the 4.0 standalone JAR is broken (see https://github.com/apache/hive/pull/5419). We should remove these unnecessary JARs either by manually fixing the standalone tarball, or manually removing the JARs.

## DOD
- [x] Hive image does not have JARs other than those needed for the standalone metastore (prefer fixing tarball)
  - [x] Automated test for Hive is added
- [x] Spark and Hive inter-op as before
  - [x] Thrift service
  - [x] Metastore catalog

## Test Strategy/Script
1. Create downstream project from the archetype:
    ```sh
    mvn archetype:generate '-DarchetypeGroupId=com.boozallen.aissemble'\
                           '-DarchetypeArtifactId=foundation-archetype'\
                           '-DarchetypeVersion=1.11.0-SNAPSHOT'\
                           '-DgroupId=org.test'\
                           '-Dpackage=org.test'\
                           '-DprojectGitUrl=test.org/test.git'\
                           '-DprojectName=Test Project'\
                           '-DartifactId=final-540'\
    && cd final-540
    ```
2. Add a [SparkPipeline.json](https://github.com/user-attachments/files/18593544/SparkPipeline.json) with hive persistence to _final-540-pipeline-models/src/main/resources/pipelines_
3. Fully generate the project
    1. Do a final build with `-Dmaven.build.cache.skipCache` to ensure no hidden manual actions
4. **OTS only:**
    1. Point `spark-infrastructure` Hive dependency in Chart.yaml to the locally built & exploded hive chart
5. Update the Ingest step to write a sample dataset (at _final-540-pipelines/spark-pipeline/src/main/java/org/test/Ingest.java_)
    ```diff
     package org.test;
     
     import jakarta.enterprise.context.ApplicationScoped;
     
    +import org.apache.spark.sql.Dataset;
    +import org.apache.spark.sql.Encoders;
    +import org.apache.spark.sql.SaveMode;
     import org.slf4j.Logger;
     import org.slf4j.LoggerFactory;
     
    -import java.time.Instant;
    -import java.util.Map;
    +import java.util.List;
     
    public class Ingest extends IngestBase {
       ...
           /**
          * Performs the business logic for Ingest.
          */
         @Override
         protected void executeStepImpl() {
    -        // TODO: Add your business logic here for this step!
    -        logger.error("Implement executeStepImpl(..) or remove this pipeline step!");
    +        Dataset<String> dataset = sparkSession.createDataset(List.of("one", "two", "three"), Encoders.STRING());
    +        dataset.write().format("hive").mode(SaveMode.Append).saveAsTable("tsttab");
    
         }
      ...
     }
    ```
6. Work around #549 by editing _final-540-deploy/src/main/resources/apps/spark-infrastructure/values.yaml_ with the following changes:
    ```diff
    aissemble-thrift-server-chart:
      app:
        name: "thrift-server"
     
    +  deployment:
    +    envFromSecret:
    +      AWS_ACCESS_KEY_ID:
    +        secretName: remote-auth-config
    +        key: AWS_ACCESS_KEY_ID
    +      AWS_SECRET_ACCESS_KEY:
    +        secretName: remote-auth-config
    +        key: AWS_SECRET_ACCESS_KEY
    +
       sparkConf: |
         spark.hadoop.fs.s3a.endpoint=http://s3-local:4566
    +    spark.hadoop.fs.s3a.access.key=${env:AWS_ACCESS_KEY_ID}
    +    spark.hadoop.fs.s3a.secret.key=${env:AWS_SECRET_ACCESS_KEY}
         spark.hadoop.fs.s3.impl=org.apache.hadoop.fs.s3a.S3AFileSystem
         spark.hive.server2.thrift.port=10000
         spark.hive.server2.thrift.http.port=10001
    ...

    aissemble-hive-metastore-service-chart:
      mysql:
        auth:
          # Note: Changing these values requires removal of the `hive-metastore-db-0` PVC, or manual modification of the
          # persisted database.
          username: hive
          rootPassword: hive
    ...
     
    +  deployment:
    +    env:
    +      - name: AWS_SECRET_ACCESS_KEY
    +        valueFrom:
    +          secretKeyRef:
    +            name: remote-auth-config
    +            key: AWS_SECRET_ACCESS_KEY
    +      - name: AWS_ACCESS_KEY_ID
    +        valueFrom:
    +          secretKeyRef:
    +            name: remote-auth-config
    +            key: AWS_ACCESS_KEY_ID
    +
       configMap:
         metastoreServiceConfig:
           properties:
   ```
7. Build and deploy the project: `mvn clean install && tilt up; tilt down`
8. Run the `spark-pipeline` resource and wait for the pipeline to finish
9. Query with thrift:
    ```sh
    SPARK_OP_PODID=$(kubectl get pods -A -l app.kubernetes.io/name=spark-operator --no-headers -o custom-columns=":metadata.name")
    kubectl exec $SPARK_OP_PODID -- ../bin/beeline \
      -u "jdbc:hive2://thrift-server:10001/default;transportMode=http;httpPath=cliservice" \
      -e 'select * from tsttab;'
    ```
10. Verify the output from the above query is:
    ```sh
    +--------+
    | value  |
    +--------+
    | one    |
    | two    |
    | three  |
    +--------+
    ```

epwilkins:
- Add automated tests to ensure the hive metastore can start
 - Better support using the Hive image as a standalone artifact by moving logic buried in the Helm chart into the base image
 - Switch to the standalone-metastore version of Hive
 - Update Hadoop from 3.3.6 to 3.4.1
 - Patch in compatible updated library jars to fix issues

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -35,6 +35,7 @@
         <version.fermenter>2.10.5</version.fermenter>
         <version.fermenter.legacy.tools>2.8.0</version.fermenter.legacy.tools>
         <version.habushu.plugin>2.17.0</version.habushu.plugin>
+        <version.hive>4.0.0</version.hive>
         <version.python>3.11.4</version.python>
         <version.help.plugin>3.5.0</version.help.plugin>
         <version.krausening>20</version.krausening>
@@ -87,9 +88,9 @@
         <version.sedona>1.4.0</version.sedona>
         <version.geotools.wrapper>1.4.0-28.2</version.geotools.wrapper>
         <version.mysql-connector>8.0.30</version.mysql-connector>
-        <version.hadoop>3.3.4</version.hadoop>
+        <version.hadoop>3.4.1</version.hadoop>
         <version.neo4j>4.1.5_for_spark_3</version.neo4j>
-        <version.aws.sdk.bundle>1.12.262</version.aws.sdk.bundle>
+        <version.aws.sdk.bundle>1.12.780</version.aws.sdk.bundle>
         <version.baton>1.1.1</version.baton>
         <version.quarkus.cucumber>1.0.0</version.quarkus.cucumber>
         <version.quarkus.cucumber.java>${version.cucumber}</version.quarkus.cucumber.java>
--- a/extensions/extensions-docker/aissemble-hive-service/pom.xml
+++ b/extensions/extensions-docker/aissemble-hive-service/pom.xml
@@ -15,8 +15,9 @@
     <packaging>docker-build</packaging>
 
     <properties>
-        <dockerbuild.jars.directory>target/dockerbuild/jars</dockerbuild.jars.directory>
-        <version.hive.metastore>4.0.0</version.hive.metastore>
+        <dockerbuild.jars.directory>target/dockerbuild/lib</dockerbuild.jars.directory>
+        <dockerbuild.bin.directory>target/dockerbuild/bin</dockerbuild.bin.directory>
+        <dockerbuild.patch.directory>target/dockerbuild/patch</dockerbuild.patch.directory>
     </properties>
 
     <dependencies>
@@ -26,9 +27,73 @@
             <version>3.2.0</version>
         </dependency>
         <dependency>
-            <groupId>mysql</groupId>
-            <artifactId>mysql-connector-java</artifactId>
-            <version>${version.mysql-connector}</version>
+            <groupId>org.apache.hive</groupId>
+            <artifactId>hive-standalone-metastore-server</artifactId>
+            <version>${version.hive}</version>
+            <classifier>bin</classifier>
+            <type>tar.gz</type>
+        </dependency>
+
+        <!-- Patch JARs -->
+        <dependency>
+            <groupId>io.airlift</groupId>
+            <artifactId>aircompressor</artifactId>
+            <version>0.27</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.avro</groupId>
+            <artifactId>avro</artifactId>
+            <version>1.11.4</version>
+        </dependency>
+        <dependency>
+            <groupId>dnsjava</groupId>
+            <artifactId>dnsjava</artifactId>
+            <version>3.6.2</version>
+        </dependency>
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-protobuf</artifactId>
+            <version>1.70.0</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>33.4.0-jre</version>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+            <version>2.18.2</version>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-io</artifactId>
+            <version>9.4.57.v20241219</version>
+        </dependency>
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-codec-http2</artifactId>
+            <version>4.1.117.Final</version>
+        </dependency>
+        <dependency>
+            <groupId>com.nimbusds</groupId>
+            <artifactId>nimbus-jose-jwt</artifactId>
+            <version>9.48</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.protobuf</groupId>
+            <artifactId>protobuf-java</artifactId>
+            <version>3.25.5</version>
+        </dependency>
+        <dependency>
+            <groupId>org.xerial.snappy</groupId>
+            <artifactId>snappy-java</artifactId>
+            <version>1.1.10.7</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.zookeeper</groupId>
+            <artifactId>zookeeper</artifactId>
+            <version>3.9.3</version>
         </dependency>
     </dependencies>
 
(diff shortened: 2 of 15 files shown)


extensions/extensions-docker/aissemble-hive-service/pom.xml
```diff
@@ -15,8 +15,9 @@
     <packaging>docker-build</packaging>
 
     <properties>
-        <dockerbuild.jars.directory>target/dockerbuild/jars</dockerbuild.jars.directory>
-        <version.hive.metastore>4.0.0</version.hive.metastore>
+        <dockerbuild.jars.directory>target/dockerbuild/lib</dockerbuild.jars.directory>
+        <dockerbuild.bin.directory>target/dockerbuild/bin</dockerbuild.bin.directory>
```
epwilkins, line 19: I: Destination for the Hive standalone metastore tarball. We download this via Maven to take advantage of caching.

extensions/extensions-docker/aissemble-hive-service/pom.xml
```diff
@@ -15,8 +15,9 @@
...
+        <dockerbuild.patch.directory>target/dockerbuild/patch</dockerbuild.patch.directory>
```
epwilkins, line 20: I: Destination for JARs that we want to patch in the final image. We download this via Maven to take advantage of caching.

pom.xml
```diff
@@ -60,7 +60,6 @@
         <ossrh.repo.url>https://oss.sonatype.org/content/repositories</ossrh.repo.url>
         <s01.ossrh.repo.url>https://s01.oss.sonatype.org/content/repositories</s01.ossrh.repo.url>
         <ghcr.io.repo.url>https://maven.pkg.github.com/boozallen/aissemble</ghcr.io.repo.url>
-        <version.avro>1.10.1</version.avro>
```
epwilkins, line 63: I: This was completely unused

extensions/extensions-docker/pom.xml
```diff
@@ -86,6 +86,58 @@
                         </execution>
                     </executions>
                 </plugin>
+                <plugin>
+                    <groupId>${group.fabric8.plugin}</groupId>
+                    <artifactId>docker-maven-plugin</artifactId>
+                    <executions>
+                        <execution>
+                            <id>test-image</id>
```
epwilkins, line 94: I: Since we now have 3 images that we test, moved the common config up to the Docker parent. Because of limitations around how executions can be activated, these test executions are set to the phase `none` to disable them by default. Modules that have automated tests added will need to ensure they hook into the same execution via `id` and set the `phase` to `integration-test`/`post-integration-test` (see POM comment).

extensions/extensions-docker/aissemble-hive-service/src/main/resources/docker/Dockerfile
```diff
@@ -1,32 +1,64 @@
...
+ARG BIN_DIR
+ARG PATCH_DIR
 
 ENV HADOOP_HOME=/opt/hadoop
 ENV HIVE_HOME=/opt/hive
+ENV HIVE_VER=$METASTORE_VERSION
+
+# Install hadoop
+RUN cd /tmp \
+    && wget https://archive.apache.org/dist/hadoop/common/hadoop-${HADOOP_VERSION}/hadoop-${HADOOP_VERSION}.tar.gz -O - \
```
epwilkins: I: This download takes forever, so I moved it as high in the build as possible.  Unfortunately, the Hadoop project does not cleanly deploy a Maven artifact that we can use like Hive does. I also looked at using the Hadoop Docker image like we were doing with the Hive image previously, but they only provide AMD images. It probably would have been safe, and I can switch to it if there's strong enough support for it. My reasoning for going this route is that, on balance, this layer should rarely be invalidated except when the Docker cache is cleared which often means cached images are also cleared. Therefore, the time cost ends up being roughly equivalent with less risk of AMD/ARM issues.

extensions/extensions-docker/aissemble-hive-service/pom.xml
```diff
@@ -66,10 +177,39 @@
...
+                        <id>get-patch-jars</id>
+                        <phase>prepare-package</phase>
+                        <goals>
+                            <goal>copy-dependencies</goal>
+                        </goals>
+                        <configuration>
+                            <outputDirectory>${dockerbuild.patch.directory}</outputDirectory>
+                            <includeTypes>jar</includeTypes>
+                            <excludeArtifactIds>delta-hive_2.12,mysql-connector-java</excludeArtifactIds>
+                            <useRepositoryLayout>true</useRepositoryLayout>
```
epwilkins, line 186: I: By using this layout we encode the group, artifact and version of each JAR into the path, making discovery of candidates for replacement _much_ simpler.

extensions/extensions-docker/aissemble-hive-service/src/main/resources/scripts/setup.sh
```diff
@@ -0,0 +1,89 @@
...
+
+#---
+## Patches a lib directory with given updated JARs by matching on base name. For example, given the replace directory has
+## the jar `commons-math-3.4.jar` and the target directory has `commons-math-2.1.jar`, the existing 2.1 JAR is deleted
+## and the updated 3.4 jar is copied into the directory.
+##
+## @Param: location of replacement JARs
+## @Param: location to search for JARs to replace
+#---
+update_jars() {
```
epwilkins, line 34: I: I think we could benefit from using this same approach for the Spark image updates, again to take advantage of Maven's caching, however there wasn't enough benefit in my mind to justify reworking that build at this time. (Especially since much of that logic will be made obsolete when Spark 4 comes out.)

extensions/extensions-helm/extensions-helm-spark-infrastructure/aissemble-hive-metastore-service-chart/templates/deployment.yaml
```diff
@@ -33,25 +33,13 @@ spec:
       nodeSelector:
         {{- toYaml . | nindent 12}}
       {{- end }}
-      initContainers:
-        - name: "init-or-upgrade-schema"
```
epwilkins, line 37: I: In the stock Hive image, this was being done as part of `entrypoint.sh`. I pulled in their script and adapted it for the standalone version of Hive to make the image more directly usable outside of the context of this Helm chart. It ends up being logically equivalent.

extensions/extensions-docker/aissemble-hive-service/src/main/resources/docker/Dockerfile
```diff
@@ -1,32 +1,64 @@
...
 ADD ${JARS_DIR}/* $HIVE_HOME/lib/
 
 # Remove jars with open vulnerabilities. These jars are included in the apache hive image but not necessary
 # when running the hive metastore only
-RUN rm ${HIVE_HOME}/lib/avatica-1.12.0.jar ${HIVE_HOME}/lib/htrace-core-3.1.0-incubating.jar \
-    ${HADOOP_HOME}/share/hadoop/yarn/timelineservice/lib/htrace-core-3.1.0-incubating.jar  
+RUN rm ${HADOOP_HOME}/share/hadoop/yarn/timelineservice/lib/htrace-core-3.1.0-incubating.jar
 
 USER hive
+WORKDIR $HIVE_HOME
```
epwilkins, line 72: S: This along with the change to JAVA_HOME in the base chart are potential breaking changes. Update DRAFT_RELEASE_NOTES accordingly.

build-parent/pom.xml
```diff
@@ -87,9 +88,9 @@
         <version.sedona>1.4.0</version.sedona>
         <version.geotools.wrapper>1.4.0-28.2</version.geotools.wrapper>
         <version.mysql-connector>8.0.30</version.mysql-connector>
-        <version.hadoop>3.3.4</version.hadoop>
+        <version.hadoop>3.4.1</version.hadoop>
```
csun-cpointe, line 91: Q: should we also update the hadoop package version in the yaml file ex: spark-infrastructure value template

epwilkins: Good catch. This package is actually already on the image by default, so I'll just remove its specification from the template.

csun-cpointe (approved):

extensions/extensions-docker/aissemble-hive-service/pom.xml
```diff
@@ -42,22 +117,58 @@
...
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>exec-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>test-image-cleanup</id>
+                        <phase>post-integration-test</phase>
```
carter-cundiff, line 153: Q: Should there be more to this execution?

extensions/extensions-docker/aissemble-hive-service/pom.xml
```diff
@@ -66,10 +177,39 @@
                         <goals>
                             <goal>copy-dependencies</goal>
                         </goals>
+                        <configuration>
+                            <outputDirectory>${dockerbuild.jars.directory}</outputDirectory>
+                            <includeTypes>jar</includeTypes>
+                            <includeArtifactIds>delta-hive_2.12,mysql-connector-java</includeArtifactIds>
```
carter-cundiff: Looks like we go rid of `mysql-connector-java` as a dependency, can remove from here

carter-cundiff: Can make this execution ID more descriptive as well instead of `unpack`

extensions/extensions-docker/aissemble-hive-service/pom.xml
epwilkins: This is just enabling the execution. See the [comment](https://github.com/boozallen/aissemble/pull/551/files/d5d50d6c291bef2dd5f9a5c8952e1c92456c9afb#r1934534788) on the parent Docker POM

extensions/extensions-docker/aissemble-hive-service/pom.xml
epwilkins: Good catch.

carter-cundiff (approved):

PR #567, opened 2025-02-05
[538] Supporting functionality for encryption of records with relations Part 1

aaron-gary:
Added ability to compose a Spark transform that can encrypt nested fields.

Added exception to generated code when an encryption policy exists and the associated record has relations.  This is intended to inform the user that encryption of nested fields is not fully implemented.

### DIFF
--- a/foundation/foundation-mda/pom.xml
+++ b/foundation/foundation-mda/pom.xml
@@ -170,6 +170,12 @@
             <groupId>net.masterthought</groupId>
             <artifactId>cucumber-reporting</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.boozallen.aissemble</groupId>
+            <artifactId>foundation-encryption-policy-java</artifactId>
+            <version>${version.aissemble}</version>
+        </dependency>
+
     </dependencies>
 
 </project>
--- a/foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/BaseStepDecorator.java
+++ b/foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/BaseStepDecorator.java
@@ -440,4 +440,21 @@ public class BaseStepDecorator implements Step {
 	public boolean isGeneric() {
 		return "generic".equalsIgnoreCase(getType());
 	}
+
+	/**
+	 * Whether or not the step has an inbound native collection type defined
+	 * and also contains relations.
+	 *
+	 * @return true if the step has an inbound native collection type defined and has relations.
+	 */
+	public boolean hasInboundNativeCollectionTypeAndRelations() {
+		boolean isNative = getInbound() != null && getInbound().getNativeCollectionType() != null;
+		boolean hasRelations = false;
+		if(getInbound() != null && getInbound().getRecordType() != null) {
+			Record thisRecord = getInbound().getRecordType().getRecordType();
+			hasRelations = thisRecord.getRelations() != null && !thisRecord.getRelations().isEmpty();
+		}
+
+		return isNative && hasRelations;
+	}
 }
--- a/foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/Record.java
+++ b/foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/Record.java
@@ -10,6 +10,7 @@ package com.boozallen.aiops.mda.metamodel.element;
  * #L%
  */
 
+import com.boozallen.aiops.mda.metamodel.AissembleModelInstanceRepository;
 import org.technologybrewery.fermenter.mda.metamodel.element.NamespacedMetamodel;
 
 import java.util.Collection;
@@ -48,6 +49,16 @@ public interface Record extends NamespacedMetamodel {
      * @return fields
      */
     List<RecordField> getFields();
+
+
+    /**
+     * Returns the fields ids contained in this field container, including field ids from related records.
+     *
+     * @param metadataRepo a configured MetadataRepo
+     * @return a list of field ids
+     */
+    List<String> getFieldIds(AissembleModelInstanceRepository metadataRepo);
+
 
     /**
      * Returns the list of supported Frameworks (e.g. PySpark).
--- a/foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/RecordElement.java
+++ b/foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/RecordElement.java
@@ -11,6 +11,7 @@ package com.boozallen.aiops.mda.metamodel.element;
  */
 
 import com.boozallen.aiops.mda.ManualActionNotificationService;
+import com.boozallen.aiops.mda.metamodel.AissembleModelInstanceRepository;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
@@ -20,6 +21,7 @@ import org.technologybrewery.fermenter.mda.metamodel.element.NamespacedMetamodel
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 
 /**
  * Represents a record instance.
@@ -89,6 +91,49 @@ public class RecordElement extends NamespacedMetamodelElement implements Record 
     @Override
     public List<RecordField> getFields() {
         return fields;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public List<String> getFieldIds(AissembleModelInstanceRepository metadataRepo) {
+        List<String> fieldIds = new ArrayList<>();
+        for(RecordField recordField: this.getFields()) {
+            fieldIds.add(recordField.getName());
+        }
+
+        List<String> relatedFieldIds = getFieldIdsFromRelations(metadataRepo);
+        fieldIds.addAll(relatedFieldIds);
+
+        return fieldIds;
+    }
+
+    /***
+     * This method gets the fully qualified field ids from relations (dot notation).
+     * For example: FieldBParent.FieldB
+     * @param metadataRepo the configured metadataRepop
+     * @return a list of fully qualified field names
+     */
+    private List<String> getFieldIdsFromRelations(AissembleModelInstanceRepository metadataRepo) {
+        List<String> fieldIds = new ArrayList<>();
+        List<Relation> relations =  getRelations();
+        for(Relation relation: relations) {
+            String relationPackage = relation.getPackage();
+            String relationName = relation.getName();
+            Record relatedRecord = metadataRepo.getRecord(relationPackage, relationName);
+
+            if(relatedRecord != null) {
+
+                List<String> updatedList = relatedRecord.getFieldIds(metadataRepo).stream()
+                        .map(s -> relation.getColumn() + "." + s) // Prepend the prefix
+                        .collect(Collectors.toList());
+
+                fieldIds.addAll(updatedList);
+            }
+        }
+
+        return fieldIds;
     }
 
     /**
--- a/foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/RelationElement.java
+++ b/foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/RelationElement.java
@@ -92,6 +92,15 @@ public class RelationElement extends NamespacedMetamodelElement implements Relat
         this.documentation = documentation;
     }
 
+    /***
+     * Sets the Column name.
+     * @param column the column name
+     */
+    public void setColumn(String column) {
+        // This method is used in testing. i.e. not normally set from the RelationElement directly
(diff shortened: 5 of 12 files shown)


foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/RecordElement.java
```diff
@@ -91,6 +93,53 @@ public List<RecordField> getFields() {
         return fields;
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public List<String> getFieldIds(AissembleModelInstanceRepository metadataRepo) {
```
aaron-gary, line 100: The RecordElement does not have a context for metadataRepo, so opted to have the calling code supply it.
This is mainly so we can access the related records.

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/BaseStepDecorator.java
```diff
@@ -440,4 +440,21 @@ public List<FileStore> getDecoratedFileStores() {
...
+	/**
+	 * Whether or not the step has an inbound native collection type defined
+	 * and also contains relations.
+	 *
+	 * @return true if the step has an inbound native collection type defined and has relations.
+	 */
+	public boolean hasInboundNativeCollectionTypeAndRelations() {
+		boolean isNative = getInbound() != null && getInbound().getNativeCollectionType() != null;
+		boolean hasRelations = false;
+		if(getInbound().getRecordType() != null) {
```
cwoods-cpointe: c/q: do we need a null check on getInbound()?

aaron-gary: Good call.  We should be checking that

- [x] check for null

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/RecordElement.java
```diff
@@ -91,6 +93,49 @@ public List<RecordField> getFields() {
...
+        List<Relation> relations =  getRelations();
+        for(Relation relation: relations) {
+            String relationPackage = relation.getPackage();
+            String relationName = relation.getName();
+            Record relatedRecord = metadataRepo.getRecord(relationPackage, relationName);
+
+            if(relatedRecord != null) {
+
+                List<String> updatedList = relatedRecord.getFieldIds(metadataRepo).stream()
+                        .map(s -> relation.getColumn() + "." + s) // Prepend the prefix
```
cwoods-cpointe, line 129: q: surprised to see relation.getColumn here instead of relation.getName. Is there a reason we are not using the name?

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/RelationElement.java
```diff
@@ -92,6 +92,16 @@ public void setDocumentation(String documentation) {
         this.documentation = documentation;
     }
 
+    /**
+     * Sets the documentation value.
```
cwoods-cpointe: a: javadocs are off

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/RelationElement.java
```diff
@@ -92,6 +92,16 @@ public void setDocumentation(String documentation) {
...
+     *
+     * @param documentation
+     *            documentation text
+     */
+    public void setColumn(String column) {
```
cwoods-cpointe, line 99: q: is this function used? I dont think we have any setters in the elements

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/util/RecordFieldEncryptionUtil.java
```diff
@@ -0,0 +1,123 @@
...
+ */
+public class RecordFieldEncryptionUtil {
+
+    public RecordFieldEncryptionUtil() {
+    }
+
+    /***
+     * This method constructs a full Spark transform of the given fields and ensures
+     * fields are wrapped in an encryption UDF wrapper when a field requires encryption
+     * @param encryptionPolicyInput A list of record fields
```
cwoods-cpointe: a: params are off

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/Record.java
```diff
@@ -49,6 +50,15 @@ public interface Record extends NamespacedMetamodel {
      */
     List<RecordField> getFields();
 
+
+    /**
+     * Returns the fields ids contained in this field container.
+     *
+     * @return a list of field ids
+     */
+    List<String> getFieldIds(AissembleModelInstanceRepository metadataRepo);
```
csun-cpointe, line 60: A: missing a `@param` in the java doc

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/RecordElement.java
aaron-gary: Good question.  The Spark transform needs to associate the nested field with it's parent.  In this case it's the column name not the relation name.

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/RelationElement.java
aaron-gary: Good catch

- [x] fix java docs

foundation/foundation-mda/src/main/resources/templates/data-delivery-pyspark/encryption.py.vm
```diff
@@ -28,14 +28,17 @@
             # If that property is missing then we should ignore the policy.
             if encrypt_policy.encryptPhase:
                 if self.step_phase.lower() == encrypt_policy.encryptPhase.lower():
-
+                #if (${step.hasInboundNativeCollectionTypeAndRelations()})
```
cwoods-cpointe: q: could we move this up and fail faster?

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/RelationElement.java
aaron-gary: It's only used in a test when we build the relation object (as opposed to when the mda creates it).  I can add a comment

- [x] Add comment

foundation/foundation-mda/src/test/java/com/boozallen/aiops/mda/metamodel/element/RecordSteps.java
```diff
@@ -35,13 +40,12 @@
 import io.cucumber.java.en.Given;
 import io.cucumber.java.en.Then;
 import io.cucumber.java.en.When;
-//import org.apache.commons.lang3.RandomStringUtils;
-//import org.technologybrewery.fermenter.mda.metamodel.element.FieldElement;
```
cwoods-cpointe, line 39: a: unused imports

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/util/RecordFieldEncryptionUtil.java
aaron-gary: Good catch

- [x] Fix java docs

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/Record.java
aaron-gary: Good catch

- [x] Fix java docs

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/util/RecordFieldEncryptionUtil.java
```diff
@@ -0,0 +1,123 @@
...
+        }
+        sb.append(", ");
+    }
+
+    /***
+     * This method checks if a field requires encryption.  If it does need encryption it will be
+     * wrapped in a UDF syntax. for example: encryptUDF(x.data.ssn, "AES_ENCRYPT")
+     * @param aFieldId the field name (including the dot path if it is a nested field)
+     * @param encryptionPolicyInput contains the encrypt fields and encrypt algorithm
+     * @return
```
csun-cpointe: 👍

foundation/foundation-mda/src/main/resources/templates/data-delivery-pyspark/encryption.py.vm
aaron-gary: There is a trade off with moving the exception up.  In it's current position it will only fail if the policy applies to a specific pipeline step, but I suppose it wouldn't hurt if it fails for all pipeline steps(?)

- [x] Move exception up

foundation/foundation-mda/src/test/java/com/boozallen/aiops/mda/metamodel/element/RecordSteps.java
aaron-gary: Ah!  Good catch.  I thought I caught all of these.

- [x] Remove unused imports

foundation/foundation-mda/src/main/resources/templates/data-delivery-pyspark/encryption.py.vm
aaron-gary: Now that I think about it - if we move it up the user won't have to wait for pre-processing to complete before finding out that the encryption will cause an exception.  Good catch

foundation/foundation-mda/pom.xml
```diff
@@ -170,6 +170,12 @@
             <groupId>net.masterthought</groupId>
             <artifactId>cucumber-reporting</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.boozallen.aissemble</groupId>
+            <artifactId>foundation-encryption-policy-java</artifactId>
```
epwilkins, line 175: Q: where is this being used?

csun-cpointe (approved):

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/RecordElement.java
aaron-gary: RecordFieldEncryptionUtil.java so we can reuse EncryptionPolicyInput.  It's not necessary, but we would be "repeating ourself" to rcreate it.

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/util/RecordFieldEncryptionUtil.java
```diff
@@ -0,0 +1,124 @@
...
+
+import com.boozallen.aissemble.data.encryption.policy.json.EncryptionPolicyInput;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Helps with encryption of nested fields
+ */
+public class RecordFieldEncryptionUtil {
```
epwilkins: Ah this must be the answer to my question. IMO, this would make more sense in the foundation encryption module itself instead of foundation MDA.

aaron-gary: Good observation.  Agree

- [ ] Investigate moving to foundation encryption

epwilkins: Actually if this is only used in tests at the moment, I'm a little worried it will lead to confusion. We could make a patch from it and attach it to the relation encryption issue so we don't lose the work when we pick back up encryption, but it also doesn't cause confusion about how encryption works in the mean time.

foundation/foundation-mda/src/main/resources/templates/data-delivery-pyspark/encryption.py.vm
```diff
@@ -23,19 +23,23 @@
         policy_manager = DataEncryptionPolicyManager.getInstance()
         retrieved_policies = policy_manager.policies
 
+        #if (${step.hasInboundNativeCollectionTypeAndRelations()})
+        if(len(retrieved_policies.items()) > 0):
+            raise NotImplementedError("Encryption of records that contain relations is not yet supported.")
```
epwilkins, line 28: A: if it's not too bad to hammer out, it'd be nice to add an assertion for this in test-mda-models to verify that we're indeed throwing an exception when we need to.

It would also give us an easy hook to assert functionality when we implement this feature

epwilkins (approved):

carter-cundiff (approved):

Issue #578, opened 2025-02-12
BUG: Simple pipeline never shuts down


epwilkins:
## Description
A simple, out-of-the-box pipeline with no business logic hangs when executed via Spark Operator. The pipeline runs through all steps and shuts down the Spark session successfully, but the pipeline driver pod stays running instead of moving to a completed state.

### Steps to Reproduce
Clear, specific, and detailed steps taken to enable reproduction of the bug for investigation.
1. Create a project with a one step pipeline that has `dataProfiling`, `provenance`, and `alerting` disabled (e.g. [SamplePipeline.json](https://github.com/user-attachments/files/18771079/SamplePipeline.json))
2. Fully generate the project
3. Deploy the project via Tilt
4. After all resources start, execute the pipeline from the Tilt UI
5. Wait for the pipeline to finish and shutdown the Spark session successfully
6. Check the status of the driver pod in Kubernetes. E.g.
```sh
kubectl get pod sample-pipeline-driver
```

### Expected Behavior
The driver pod has shutdown and transitioned to the **Completed** state

### Actual Behavior
The pipeline driver pod is still in a **Running** state, consuming cluster resources

### Additional Context
aiSSEMBLE version: 1.11.1
Environment: local deployment (also seen in higher environments)
Blocker: yes
Pipeline tail logs:
```
25/02/12 16:58:45 INFO SharedState: Warehouse path is 's3a://spark-infrastructure/warehouse'.
25/02/12 16:58:48 ERROR NoOpStep: Implement executeStepImpl(..) or remove this pipeline step!
25/02/12 16:58:48 INFO SparkContext: SparkContext is stopping with exitCode 0.
25/02/12 16:58:48 INFO SparkUI: Stopped Spark web UI at http://sample-pipeline-e36cf094fb188d7f-driver-svc.default.svc:4040/
25/02/12 16:58:48 INFO KubernetesClusterSchedulerBackend: Shutting down all executors
25/02/12 16:58:48 INFO KubernetesClusterSchedulerBackend$KubernetesDriverEndpoint: Asking each executor to shut down
25/02/12 16:58:48 WARN ExecutorPodsWatchSnapshotSource: Kubernetes client has been closed.
25/02/12 16:58:49 INFO MapOutputTrackerMasterEndpoint: MapOutputTrackerMasterEndpoint stopped!
25/02/12 16:58:49 INFO MemoryStore: MemoryStore cleared
[...]

PR #584, opened 2025-02-20
[#578] only generate messaging files if messaging is needed

epwilkins:


foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/ManualActionNotificationService.java
```diff
@@ -312,41 +312,6 @@ private void appendTiltHelmBuild(String appName, String deployArtifactId, String
...
-     * Adds a notification to update the Tiltfile.
-     * NOTE: This "k8s_yaml" line is also being output by appendTiltHelmBuild() and addSparkWorkerTiltResources() so
-     * we should consider refactoring to extract out common code.
-     *
-     * @param context          the generation context
-     * @param appName          the application name
-     * @param deployArtifactId the deploy artifact ID
-     * @param yamlFileName     the deploy artifact ID
-     */
-    public void addYamlTiltFileMessage(final GenerationContext context, final String appName,
```
epwilkins, line 325: I: This was no longer in use anywhere as we removed them when we converted to building the Docker images during the Maven build locally in version 1.11

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/generator/PipelineMessagingResourcesGenerator.java
```diff
@@ -23,7 +23,7 @@
  * Generates configuration code with no model interaction. This is often useful for
  * configuration files that must exist in some form or similar constructs.
  */
-public class TargetedPipelineResourcesGenerator extends AbstractResourcesGenerator {
+public class PipelineMessagingResourcesGenerator extends AbstractResourcesGenerator {
```
epwilkins, line 26: I: Renamed since I was changing the behavior of `shouldGenerate` to be messaging-specific, and this class was only being used for messaging resources anyway.

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/generator/TargetedPipelineJavaGenerator.java
```diff
@@ -84,13 +88,17 @@ public void generate(GenerationContext generationContext) {
             }
         }
         if(javaTargetPipeline.isAlertingSupportNeeded()) {
-            manualActionNotificationService.addNoticeToUpdateKafkaConfig(generationContext, "alerts");
+            manualActionNotificationService.addNoticeToUpdateKafkaConfig(generationContext, ALERT_CHANNEL);
+            // TODO: StringSerializer actually throws an exception. We need an AlertSerializer in alerting-core
```
epwilkins, line 92: I: Since this was already broken, it was outside the scope of this ticket to create a serializer

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/generator/TargetedPipelineJavaGenerator.java
```diff
@@ -84,13 +88,17 @@ public void generate(GenerationContext generationContext) {
...
+            manualActionNotificationService.addSmallRyeConnectorMessage(generationContext, artifactId, "Alert Producer", ALERT_CHANNEL, "org.apache.kafka.common.serialization.StringSerializer");
         }
         if(javaTargetPipeline.isMetadataNeeded()) {
-            manualActionNotificationService.addNoticeToUpdateKafkaConfig(generationContext, "metadata-ingest");
+            manualActionNotificationService.addNoticeToUpdateKafkaConfig(generationContext, METADATA_CHANNEL);
+            manualActionNotificationService.addSmallRyeConnectorMessage(generationContext, artifactId, "Metadata Producer", METADATA_CHANNEL, "com.boozallen.aissemble.core.metadata.producer.MetadataSerializer");
         }
         if(javaTargetPipeline.getDataLineage()) {
             manualActionNotificationService.addNoticeToUpdateKafkaConfig(generationContext, Constants.DATA_LINEAGE_CHANNEL_NAME);
+            manualActionNotificationService.addSmallRyeConnectorMessage(generationContext, artifactId, "Data Lineage Emitter", Constants.DATA_LINEAGE_CHANNEL_NAME, "org.apache.kafka.common.serialization.StringSerializer");
```
epwilkins, line 101: I: Since data lineage messaging transport is optional, you could make an argument that we don't show this and just rely on the docs, but IMO it's easy enough to disable this manual action for projects that don't want it.

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/BasePipelineDecorator.java
```diff
@@ -158,17 +158,13 @@ public String getCapitalizedName() {
...
-				}
-			}
-			if (step.getOutbound() != null) {
-				if ("messaging".equals(step.getOutbound().getType())) {
-					return true;
-				}
+	/**
+	 * @return true if any step uses `messaging` or `multimessaging` type inbound/outbound
+ 	 */
+	public boolean hasMessagingSteps() {
```
epwilkins, line 164: I: Renamed to reduce confusion with the new `isMessagingSupportNeeded` method

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/BaseStepDecorator.java
```diff
@@ -288,26 +288,6 @@ public boolean isNative(StepDataBinding stepDataBinding) {
...
     }
 
-    /**
-     * Whether or not the given step data binding has type 'messaging'.
-     * 
-     * @param stepDataBinding
-     *            the step data binding to check
-     * @return true if the given step data binding has type 'messaging'
-     */
-	public boolean isMessaging(StepDataBinding stepDataBinding) {
```
epwilkins, line 298: I: Relocated after changing visibility from `public` to `protected` to maintain method grouping. No other changes were made when relocating.

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/element/java/JavaStep.java
```diff
@@ -176,16 +176,6 @@ public Set<String> getImplImports() {
         return imports;
     }
 
-    @Override
-    public boolean isMessaging(StepDataBinding stepDataBinding) {
-        return stepDataBinding != null
-                && ("messaging".equalsIgnoreCase(stepDataBinding.getType()));
-    }
-
-    public boolean isMultiMessaging(StepDataBinding stepDataBinding) {
```
epwilkins, line 185: I: These were identical implementations to the parent methods, so just removed them.

foundation/foundation-mda/src/main/resources/templates/notifications/group-pom-dependency.vm
```diff
@@ -0,0 +1,5 @@
+You must add the following dependencies to ${pom} to build your pipeline!
```
epwilkins: A: Left-over file from a different cleanup activity that I ended up dropping. Delete this before merging.

foundation/foundation-mda/src/main/resources/templates/notifications/notification.yaml.tiltfile.vm
epwilkins, line 1: I: This notification file was no longer being used after the removal of the relevant method in `ManualActionNotificationService`

foundation/foundation-metadata/foundation-metadata-producer/src/main/resources/META-INF/microprofile-config.properties
epwilkins, line 1: I: Since metadata is always included as a dependency, this would cause the microprofile config to be present in the shaded pipeline jar.  For pipelines that use _any_ messaging features, we'll have generated the pipeline-specific config properties which would take precedence/overwrite this file.

I toyed with removing metadata as a default dependency and only adding it if/when it was enabled, but that work ended up being way too involved to shoe-horn into this bug fix.

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/generator/TargetedPipelineJavaGenerator.java
carter-cundiff: Wouldn't these configurations already be generated into the `microprofile-config.properties` file? Or is the idea that these will be responsive to upstream pipeline metamodel changes?

epwilkins: The latter. Because we now only generate the properties file when messaging is required, it is the case that usually the config would be present. But consider a pipeline that uses messaging inbound/outbound, and enables metadata later.

carter-cundiff (approved):

PR #586, opened 2025-02-25
[#585] fix Nvidia ARM build

### LINKED ISSUE #585: BUG: Docker image builds are failing during CI job
## Description
The nightly build on Thursday started failing to build the `aissemble-nvidia` image due to a `qemu: segmentation fault` during the ARM image build. It appears that the only change was the release of version 28.0.0 of the Docker engine. This led to the the docker:dind image tag being updated over night, which is used by the gh-arc-runner to expose the docker daemon of the k8s node to the runner.  The GH ARC chart hard-codes the dind image tag in the templates, so it's not readily configurable.

Note this issue could not be reproduced locally on ARM machines, so is likely to be specific to our CI environment.

## Definition of Done
 - Get a working build ASAP that can enable work in the short run as we find a more long-term solution

## Test Steps
1. Verify the latest aissemble build job was successful
    - For OTS: verify it was executed on the branch `585-fix-ci-docker-build`

## Implementation Details
 - See if the old Jenkins build still works
 - While it's building (and if it doesn't work) explore using Kustomize to update the the dind image tag to 27.5.1

epwilkins:
This change set does two things:
 - Add a kustomization to fix the Docker `dind` image to a specific version (v27.5.1)
 - Update the buildkit image tag for the docker-maven-plugin to a fixed version (v0.19.0)

The original issue appeared like it might be caused by the updated `dind` image which corresponded with the release of Docker Engine 28.0. The actual cause ended up being the update to the `buildkit-stable-1` tag of the buildkit image we were using (also corresponding to the engine).

### DIFF
--- a/devops/kustomization.yaml
+++ b/devops/kustomization.yaml
@@ -1,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - resources.yaml
+images:
+  - name: docker
+    newTag: 27.5.1-dind
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -581,7 +581,7 @@
                                                 <builderName>maven</builderName>
                                                 <driverOpts>
                                                     <!-- Fully qualify build driver image name to ensure any configured credentials are used -->
-                                                    <image>docker.io/moby/buildkit:buildx-stable-1</image>
+                                                    <image>docker.io/moby/buildkit:v0.19.0</image>
                                                 </driverOpts>
                                                 <cacheTo>type=local,dest=${user.home}/.docker/cache</cacheTo>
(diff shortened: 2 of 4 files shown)


devops/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - resources.yaml
+images:
+  - name: docker
+    newTag: 27.5.1-dind
```
epwilkins, line 8: I: While this is certainly not strictly necessary, I reasoned that it was likely good for build stabilization for the same reason as the buildkit image tag change.

woods-cpointe (approved):

cwoods-cpointe (approved):

carter-cundiff (approved):

Issue #589, opened 2025-02-27
Feature: Introduce UBI base containers where possible


csun-cpointe:
## Description
RHAT is more responsive to the security concerns. It is easier to get the security issue fixed if using the RHAT image, which helps us stay ahead of any CVE so we want to switch the current baseline images to use the RHAT UBI image if we are not using (and there is no) official base images or the community image.

The goal of the ticket is to ensure that we can adapt the new CVE more rapidly.

## DOD
Acceptance criteria required to realize the requested feature
- [x] Scan through all the baseline images and list out all the community images for further discussion regarding whether we should rewrite using the RHAT image
- [x] Upgrade the `aissemble-hive-service` image to use RHAT image if there is no published official image that we can use.
   - [x] Existing unit test can pass

## Test Strategy/Script
How will this feature be verified?
1. Create downstream project from the archetype:
    ```sh
    mvn archetype:generate '-DarchetypeGroupId=com.boozallen.aissemble'\
                           '-DarchetypeArtifactId=foundation-archetype'\
                           '-DarchetypeVersion=1.12.0-SNAPSHOT'\
                           '-DgroupId=org.test'\
                           '-Dpackage=org.test'\
                           '-DprojectGitUrl=test.org/test.git'\
                           '-DprojectName=Test Project'\
                           '-DartifactId=test-589'\
    && cd test-589
    ```
2. Add a [SparkPipeline.json](https://github.com/user-attachments/files/18593544/SparkPipeline.json) with hive persistence to test-589-pipeline-models/src/main/resources/pipelines_
3. Fully generate the project by repeating the build until no more manual actions
4. Update the Ingest step to write a sample dataset (at test-589-pipelines/spark-pipeline/src/main/java/org/test/Ingest.java_)
    ```diff
     package org.test;
     
     import jakarta.enterprise.context.ApplicationScoped;
     
    +import org.apache.spark.sql.Dataset;
    +import org.apache.spark.sql.Encoders;
    +import org.apache.spark.sql.SaveMode;
     import org.slf4j.Logger;
     import org.slf4j.LoggerFactory;
     
    -import java.time.Instant;
    -import java.util.Map;
    +import java.util.List;
     
    public class Ingest extends IngestBase {
       ...
           /**
          * Performs the business logic for Ingest.
          */
         @Override
         protected void executeStepImpl() {
    -        // TODO: Add your business logic here for this step!
    -        logger.error("Implement executeStepImpl(..) or remove this pipeline step!");
[...]

PR #592, opened 2025-02-28
#589 update the aissemble-hive-service base image to use RHAT UBI

csun-cpointe:


foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/generator/S3LocalValuesKubernetesGenerator.java
```diff
@@ -14,7 +14,6 @@
 import com.boozallen.aiops.mda.generator.common.VelocityProperty;
 import com.boozallen.aiops.mda.metamodel.AissembleModelInstanceRepository;
 import com.boozallen.aiops.mda.metamodel.element.Pipeline;
-import com.boozallen.aiops.mda.metamodel.element.python.MachineLearningPipeline;
```
csun-cpointe, line 17: I: a quick cleanup

extensions/extensions-docker/aissemble-hive-service/src/main/resources/docker/Dockerfile
```diff
@@ -11,7 +11,11 @@ ENV HADOOP_HOME=/opt/hadoop
 ENV HIVE_HOME=/opt/hive
 ENV HIVE_VER=$METASTORE_VERSION
 
+USER root
 # Install hadoop
+RUN set -eux; \
+    microdnf install -y wget gzip util-linux
+
```
carter-cundiff, line 18: A: Think it makes more send to keep the `# Install hadoop` comment with the `wget` command below

DRAFT_RELEASE_NOTES.md
```diff
@@ -80,6 +80,10 @@ To start your aiSSEMBLE upgrade, update your project's pom.xml to use the 1.11.0
 
 ## Conditional Steps
 
+### For projects that have customized the Hive service
+The hive service base image has been switched from `docker.io/eclipse-temurin:17-jre` to `registry.access.redhat.com/ubi9/openjdk-17-runtime:1.21`. The largest
+change will be switching from `apt-get` to `microdnf` as the package manager.
```
carter-cundiff: The largest change will be switching from `apt-get` to `microdnf` as the package manager. `->` Projects should update their Hive service Dockerfile to use `microdnf` instead `apt-get` as the package manager if applicable.

carter-cundiff (approved):

DRAFT_RELEASE_NOTES.md
epwilkins: They could have install scripts as part of their startup config in k8s/Helm YAML, which would also need changes.

extensions/extensions-docker/aissemble-hive-service/src/main/resources/docker/Dockerfile
```diff
@@ -11,7 +11,11 @@ ENV HADOOP_HOME=/opt/hadoop
 ENV HIVE_HOME=/opt/hive
 ENV HIVE_VER=$METASTORE_VERSION
 
+USER root
 # Install hadoop
+RUN set -eux; \
+    microdnf install -y wget gzip util-linux
```
epwilkins: Q: Is curl pre-installed? We could just switch from wget to that if so. Also, what is util-linux for?

csun-cpointe: yeah.. it seems curl is pre-installed. Will switch to use curl. For the util-linux, rev is required when running the hive/setup.sh. and rev is packaged inside the util-linux.

epwilkins (approved):

PR #606, opened 2025-03-07
Stabilize CI build

csun-cpointe:
The following changes were made to help stabilize the build and improve build times. The primary change was to incorporate a Docker builder using the Kubernetes driver, so that all images can be built on their native architecture.  Virtualization was the root of many of the problems over the last couple of weeks. This has the added benefit of significantly speeding up our image builds (~32 mins faster overall).

 - enable k8s driver for native multi-arch builds
 - update test Teams alerting webhook
 - only save caches if the job wasn't cancelled

For the webook change, the old webhook method was retired by Microsoft, and a new URL needed to be generated to switch to the new one.

See: https://devblogs.microsoft.com/microsoft365dev/retirement-of-office-365-connectors-within-microsoft-teams/

There are two outflows from this that could make sense to tackle:
 - enable use of the k8s driver in downstream projects that build on k8s
 - remove the Alerting Teams IT as it arguably doesn't add much value


devops/runnerset-values.yaml
```diff
@@ -5,3 +5,5 @@ maxRunners: 3
 template:
   spec:
     serviceAccountName: "arc-runner-set-aissemble-gha-rs-custom-permissions"
+    nodeSelector:
+      beta.kubernetes.io/instance-type: t4g.xlarge
```
csun-cpointe, line 9: I: updated the GH runner to target the ARM node since it's a bit faster than x86 so that we don't have to worry about virtualization issues with the k8s driver we can switch

carter-cundiff (approved):

Issue #619, opened 2025-03-17
Feature: Update aiSSEMBLE licensing to give users more flexibility in license selection


meliz19:
## Description
The purpose of this ticket is to update aiSSEMBLE licensing to give users more flexibility in the type of license they use in their project. 

## DOD
- [x] Update aissemble to use: `license-maven-plugin:2.5.0` and `booz-allen-licenses:1.3.0`. See https://github.com/boozallen/aissemble-parent/issues/11
- [x] Add request for organizationName  in archetype questions (default to Booz Allen)
- [x] ~Update archetype tests where necessary~ N/A
- [x] Update documentation where necessary
- [x] ~Potentially ask which license they want to use during archetype (though this may be more complicated than it is worth)~ The juice isn't worth the squeeze. It'd require a hacky solution. 

## Notes
- Make sure that `organizationName` only affects the license

## Test Strategy/Script
- Build aissemble (`mvn clean install`)
- Build a downstream project:
```
mvn archetype:generate -U -DarchetypeGroupId=com.boozallen.aissemble \
  -DarchetypeArtifactId=foundation-archetype \
  -DarchetypeVersion=1.13.0-SNAPSHOT \
  -DgroupId=com.test \
  -DartifactId=test-619 \
  -DprojectGitUrl=https://github.com/meliz19/test-619 \
  -DprojectName=test-619 \
  -DlicenseName=specific-client-license \
  -DorganizationName="Ministry of Magic" \
  && cd test-619
```
- Build the downstream project (`mvn clean install`)
- Run `cat LICENSE.txt` and verify it has the following contents: "Copyright (C) 2021 Ministry of Magic"

meliz19:
Completed DoD with @chang-annie

d-ryan-ashcraft:
## Test Results: ✅ PASSED
```
% cat LICENSE.txt
Copyright (C) 2021 Ministry of Magic
```

PR #620, opened 2025-03-19
[#619] Updated licensing versions and added organizationName to arche…

meliz19:
…type questionnaire

#619

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -737,46 +737,6 @@
                     </executions>
                 </plugin>
                 <plugin>
-                    <groupId>org.codehaus.mojo</groupId>
-                    <artifactId>license-maven-plugin</artifactId>
-                    <configuration>
-                        <licenseName>booz-allen-public-license</licenseName>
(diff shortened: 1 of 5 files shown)


build-parent/pom.xml
```diff
@@ -736,46 +736,6 @@
                         </execution>
                     </executions>
                 </plugin>
-                <plugin>
```
meliz19, line 739: I: Removing as this exact block is in the [aissemble-parent](https://github.com/boozallen/aissemble-parent/blob/dev/pom.xml).

d-ryan-ashcraft (approved):

docs/modules/ROOT/pages/archetype.adoc
```diff
@@ -111,6 +111,12 @@ under mvn license:license-list
 
 | No
 
+| ``organizationName``
+| The name of the organization to use in the project
+| Defaults to "Booz Allen Hamilton Inc."
```
carter-cundiff: Should we include `and can be updated in the <repository> section of the root pom.xml` like the other properties?

carter-cundiff (approved):

PR #624, opened 2025-03-20
[#614] make version and package consistent in archetype

### LINKED ISSUE #614: BUG: Default root POM does not upgrade automatically
## Description
Archetype embeds a static aiSSEMBLE version number in the root POM instead of using the `version.aissemble` property. This means the number must be manually updated when upgrading between versions of aiSSEMBLE.

## Steps to Reproduce
1. Create a project from the previous version of the archetype
2. Open the root POM of the project and update the `build-parent` version to the latest version of aiSSEMBLE
3. Run `mvn clean install`
4. Check the version number of the baseline dependencies in the root POM:
    ```sh
    mvn dependency:tree -pl :<project-name> | grep com.boozallen.aissemble
    ```

### Expected Behavior
All baseline aiSSEMBLE dependencies in the root POM have been updated to the latest version.

### Actual Behavior
The baseline aiSSEMBLE dependencies still reference the initial, un-upgraded version.

## Test Script
1. Create a project from the latest archetype
    ```sh
    mvn archetype:generate '-DarchetypeGroupId=com.boozallen.aissemble' \
                           '-DarchetypeArtifactId=foundation-archetype' \
                           '-DarchetypeVersion=1.13.0-SNAPSHOT' \
                           '-DgroupId=org.test' \
                           '-Dpackage=org.test.pkg' \
                           '-DprojectGitUrl=test.org/test.git' \
                           '-DprojectName=Test Project' \
                           '-DartifactId=bug-614' \
    && cd bug-614
    ```
2. Ensure the aissemble version is only hard-coded in the `<parent>` section of the root pom.xml
3. Ensure `<basePackage>com.boozallen</basePackage>` is not hard-coded anywhere in the project
    - All but one of the instances should be the package configured with the archetype: _org.test.pkg_

epwilkins:
Primarily fixes this ticket by using the `version.aissemble` POM property for the dependencies in the root POM, thereby automatically updating on upgrade. Also aligns Fermenter `basePackage` configurations. Some places were using the hard-coded `com.boozallen`, some were using `project.groupId` and some were using the archetype variable `package`. The latter seemed the most correct, so all instances were updated to use that _except_ the instance in *-pipeline-models. That case seems to just be a misconfigured MDA profile/template, and honestly could probably be moved to the archetype anyway.


foundation/foundation-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -66,33 +66,33 @@
         <dependency>
             <groupId>com.boozallen.aissemble</groupId>
             <artifactId>artifacts-maven-plugin</artifactId>
-            <version>${archetypeVersion}</version>
+            <version>${version.aissemble}</version>
```
epwilkins, line 69: I: Primary fix for upgrade issue.

meliz19 (approved):

aaron-gary (approved):

PR #640, opened 2025-04-01
[#636] add docker caches to long builds

### LINKED ISSUE #636: Feature: Fix failing Jenkins Controller image build
## Description
A Jenkins plugin is no longer installable, causing the Jenkins controller image to fail to build.
```sh
[INFO] DOCKER>  > [ 8/11] RUN jenkins-plugin-cli --plugins workflow-cps:2692.v76b_089ccd026:
[INFO] DOCKER> 1.869 Plugin workflow-cps:2692.v76b_089ccd026 unable to find dependant plugin ace-editor in update center https://updates.jenkins.io/update-center.json
```

## DOD
- [x] Jenkins Controller image build is succeeding

## Test Strategy/Script
1. `mvn clean install -pl :aissemble-jenkins-controller`
2. Verify build succeeds

epwilkins:
Updates the images with the longest build-times to layer in a mounted Docker cache for apt, poetry, and pip.  Also updates the Spark jars update logic to follow the `aissemble-hive-service` pattern of downloading the updated jars during the Maven build to utilize the m2 cache instead of using `wget` during the image build itself.

Updated build workflow to use the cache key calculated at start, since the generated `target` directories throw off the hash calculation.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -112,18 +112,18 @@ jobs:
         if: ${{ !cancelled() && steps.cached-docker-build.outputs.cache-hit != 'true' }}
         with:
           path: ~/.docker/cache
-          key: docker-cache-${{ hashFiles('**/Dockerfile') }}
+          key: ${{ steps.cached-docker-build.outputs.cache-primary-key }}
       - name: Save m2 repository cache
         id: save-m2-repo
         uses: runs-on/cache/save@v4
         if: ${{ !cancelled() && steps.cached-m2-repo.outputs.cache-hit != 'true' }}
         with:
           path: ~/.m2/repository
-          key: maven-${{ hashFiles('**/pom.xml') }}
+          key: ${{ steps.cached-m2-repo.outputs.cache-primary-key }}
       - name: Save m2 build cache
         id: save-m2-build
         uses: runs-on/cache/save@v4
         if: ${{ !cancelled() && steps.cached-m2-build.outputs.cache-hit != 'true' }}
         with:
           path: ~/.m2/build-cache
-          key: maven-build-cache-${{ hashFiles('**/pom.xml') }}
+          key: ${{ steps.cached-m2-build.outputs.cache-primary-key }}
--- a/extensions/extensions-docker/aissemble-hive-service/src/main/resources/docker/Dockerfile
+++ b/extensions/extensions-docker/aissemble-hive-service/src/main/resources/docker/Dockerfile
@@ -28,8 +28,8 @@ RUN --mount=source=${BIN_DIR},target=/tmp/bin cd /tmp \
 
 # Update library jars used by Hive and Hadoop
 COPY --chmod=755 ./src/main/resources/scripts/setup.sh $HIVE_HOME/setup.sh
-ADD ${PATCH_DIR}/* /tmp/patch-jars
-RUN $HIVE_HOME/setup.sh /tmp/patch-jars && rm -rf /tmp/patch-jars
+RUN --mount=target=/tmp/patch-jars,type=bind,source=${PATCH_DIR} \
+    $HIVE_HOME/setup.sh /tmp/patch-jars
 
 # Hadoop ships with source jars, which aren't necessary and bloat the image, also remove YARN support as we only support K8s
 RUN rm -rf $HADOOP_HOME/share/hadoop/common/sources \
--- a/extensions/extensions-docker/aissemble-jenkins/aissemble-jenkins-agent/src/main/resources/docker/Dockerfile
+++ b/extensions/extensions-docker/aissemble-jenkins/aissemble-jenkins-agent/src/main/resources/docker/Dockerfile
@@ -2,9 +2,11 @@ FROM docker.io/jenkins/ssh-agent:latest-debian-jdk17
 
 LABEL org.opencontainers.image.source="https://github.com/boozallen/aissemble"
 
-USER root
-RUN apt --yes update && apt install --yes pipx && pipx ensurepath && apt clean
-RUN pipx install poetry
+RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
+    --mount=type=cache,target=/var/lib/apt,sharing=locked \
+    apt --yes update && apt install --yes pipx && pipx ensurepath
+RUN --mount=type=cache,target=/root/.cache/pip \
+    pipx install poetry
 
 RUN mkdir /var/pyenv
 RUN ls -l -a /var/pyenv
--- a/extensions/extensions-docker/aissemble-spark/pom.xml
+++ b/extensions/extensions-docker/aissemble-spark/pom.xml
@@ -14,49 +14,296 @@
 
     <packaging>docker-build</packaging>
 
+    <properties>
+        <dockerbuild.patch.directory>target/dockerbuild/patch</dockerbuild.patch.directory>
+        <version.netty>4.1.118.Final</version.netty>
+        <version.avro>1.11.4</version.avro>
+        <version.derby>10.16.1.1</version.derby>
+        <version.hive.client>2.3.10</version.hive.client>
+        <version.parquet>1.15.0</version.parquet>
+        <version.zookeeper>3.9.3</version.zookeeper>
+    </properties>
+
     <dependencies>
         <dependency>
-            <groupId>com.boozallen.aissemble</groupId>
-            <artifactId>foundation-messaging-python-service</artifactId>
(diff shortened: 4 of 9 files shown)


extensions/extensions-docker/aissemble-hive-service/src/main/resources/docker/Dockerfile
```diff
@@ -28,8 +28,8 @@ RUN --mount=source=${BIN_DIR},target=/tmp/bin cd /tmp \
 
 # Update library jars used by Hive and Hadoop
 COPY --chmod=755 ./src/main/resources/scripts/setup.sh $HIVE_HOME/setup.sh
-ADD ${PATCH_DIR}/* /tmp/patch-jars
-RUN $HIVE_HOME/setup.sh /tmp/patch-jars && rm -rf /tmp/patch-jars
+RUN --mount=target=/tmp/patch-jars,type=bind,source=${PATCH_DIR} \
```
epwilkins, line 31: I: This results in the exact same output image, but reduces CPU churn since we don't have to `rm -rf`.

extensions/extensions-docker/aissemble-spark/pom.xml
```diff
@@ -14,56 +14,296 @@
...
     </dependencies>
 
     <build>
         <plugins>
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-dependency-plugin</artifactId>
                 <executions>
                     <execution>
-                        <id>copy-service-jar</id>
```
epwilkins, line 41: I: the previous copy execution outputs were not being used during the docker build at all.

extensions/extensions-docker/aissemble-spark/src/main/resources/scripts/setup.sh
```diff
@@ -10,54 +10,77 @@
...
+## Patches a lib directory with given updated JARs by matching on base name. For example, given the replace directory has
+## the jar `commons-math-3.4.jar` and the target directory has `commons-math-2.1.jar`, the existing 2.1 JAR is deleted
+## and the updated 3.4 jar is copied into the directory.
 ##
-## @Arguments: list of maven coordinates in `group:artifact:version:classifier` format where `:classifier` is optional
+## @Param: location of replacement JARs
+## @Param: location to search for JARs to replace
 #---
-update_maven_jars() {
+update_jars() {
```
epwilkins, line 34: I: taken from `extensions/extensions-docker/aissemble-hive-service/src/main/resources/scripts/setup.sh`

extensions/extensions-docker/aissemble-spark/src/main/resources/scripts/setup.sh
```diff
@@ -102,52 +125,12 @@ setup(
...
 
-SPARK_JARS="$1/jars"
-SPARK_VERSION=$2
-HADOOP_VERSION=$3
+PATCH_JARS="$1"
+SPARK_JARS="$2/jars"
+SPARK_VERSION=$3
+HADOOP_VERSION=$4
 
-update_maven_jars "com.google.code.gson:gson:2.8.9 \
```
epwilkins, line 109: I: These were all moved to the POM as dependencies to be staged with the maven-dependency-plugin, ~except `derbytools` and `derbyshared`, as it turns out those jars were not present on the image so were not required to patch.~

`derbytools` and `derbyshared` add back some classes that were extracted in the newer releases of Derby. Without them, the `EmbeddedDriver` is not available.

extensions/extensions-docker/aissemble-versioning/pom.xml
```diff
@@ -19,6 +19,9 @@
             <plugin>
                 <groupId>org.technologybrewery.habushu</groupId>
                 <artifactId>habushu-maven-plugin</artifactId>
+                <configuration>
+                  <updateDockerfile>false</updateDockerfile>
```
epwilkins, line 23: I: Because we're adding the pip & poetry cache to the generated code, we need to turn off injection.  If this approach works without drawbacks, we can roll it into Habushu and remove this config override.

extensions/extensions-docker/aissemble-jenkins/aissemble-jenkins-agent/src/main/resources/docker/Dockerfile
```diff
@@ -2,9 +2,11 @@ FROM docker.io/jenkins/ssh-agent:latest-debian-jdk17
 
 LABEL org.opencontainers.image.source="https://github.com/boozallen/aissemble"
 
-USER root
-RUN apt --yes update && apt install --yes pipx && pipx ensurepath && apt clean
-RUN pipx install poetry
+RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
+    --mount=type=cache,target=/var/lib/apt,sharing=locked \
+    apt --yes update && apt install --yes pipx && pipx ensurepath
+RUN --mount=type=cache,target=/root/.cache/pip \
```
csun-cpointe, line 8: Q: why do we specify the `type=cache` here vs `type=bind` in `aissemble-hive-service` image

csun-cpointe (approved):

.github/workflows/build.yml
```diff
@@ -112,18 +112,18 @@ jobs:
         if: ${{ !cancelled() && steps.cached-docker-build.outputs.cache-hit != 'true' }}
         with:
           path: ~/.docker/cache
-          key: docker-cache-${{ hashFiles('**/Dockerfile') }}
+          key: ${{ steps.cached-docker-build.outputs.cache-primary-key }}
```
csun-cpointe, line 115: Q: is this `cache-primary-key` specified somewhere? didn't seem find it in the exist files
```
- name: Load docker build cache
        id: cached-docker-build
        uses: runs-on/cache/restore@v4
        if: ${{ ! github.event.schedule }}
        with:
          path: ~/.docker/cache
          key: docker-cache-${{ hashFiles('**/Dockerfile') }}
          restore-keys: |
            docker-cache-
```

extensions/extensions-docker/aissemble-spark/pom.xml
```diff
@@ -14,56 +14,296 @@
...
+        <dependency>
+            <groupId>org.apache.hive.shims</groupId>
+            <artifactId>hive-shims-common</artifactId>
+            <version>2.3.10</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hive.shims</groupId>
+            <artifactId>hive-shims-scheduler</artifactId>
+            <version>2.3.10</version>
```
csun-cpointe: Q: should we move these hive-shim related versions to the build-parent pom file? seems a lot of repeating version `2.3.10`

extensions/extensions-docker/aissemble-spark/pom.xml
```diff
@@ -14,56 +14,296 @@
...
+        <dependency>
+            <groupId>org.apache.parquet</groupId>
+            <artifactId>parquet-hadoop</artifactId>
+            <version>1.15.0</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.parquet</groupId>
+            <artifactId>parquet-common</artifactId>
+            <version>1.15.0</version>
```
csun-cpointe: same here

.github/workflows/build.yml
epwilkins: It's set to the computed value of `key` in the `cache/restore` step. See https://github.com/actions/cache/tree/main/save#case-1---where-a-user-would-want-to-reuse-the-key-as-it-is

extensions/extensions-docker/aissemble-spark/pom.xml
epwilkins: I'm a little reluctant to pull them up to the build-parent as they're ostensibly temporary until Spark 4 comes out. I could pull them into properties in this pom file though, to de-duplicate them.

extensions/extensions-docker/aissemble-jenkins/aissemble-jenkins-agent/src/main/resources/docker/Dockerfile
epwilkins: `bind` is just a read-only volume mount for building without using `COPY` or `ADD`. `cache` is a special mount type that's saved by docker and re-used between builds.

csun-cpointe: thank you for the explanation!

PR #644, opened 2025-04-03
[#642] various build optimizations

### LINKED ISSUE #642: Feature: Improve build reliability and speed
## Description
In #606 we added logic to enable a Docker builder with the Kubernetes driver in order to perform native multi-arch builds.  However, the approach there was flawed as the `extensions-docker` project is responsible for creating the builder, but usually only runs during the nightly build and hits the Maven build cache for subsequent builds. This means subsequent builds rely on the ARM GH builder to virtualize x86 archs to do the build using a container driver.  This is _extremely_ slow.  We should move the logic for creating the builder to the GH build workflow file instead. Additionally optimize the build by looking for other places where we're doing more work than needed.

## DOD
- [x] Move docker builder creation from the _extensions/extensions-docker/pom.xml_ to _.github/workflows/build.yml_
- [x] Reduce docker cache from max to min
- [x] Use the same builder for the build phase as the push phase
- [x] Only build the image during the build phase if the image has tests
- [x] Use `generate-sources` instead of `install` to do the final archetype test check for manual actions
- [x] Only save the Docker cache when it changes by calculating a hash from the contents

## Test Strategy/Script
1. Verify the latest build succeeded
    - on `improve-build` for OTS and `dev` for final test 
2. Verify that the build has steps for provisioning and tearing down the docker builder
3. Verify the docker builds say they are using the k8s driver by searching the build log for `building with "k8s-multiarch" instance using kubernetes driver`
4. Verify the `default-build` goal for the following modules actually built the image for the tests
    - `aissemble-fast-api`
    - `aissemble-hive-service`
    - `aissemble-spark`
    - `aissemble-spark-infrastructure`
    - `aissemble-pipeline-invocation`
5. Verify the docker cache save is skipped if there is no docker build but the maven caches are saved
 
Kubernetes Driver Docs: https://docs.docker.com/build/builders/drivers/kubernetes/
Approach for hashing files outside the working directory: https://github.com/actions/toolkit/issues/1035#issuecomment-2070390607
GH-script Action: https://github.com/actions/github-script

epwilkins:
### Docker Improvements:
Moves the provisioning and teardown of the k8s docker builder to the GH action to avoid the Maven build cache skipping the logic. Reduces the cache level to the default `min` instead of max, as it _appears_ that the layer cache isn't really working with the k8s builder, just the base image cache. Skips the image build in CI by default (like the previous behavior) with images that have tests specifically turning the image build back on to ensure the tests are run against the latest image.

### Archetype Improvements:
Uses generate-sources for final check in archetype test. We still do a final build to verify the project builds successfully, but the check for manual actions is done separately to avoid a full rebuild. This should save about 9-10 minutes.

### GH Cache Improvements:
Uses the `github-script` action to call the same hashing function used by the `runs-on/cache` action to detect whether caches actually had their contents modified, regardless of cache-hit. This saves time when some POM or Dockerfile change did not actually result in new dependencies, or new layers. Because of the inherent limitations of the way the gh runner logic parses [`hashFiles`](https://github.com/actions/runner/blob/c1095ae2d100886a7865f466357f09ef30dec1b9/src/Runner.Worker/Expressions/HashFilesFunction.cs#L44) we aren't able to directly call `hashFiles` with a [different working directory](https://github.com/actions/toolkit/issues/1035#issuecomment-1678272606) so instead, we use the `github-script` action to call `glob.hashFiles` directly.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,6 +27,7 @@ jobs:
     env:
       DOCKER_CONFIG: /home/runner/.docker
       RUNS_ON_S3_BUCKET_CACHE: aissemble-github-cache
+      DOCKER_BUILDER_NAME: k8s-multiarch
     steps:
       - uses: actions/checkout@v4
         with:
@@ -61,18 +62,14 @@ jobs:
         if: ${{ ! github.event.schedule }}
         with:
           path: ~/.m2/repository
-          key: maven-${{ hashFiles('**/pom.xml') }}
-          restore-keys: |
-            maven-
+          key: maven-repo-cache
       - name: Load m2 build cache
         id: cached-m2-build
         uses: runs-on/cache/restore@v4
         if: ${{ ! github.event.schedule }}
         with:
           path: ~/.m2/build-cache
-          key: maven-build-cache-${{ hashFiles('**/pom.xml') }}
-          restore-keys: |
-            maven-build-cache-
+          key: maven-build-cache
       #NB: Not saving poetry cache on failure in case it's a failure caused by an in-flight python package release
       - name: Poetry cache
         id: cached-poetry
@@ -83,6 +80,26 @@ jobs:
           key: poetry-cache-${{ hashFiles('**/pyproject.toml') }}
           restore-keys: |
             poetry-cache-
+      - name: Hash starting Docker cache
+        id: docker-cache-start
+        uses: actions/github-script@v7
+        with:
+          result-encoding: string
+          script: await glob.hashFiles(`~/.docker/cache/**`, require('os').homedir())
+      - name: Provision Docker builder
+        id: provision-docker-builder
+        run: |
+          docker builder create --name=$DOCKER_BUILDER_NAME \
+                                --bootstrap \
+                                --driver=kubernetes \
+                                --platform=linux/amd64 \
+                                --driver-opt="nodeselector=kubernetes.io/arch=amd64","image=docker.io/moby/buildkit:v0.19.0"
+          docker builder create --name=$DOCKER_BUILDER_NAME \
+                                --append \
+                                --bootstrap \
+                                --driver=kubernetes \
+                                --platform=linux/arm64 \
+                                --driver-opt="nodeselector=kubernetes.io/arch=arm64","image=docker.io/moby/buildkit:v0.19.0"
       # Generate the settings.xml for ghcr.io, pypi, & dev-pypi server profiles
       - name: Create settings.xml
         id: create-settings-xml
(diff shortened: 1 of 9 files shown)


.github/workflows/build.yml
```diff
@@ -83,6 +84,26 @@ jobs:
           key: poetry-cache-${{ hashFiles('**/pyproject.toml') }}
           restore-keys: |
             poetry-cache-
+      - name: Hash starting Docker cache
+        id: docker-cache-start
+        uses: actions/github-script@v7
+        with:
+          result-encoding: string
+          script: await glob.hashFiles(`~/.docker/cache/**`, require('os').homedir())
```
epwilkins, line 88: I: '~' is only interpreted in the search glob patterns, so have to use `os` for the working directory

.github/workflows/build.yml
```diff
@@ -83,6 +84,26 @@ jobs:
...
+                                --bootstrap \
+                                --driver=kubernetes \
+                                --platform=linux/amd64 \
+                                --driver-opt="nodeselector=kubernetes.io/arch=amd64","image=docker.io/moby/buildkit:v0.19.0"
+          docker builder create --name=$DOCKER_BUILDER_NAME \
+                                --append \
+                                --bootstrap \
+                                --driver=kubernetes \
+                                --platform=linux/amd64 \
+                                --driver-opt="nodeselector=kubernetes.io/arch=arm64","image=docker.io/moby/buildkit:v0.19.0"
```
epwilkins, line 102: I: These commands are identical to those removed from the `exec-maven-plugin` executions in `extensions-docker/pom.xml`

.github/workflows/build.yml
```diff
@@ -61,7 +62,7 @@ jobs:
         if: ${{ ! github.event.schedule }}
         with:
           path: ~/.m2/repository
-          key: maven-${{ hashFiles('**/pom.xml') }}
+          key: maven-repo-cache
           restore-keys: |
```
epwilkins: A: Restore keys are no longer needed with static cache key

extensions/extensions-docker/pom.xml
```diff
@@ -150,78 +150,36 @@
         <profile>
             <id>gh-build</id>
             <properties>
-                <docker.builder.name>k8s-multiarch</docker.builder.name>
-                <!-- Ideally, we'd pass this in with `config` flag while provisioning the builder, but it doesn't appear to be working. -->
+                <docker.builder.name>${env.DOCKER_BUILDER_NAME}</docker.builder.name>
```
epwilkins, line 153: I: Retrieve docker builder from env var set in GH action. The deleted comment was outdated even before this change.

extensions/extensions-docker/pom.xml
```diff
@@ -150,78 +150,36 @@
...
                         <plugin>
                             <groupId>${group.fabric8.plugin}</groupId>
                             <artifactId>docker-maven-plugin</artifactId>
+                            <configuration>
+                                <images>
+                                    <image>
+                                        <build>
+                                            <buildx>
+                                                <builderName>${docker.builder.name}</builderName>
+                                                <dockerStateDir>${docker.builder.stateDir}</dockerStateDir>
```
epwilkins, line 168: I: Using the same builder for the `default-build` on images with ITs means we can re-use the cache during `default-deploy` for the native architecture.

foundation/foundation-archetype/test-project-archetype.sh
```diff
@@ -308,13 +312,12 @@ cd ../..
 echo -e "\nINFO: Generating project structure for ml pipelines\n"
 runBuildAndUpdateDeploy
 
-echo -e "\nINFO: updating the test-generator-deploy/pom.xml based on static code as it's hard to predict\n"
-plugins=$(esc < ../../../plugins-to-add-to-archetype_test-generator-deploy_pom.xml)
-sub "s/ *<\/plugins>/$plugins        <\/plugins>/" test-generator-deploy/pom.xml
+echo -e "\nINFO: Running final build to ensure success"
+./mvnw clean install
 
 echo -e "\nINFO: Running full build to check that the build passes w/o any manual actions needed\n"
```
epwilkins: A: Reword, as the check for build success happens in the previous build. This is just for manual actions

foundation/foundation-archetype/test-project-archetype.sh
```diff
@@ -308,13 +312,12 @@ cd ../..
 echo -e "\nINFO: Generating project structure for ml pipelines\n"
 runBuildAndUpdateDeploy
 
-echo -e "\nINFO: updating the test-generator-deploy/pom.xml based on static code as it's hard to predict\n"
-plugins=$(esc < ../../../plugins-to-add-to-archetype_test-generator-deploy_pom.xml)
-sub "s/ *<\/plugins>/$plugins        <\/plugins>/" test-generator-deploy/pom.xml
+echo -e "\nINFO: Running final build to ensure success"
+./mvnw clean install
```
epwilkins: C: Exit with non-zero code if build fails
```suggestion
./mvnw clean install || { echo -e '\n\n\t**** MAVEN BUILD FAILED ****\n\n' ; exit 1; }
```

.github/workflows/build.yml
```diff
@@ -83,6 +84,26 @@ jobs:
...
+          docker builder create --name=$DOCKER_BUILDER_NAME \
+                                --bootstrap \
+                                --driver=kubernetes \
+                                --platform=linux/amd64 \
+                                --driver-opt="nodeselector=kubernetes.io/arch=amd64","image=docker.io/moby/buildkit:v0.19.0"
+          docker builder create --name=$DOCKER_BUILDER_NAME \
+                                --append \
+                                --bootstrap \
+                                --driver=kubernetes \
+                                --platform=linux/amd64 \
```
aaron-gary: Q:  I guess the mismatch between platform and driver-opt (amd - arm) is just saying the machine it's building on is amd, but the image will be arm?  In the pom.xml below the platform matched the driver-opt.

csun-cpointe (approved):

.github/workflows/build.yml
epwilkins: uhhh... no they should match. Now I'm curious how this ever worked in the test builds. 🤔

epwilkins: I think I just got lucky that the amd images chose the actual amd node, and I guess because the native arch was arm it built arm on the local pod since there wasn't a matching builder node platform?

Either way, this needs to be fixed before merge.

PR #646, opened 2025-04-10
[#641] Add helmfile to archetype and manual actions

### LINKED ISSUE #641: Feature: Integrate Helmfile into aiSSEMBLE archetype generation and manual acitons
## Description
We want to remove Tilt as our generated deployment process because it differs to much from our CI deployment. We want to move to using Helmfile instead.

## DOD
- [x] The archetype created new projects with basic helmfiles
- [x] The helmfile has environments pointing to base, dev, and ci values files
  - [x] The dev env will be default to make initial adoption easier
- [x] When building the project, manual actions are displayed to add releases to the helmfile
- [x] Baton migrations are created to
  - [x] Generates the base helmfile and adds itself to the baton ignore list so its only ran once
  - [x] Updated the aissemble version. Similar to the tiltfileMigration.
- [x] Update release notes draft for helmfile support. what and why
- [x] The archetype tests have been updated accordingly. 
- [x] The unused appDependencies is removed

### Feature Scenarios
```feature
Feature: Helmfile manual actions are displayed to users

  Scenario Outline: Helmfile manual actions are displayed for deployments
    Given a project with name "example-helmfile" exists with a data-delivery pipeline
    When the "<profile>" is generated
    Then the manual action to add the helmfile release is displayed

    Examples:
      | profile                                |
      | spark-infrastructure-deploy-v2         |
      | spark-operator-deploy-v2               |
      | inference-deploy-v2                    |
      | mlflow-deploy-v2                       |
      | policy-decision-point-deploy-v2        |
      | configuration-store-deploy-v2          |
      | aissemble-shared-infrastructure-deploy |

  Scenario: Helmfile manual actions are generated for docker-spark-python-pipelines
    Given a project with name "example-helmfile" exists with a data-delivery pipeline
    When the "docker-spark-python-pipelines" is generated
    Then the manual action to add the helmfile release for the pipeline is displayed
```
### Testing Script
Manual Actions and Archetype Test Steps:
* Generate a downstream project using aissemble 1.13.0-SNAPSHOT 
* Verify the helmfily.yaml is generated. Verify its contents contain the environments and helmdefaults
* Add a spark/pyspark pipeline into the MDA
* Build the project and start resolving the manual actions
* Verify the helmfile manual actions to add the deployment and spark pipeline releases are displayed
* Once all manual actions are completed, deploy with `helmfile apply`
* Verify all the applications **except** the pipeline comes up successfully
* Deploy the pipeline with `helmfile apply --state-values-set pipeline=<your-pipeline-name>`
* Verify the pipeline comes up successfully

Baton Migrations Test Steps
* Generate a downstream project using aissemble 1.12.1 
* Add a pipeline, build, and resolve all manual actions
* Git init the project with `git init`
* Update the aissemble build-parent to 1.13.0-SNAPSHOT
* Run the migrations with `mvn org.technologybrewery.baton:baton-maven-plugin:baton-migrate`
* Verify the helmfile is generated and the `helmfile-generation-migration` is added to the deactivated list
* Revert the changes
* Create a helmfile - `touch helmfile.yaml`
* Re-run the migrations
* Verify the helmfile was **not** modified

cwoods-cpointe:
Initial effort to leverage helmfile as the default build tool. Update the archetype build to generate the initial helmfile. Create helmfile manual actions to add releases.

Created migration to update helmfile aissemble version. Create migration to generate an initial helmfile.yaml for projects that dont have one. Migration deactivates self after running so users are not forced to adopt helmfile.

Remove unused appDependencies in notifications

Updated the archetype test to account for new helmfile actions


foundation/foundation-archetype/src/main/resources/archetype-resources/__rootArtifactId__-deploy/src/main/resources/values-ci-helmfile.yaml
```diff
@@ -0,0 +1,7 @@
+namespace: ${parentArtifactId}
```
cwoods-cpointe: i: The `values-helmfile` naming convention will likely be changed when we remove ArgoCDs values files.

foundation/foundation-archetype/test-project-archetype.sh
```diff
@@ -312,6 +337,9 @@ cd ../..
 echo -e "\nINFO: Generating project structure for ml pipelines\n"
 runBuildAndUpdateDeploy
 
+echo -e "\nINFO: Building to generate helmfile manual actions\n"
+runBuildAndUpdateHelmfile
```
cwoods-cpointe: i: This solution requires a re-build if the project after everything is added in order to get all the helmfile notifications. Unfortunately this does add another non-cached build

foundation/foundation-mda/src/main/resources/templates/notifications/notification.spark.worker.helmfile.vm
```diff
@@ -0,0 +1,24 @@
+#set($path = "/src/${pythonPipelineArtifactId}/resources/apps/")
+#if(${pipelineImplementation} == "data-delivery-spark")
+  #set($path = "/src/main/resources/apps/")
+#end
+  {{- if eq .Values.pipeline "${pipelineArtifactId}" }}
```
cwoods-cpointe, line 5: i: I played around with using the much cleaner `condition` instead of an if. Unfortunately myself an others have noticed the conditions are not updated by the `--state-values-set` so users can not override the enabled value by command line.

DRAFT_RELEASE_NOTES.md
```diff
@@ -1,5 +1,12 @@
 # Major Additions
 
+## Helmfile Integration
+Currently, aiSSEMBLE provides automation for the path to production tool Tilt and ArgoCD. Having the local and
```
epwilkins: A: GH release notes don't play nicely with manual line breaks. Better to leave it all on one line and let GH break the lines.  It's ugly for us as editors, but since this is a temp draft it's not the end of the world.

aaron-gary (approved):

DRAFT_RELEASE_NOTES.md
```diff
@@ -44,6 +51,8 @@ To reduce burden of upgrading aiSSEMBLE, the Baton project is used to automate t
 | upgrade-v1-chart-files-aissemble-version-migration | Updates the docker image tags within your project's deployment resources (`<YOUR_PROJECT>-deploy/src/main/resources/apps/`) to use the latest version of the aiSSEMBLE       |
 | data-encryption-removal-pom-migration              | Remove the data encryption dependencies from the pom file                                                                                                                    |
 | data-encryption-removal-pyproject-migration        | Remove the data encryption dependencies from the pyproject.toml file                                                                                                         |
+| helmfile-generation-migration                      | Generates an initial helmfile.yaml for manual actions to be added to. After its ran once, it will automatically be added to the deactivated list                             |
+| upgrade-helmfile-aissemble-version-migration       | Updates the aiSSEMBLE version in the deployment values file. This version is used by the helmfile.yaml                                                                       |
```
epwilkins: A: Because this always bugged me about the original tiltfile migration name
```suggestion
| helmfile-aissemble-version-migration               | Updates the aiSSEMBLE version in the deployment values file. This version is used by the helmfile.yaml                                                                       |
```

DRAFT_RELEASE_NOTES.md
```diff
@@ -117,6 +126,27 @@ values.yaml file to enable configuration store access vault:
 +         secrets.unseal.keys==key1,key2,key3
 ```
 
+## For projects wanting to leverage Helmfile
```
epwilkins: Q: Is there anyway to embed the default for this within the initial Helmfile? If not, can we generate this file as well during the migration?

foundation/foundation-archetype/src/main/resources/archetype-resources/helmfile.yaml
```diff
@@ -0,0 +1,30 @@
...
+      - ${artifactId}-deploy/src/main/resources/values-dev-helmfile.yaml
+  ci:
+    values:
+      - ${artifactId}-deploy/src/main/resources/values-helmfile.yaml
+      - ${artifactId}-deploy/src/main/resources/values-ci-helmfile.yaml
+    kubeContext: PLACEHOLDER-kubernetes-ci-context
+
+helmDefaults:
+  {{ if .Values.helm.skipDeps }}
+  skipDeps: true
```
epwilkins: A: I'd just hard-code this.  99% of the time it will be more correct to let the Maven build update the dependencies for helm, so supporting this edge case complicates the Helmfile while not providing value for most projects.

foundation/foundation-archetype/src/main/resources/archetype-resources/helmfile.yaml
epwilkins, line 1: A: Consider adding a README to the archetype with info on how the Helmfile works. Not duplicating any docs from Helmfile, just things that are specific to how aiSSEMBLE uses Helmfile - e.g. the values file options

foundation/foundation-archetype/src/main/resources/archetype-resources/helmfile.yaml
```diff
@@ -0,0 +1,30 @@
...
+  {{- end }}
+
+releases:
+  - name: kafka-cluster
```
epwilkins: Q: If we left this out would a manual action tell you to add it? If so we should drop it.  While we haven't managed to do it yet, we've long talked about removing kafka as the default messaging implementation.

foundation/foundation-archetype/test-project-archetype.sh
```diff
@@ -113,6 +113,31 @@ function runBuildAndUpdateDeploy {
   fi
 }
 
+#---
+## Runs maven build and captures any manual actions for adding releases to the helmfile, then updates the
+## helmfile based on the detected manual actions.
+#---
+function runBuildAndUpdateHelmfile {
```
epwilkins: S: You should be able to update `runBuildAndUpdateDeploy` by just piping the `tee` output to your `tee` command. E.g.:
```bash
  ./mvnw clean install | \
    tee >(awk "BEGIN {output=0} /$outputend/ {output=0} NF && output {print} /$outputstart/ {output=1}">deploy.out ) \
    tee >(awk "BEGIN {output=0} /$outputend/ {output=0} NF && output {print} /$outputstart/ {output=1}">helmfile.out ) \
    || { echo -e '\n\n\t**** MAVEN BUILD FAILED ****\n\n' ; exit 1; }
  execs=$(cat deploy.out)
  releases=$(cat helmfile.out)

Comment (epwilkins) [reply foundation/foundation-archetype/test-project-archetype.sh]: Oh. You'd have to make the `outputstart` and `outputend` variables unique. Forgot that bit. If this works you could just rename it to `runBuildAndApplyManualActions` or something like that.

File: foundation/foundation-archetype/src/main/resources/archetype-resources/__rootArtifactId__-deploy/src/main/resources/values-helmfile.yaml [hunk 10]
```diff
@@ -0,0 +1,9 @@
+namespace: default
```
epwilkins: A: Default to project name. They are limited to 63 chars, so if we had a fancy way to truncate that'd be nice. But the error message for exceeding the limit would probably be good enough to guide users to the resolution.

foundation/foundation-mda/src/main/resources/templates/notifications/notification.helm.helmfile.vm
```diff
@@ -0,0 +1,19 @@
+  - name: ${appName}
+    namespace: {{ .Values.namespace }}
```
epwilkins: S: Some resources will need to be deployed to dedicated namespaces, i.e. the config-store.

epwilkins: Not sure if this is over-complicating it, but the helmfile values file could have something like:
```yaml
namespaces:
  primary: 'project-name'
  configStore: 'project-name-config'
```
And the namespace could be set like:
```yaml
namespace: {{ .Values.namespace.primary | default .Release.Namespace }}`
```

But honestly, since this is a single project and it's unlikely that they'll want to change the namespace name very often, maybe you just hard-code the string name in `Helmfile`.

foundation/foundation-mda/src/main/resources/templates/notifications/notification.resource.tilt.vm
epwilkins, line 1: Q: Is this template even used? I think it can be removed.

foundation/foundation-archetype/src/main/resources/archetype-resources/__rootArtifactId__-deploy/src/main/resources/values-helmfile.yaml
ewilkins-csi: If we can inline the helmfile values files, this may be unnecessary

foundation/foundation-archetype/src/main/resources/archetype-resources/helmfile.yaml
epwilkins: If we can inline the helmfile values files, this may be unnecessary. Though, we will need to tell them how to execute a specific pipeline somehow.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/HelmFileVersionMigration.java
```diff
@@ -0,0 +1,86 @@
...
+
+        try (BufferedReader valuesFile = new BufferedReader((new FileReader(file)))) {
+            String line;
+
+            while ((line = valuesFile.readLine()) != null) {
+                if (line.matches(VERSION_REGEX)) {
+                    String currentAissembleVersion = line.substring(line.lastIndexOf(":") + 1);
+                    currentAissembleVersion = StringUtils.deleteWhitespace(currentAissembleVersion);
+
+                    if (isLessThanVersion(currentAissembleVersion, getAissembleVersion())) {
```
epwilkins: Q/S: Did we just copy this check from the existing Tiltfile migration, and/or do you know why this check is necessary? I'd think if a project is downgrading intentionally, that it would make sense to align the version in the deployment file as well.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/HelmFileVersionMigration.java
```diff
@@ -0,0 +1,86 @@
...
+import org.slf4j.LoggerFactory;
+import org.technologybrewery.baton.BatonException;
+
+/**
+ * Migration to update the aissemble version used in the helmfile
+ */
+public class HelmFileVersionMigration extends AbstractAissembleMigration {
+
+    private static final Logger logger = LoggerFactory.getLogger(HelmFileVersionMigration.class);
+    private static final String VERSION_REGEX = "(aissembleVersion: )(\\d+\\.\\d+\\.\\d+(?:[\\.\\-\\d+a-zA-Z]*))";
```
epwilkins, line 36: A: Recommend compiling this pattern since you'll be calling it a lot. `Pattern.compile(regex)`. Patterns also have `asMatchPredicate` which can be handy.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/HelmFileVersionMigration.java
```diff
@@ -0,0 +1,86 @@
...
+     * Migration will update the aissembleVersion in the helmfile values charts
+     * @param file helmfile values file to migrate
+     * @return true if migration was successfully performed
+     */
+    @Override
+    protected boolean performMigration(File file) {
+        boolean isMigrated;
+
+        // Update the aissemble to the desired version
+        try (BufferedReader helmfileValues = new BufferedReader((new FileReader(file)))) {
```
epwilkins: S: I don't think `helmfileValues` is used? Looks like we're just passing the `file` directly to the util method.

DRAFT_RELEASE_NOTES.md
cwoods-cpointe: We should be able to put these in place of the values file path in the helmfile env. I will test that out. I think that solution is a bit messier for the helmfile but it contains the full working solution without needing this manual change. If users want to split it out to separate files like we generate then they can.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/version_specific/HelmfileGenerationMigration.java
```diff
@@ -0,0 +1,206 @@
...
+     *
+     * @param file pom file to check
+     * @return true if the migration should run
+     */
+    @Override
+    protected boolean shouldExecuteOnFile(File file) {
+        Model model = PomHelper.getLocationAnnotatedModel(file);
+
+        // Check if the pom file is at the projects root level
+        return model.getParent() == null || AISSEMBLE_PARENT.equals(model.getParent().getArtifactId());
```
epwilkins, line 67: A: It's effectively the exact same logic, but for consistency consider:
```java
return getMavenProject().equals(getRootProject());
```
or even
```java
return getMavenProject() == getRootProject();
```

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/version_specific/HelmfileGenerationMigration.java
```diff
@@ -0,0 +1,206 @@
...
+            logger.info("Helmfile already found. Skipping creation and ignoring migration");
+        } else {
+            try {
+                InputStream initialHelmfile = getClass().getClassLoader().getResourceAsStream(HELMFILE_TEMPLATE_PATH);
+                if (initialHelmfile == null) {
+                    throw new BatonException("Could not find the helmfile template");
+                }
+                File helmfile = new File(helmfileLocation);
+                FileUtils.copyInputStreamToFile(initialHelmfile, helmfile);
+                updateArtifactId(helmfile);
```
epwilkins, line 95: A: I don't think I have a solution to this, but this will be painful for the one project we know of where the root artifact ID doesn't match the project prefix that's used for submodules.  Maybe just a note in the release notes somewhere?

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/version_specific/HelmfileGenerationMigration.java
```diff
@@ -0,0 +1,206 @@
...
+
+    }
+
+    private void updateArtifactId(File file) {
+        try (BufferedReader valuesFile = new BufferedReader((new FileReader(file)))) {
+            String line;
+
+            while ((line = valuesFile.readLine()) != null) {
+                if (line.matches(ARTIFACT_TAG_REGEX)) {
+                    String substitution = "$1" + getArtifactId() + "$3";
```
epwilkins: A: I know you're just using what's there, but this name is a little misleading. I'd think it returned the artifact ID of the current project, but it returns the artifact ID of the root project.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/version_specific/HelmfileGenerationMigration.java
```diff
@@ -0,0 +1,206 @@
...
+            try {
+                InputStream initialHelmfile = getClass().getClassLoader().getResourceAsStream(HELMFILE_TEMPLATE_PATH);
+                if (initialHelmfile == null) {
+                    throw new BatonException("Could not find the helmfile template");
+                }
+                File helmfile = new File(helmfileLocation);
+                FileUtils.copyInputStreamToFile(initialHelmfile, helmfile);
+                updateArtifactId(helmfile);
+            } catch (IOException | BatonException e) {
+                throw new BatonException("Failed to instantiate the helmfile.yaml", e);
```
epwilkins, line 97: A: Feels a little weird/noisy to have baton exceptions wrapping baton exceptions thrown from the same file. We could simply re-throw without wrapping, or re-throw without wrapping in the called methods.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/version_specific/HelmfileGenerationMigration.java
```diff
@@ -0,0 +1,206 @@
...
+            throw new BatonException("Unable to modify the helmfile's artifact ID.", e);
+        }
+    }
+
+    private boolean addMigrationToIgnoreList(File file) {
+        logger.info("Deactivating Helmfile generation migration.");
+        Model model = PomHelper.getLocationAnnotatedModel(file);
+        Plugin baton = model.getBuild().getPluginsAsMap().get(BATON_KEY);
+        if (baton == null) {
+            throw new BatonException("Failed to find Baton plugin configuration in root POM");
```
epwilkins, line 126: I: I know some projects have an "aissemble" part of their project, and a "non-aissemble" part of their project, which initially had me worried that on the umbrella project that held both parts we'd end up throwing this exception. But I actually think that if the parent isn't aissemble's `build-parent` and is `null`, then baton would only be running if it were configured. Just thought I'd leave my realization in case anyone else had pause.

epwilkins (approved):

foundation/foundation-archetype/src/main/resources/archetype-resources/helmfile.yaml
cwoods-cpointe: The manual actions request adding Kafka so Ill remove it from the archetype

foundation/foundation-mda/src/main/resources/templates/notifications/notification.resource.tilt.vm
cwoods-cpointe: I do not see it being used. Will remove

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/HelmFileVersionMigration.java
cwoods-cpointe: This regex is later used in Batons FileUtils.replaceInFile which does compile it to a pattern/matcher

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/version_specific/HelmfileGenerationMigration.java
cwoods-cpointe: `getRootProject` fails because baton doesnt call `setMavenProject()` until the migration is performed

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/HelmFileVersionMigration.java
epwilkins: I was thinking about in the `shouldExecutOnFile` method where we call `line.matches(regex)`

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/version_specific/HelmfileGenerationMigration.java
epwilkins: I'll have to dig into this some more. I thought the project was set as soon as the migration object was created.  I'd call it a bug in Baton if that's not the case.  And I guess it's not super easy to verify in unit tests since you'd be mocking the maven project anyway.

foundation/foundation-upgrade/src/test/java/com/boozallen/aissemble/upgrade/migration/extensions/HelmfileGenerationMigrationTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+import com.boozallen.aissemble.upgrade.migration.version_specific.HelmfileGenerationMigration;
+
+public class HelmfileGenerationMigrationTest extends HelmfileGenerationMigration {
+
+    private static final Path TEST_FILES_FOLDER = Paths.get("target", "test-classes", "test-files");
+
+    @Override
+    protected MavenProject getRootProject() {
+        MavenProject parentAissemlbeProject = new MavenProject();
```
epwilkins: A:
```suggestion
        MavenProject parentAissembleProject = new MavenProject();
```

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/HelmFileVersionMigration.java
cwoods-cpointe: Didnt think about the downgrading case. I can update it to check for equality instead of greater or less than

PR #665, opened 2025-04-22
[#653] set maxBuilds to 1 in build cache config

### LINKED ISSUE #653: Feature: Ensure Docker images do not become stale when using the Maven build cache
## Description
If the Maven Build cache is set to save anything more than the most recent build, there are situations where a Docker image can become stale and invalidate automated and manual testing.  Imagine in a Docker module with state `A` is built, then updated to state `B` and rebuilt.  The build cache will have a build state saved for both `A` and `B`. If the module is then reverted to state `A`, the cache-hit will prevent the Docker image build from re-triggering, and the developer/downstream ITs will assume the corresponding state `A` image is loaded when in fact the Docker daemon still has the `B` image loaded. This ended up causing a build failure to go undetected in #643.

## DOD
- [x] Locally, when a Docker module is restored from the Maven build cache, the associated Docker image is reloaded into the Docker daemon
  - If this proves difficult/impossible, we could just update the default cache config to only store 1 build as we did in #650 
  - [x] ~Ensure disk usage with this approach is reasonable~
    - Disk usage was way too high (5-6 GB per cached build) when using `docker:save`
  - [x] ~Ensure build of a fully cached build is not unreasonably slow~
    -  Build time more than tripled when using `docker:save`
- [x] Write Baton migration or upgrade instructions if the changes are too complex to automate
- [x] Update release notes

### Notes
 - We will tackle CI behavior in a follow-on (potentially when/as we address current issues with CI build-cache behavior for Python/Helm)
 - We could consider implementing this on Baseline first and roll out to downstream projects after we vet it -- if we do this, we should set the default number of builds to cache to 1 in `foundation-archetype`
   - Current thought is that since this is changing local behavior only, it's less risky to rollout without vetting heavily

## Test Strategy/Script
1. Create a project from the 1.12 archetype:
    ```sh
    mvn archetype:generate '-DarchetypeGroupId=com.boozallen.aissemble' \
                           '-DarchetypeArtifactId=foundation-archetype' \
                           '-DarchetypeVersion=1.12.1' \
                           '-DgroupId=org.test' \
                           '-Dpackage=org.test' \
                           '-DprojectGitUrl=test.org/test.git' \
                           '-DprojectName=Test Project' \
                           '-DartifactId=ots-653'\
    && cd ots-653
    ```
2. Verify that _.mvn/maven-build-cache-config.xml_ has a **maxBuilds** value of 5
3. Update the root _pom.xml_ to set the *parent/version* to **1.13.0-SNAPSHOT**
4. Run the migrations
    ```sh
    ./mvnw clean install
    ```
5. Verify that _.mvn/maven-build-cache-config.xml_ has a **maxBuilds** value of 1 with an explanatory comment above it

- Fabric8 docker-maven-plugin [`save`](https://dmp.fabric8.io/#docker:save) and [`load`](https://dmp.fabric8.io/#docker:build) options

epwilkins:
Also cleans up unnecessary attached outputs and ensures rebuilds are triggered if lock files change (i.e. an update lock file is checked in by someone else).

### DIFF
--- a/.mvn/maven-build-cache-config.xml
+++ b/.mvn/maven-build-cache-config.xml
@@ -37,20 +37,20 @@
         <attachedOutputs>
             <dirNames>
                 <dirName>../dist</dirName>
-                <dirName>quarkus-app</dirName>
                 <dirName>helm</dirName>
                 <dirName>../src/generated</dirName>
                 <dirName>../src/generated-test</dirName>
                 <!-- Needed for test-mda-models: -->
                 <dirName>../src/main/java</dirName>
                 <!-- Reports needed on CI to ensure consistent reporting of QA metrics: -->
+                <!-- Explicitly navigate to target since Habushu modules use dist as the project build directory -->
                 <dirName>../target/cucumber-reports</dirName>
                 <dirName>../target/surefire-reports</dirName>
                 <dirName>../target/failsafe-reports</dirName>
             </dirNames>
         </attachedOutputs>
         <local>
-            <!-- NB saving more than 1 build can cause issues because it does not restore the prior docker image to the docker daemon. -->
+            <!-- NB: saving more than 1 build can result in stale docker images, as build-cache restore does not restore images -->
             <maxBuildsCached>1</maxBuildsCached>
         </local>
         <projectVersioning adjustMetaInf="true"/>
--- a/foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/version_specific/MavenBuildCacheMigration.java
+++ b/foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/version_specific/MavenBuildCacheMigration.java
@@ -1,0 +1,63 @@
+package com.boozallen.aissemble.upgrade.migration.version_specific;
+
+/*-
+ * #%L
+ * aiSSEMBLE::Foundation::Upgrade
+ * %%
+ * Copyright (C) 2021 Booz Allen
+ * %%
+ * This software package is licensed under the Booz Allen Public License. All Rights Reserved.
+ * #L%
+ */
+
+import com.boozallen.aissemble.upgrade.migration.AbstractAissembleMigration;
+import org.technologybrewery.baton.BatonException;
+import org.technologybrewery.baton.util.FileUtils;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+
+/**
+ * Updates the `maxBuilds` property of the `maven-build-cache-config.xml` file to 1 with an explanatory comment in order
+ * to avoid issues with stale Docker images when restoring builds from the build cache.x
+ */
+public class MavenBuildCacheMigration extends AbstractAissembleMigration {
+    private static final Pattern MAX_BUILDS = Pattern.compile("( *)<maxBuildsCached>(.*?)</maxBuildsCached>");
+    private static final int MAX_BUILD_COUNT_CAPTURE = 2;
+
+    @Override
+    protected boolean shouldExecuteOnFile(File file) {
+        if (file.getName().endsWith(".xml")) {
+            try (Stream<String> lines = Files.lines(file.toPath())) {
+                return lines.map(MAX_BUILDS::matcher)
+                        .filter(Matcher::find)
+                        .findFirst()
+                        .map(matcher -> matcher.group(MAX_BUILD_COUNT_CAPTURE))
+                        .map(count -> !count.equals("1")).orElse(false);
+            } catch (IOException e) {
+                throw new BatonException("Failed to read build cache config", e);
+            }
+        }
+        return false;
+    }
+
+    @Override
+    protected boolean performMigration(File file) {
+        try {
+            return FileUtils.replaceInFile(file, MAX_BUILDS.pattern(), getMaxBuildsWithComment());
+        } catch (IOException e) {
+            throw new BatonException("Failed to update <maxBuilds> to 1 in: " + file.getPath(), e);
+        }
+    }
+
+    private String getMaxBuildsWithComment() {
+        // note: first capture group of MAX_BUILDS regex is the leading space
+        return """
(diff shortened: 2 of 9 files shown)


.mvn/maven-build-cache-config.xml
```diff
@@ -37,20 +37,19 @@
...
             <dirNames>
                 <dirName>../dist</dirName>
-                <dirName>quarkus-app</dirName>
                 <dirName>helm</dirName>
                 <dirName>../src/generated</dirName>
                 <dirName>../src/generated-test</dirName>
                 <!-- Needed for test-mda-models: -->
                 <dirName>../src/main/java</dirName>
                 <!-- Reports needed on CI to ensure consistent reporting of QA metrics: -->
-                <dirName>../target/cucumber-reports</dirName>
```
epwilkins, line 47: Q: Was this done because cucumber reports aren't in the project build directory for Habushu modules?

foundation/foundation-archetype/src/main/resources/archetype-resources/.mvn/maven-build-cache-config.xml
```diff
@@ -37,27 +37,23 @@
         <attachedOutputs>
             <dirNames>
                 <dirName>../dist</dirName>
-                <dirName>quarkus-app</dirName>
-                <dirName>helm</dirName>
-                <dirName>docker</dirName>
-                <dirName>dockerbuild</dirName>
-                <dirName>krausening</dirName>
```
epwilkins, line 44: I: These were all supporting docker rebuilds under Orphedomos, however for Fabric8 we'd want to use the `docker-build.tar`. See [my comment](https://github.com/boozallen/aissemble/issues/653#issuecomment-2821755911) on the ticket for a more in-depth explanation.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/version_specific/MavenBuildCacheMigration.java
```diff
@@ -0,0 +1,57 @@
...
+import org.technologybrewery.baton.util.FileUtils;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+
+public class MavenBuildCacheMigration extends AbstractAissembleMigration {
```
epwilkins, line 28: A: Class comment explaining migration

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/version_specific/MavenBuildCacheMigration.java
epwilkins, line 1: I: This migration only targets the `maxBuilds` change, and not the *.lock change.  The lock change is more complex to implement, and seems less likely to actually cause insidious bugs.

.mvn/maven-build-cache-config.xml
epwilkins: Yep. Reverting this change so reports from Habushu modules are picked up appropriately.

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/version_specific/MavenBuildCacheMigration.java
```diff
@@ -0,0 +1,57 @@
...
+    private static final Pattern MAX_BUILDS = Pattern.compile("( *)<maxBuildsCached>(.*?)</maxBuildsCached>");
+
+    @Override
+    protected boolean shouldExecuteOnFile(File file) {
+        if (file.getName().endsWith(".xml")) {
+            try (Stream<String> lines = Files.lines(file.toPath())) {
+                return lines.map(MAX_BUILDS::matcher)
+                        .filter(Matcher::find)
+                        .findFirst()
+                        .map(matcher -> matcher.group(2))
```
aaron-gary: Q: I'm not familiar with matcher.group(). Does the number 2 represent something meaningful in matcher.group(2)?  Would adding a constant make it easier to read?

foundation/foundation-upgrade/src/test/resources/test-files/version-specific/MavenBuildCacheMigration/validation/multiple-max-builds.xml
```diff
@@ -0,0 +1,128 @@
...
+                <dirName>krausening</dirName>
+                <!-- Reports needed on CI to ensure consistent reporting of QA metrics: -->
+                <dirName>../target/cucumber-reports</dirName>
+                <dirName>../target/surefire-reports</dirName>
+                <dirName>../target/failsafe-reports</dirName>
+            </dirNames>
+        </attachedOutputs>
+        <local>
+            <!-- NB: saving more than 1 build can result in stale docker images, as build-cache restore does not restore images -->
+            <maxBuildsCached>1</maxBuildsCached>
```
aaron-gary: Q: This file is named multiple-max-builds.xml.  I only see one instance of `<maxBuildsCached>1</maxBuildsCached>`.  Is there supposed to be a second block in this file, or is it just that the purpose of this file is to attempt to add a second?

aaron-gary (approved):

foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/version_specific/MavenBuildCacheMigration.java
epwilkins: It's the capture group from the regex, so it obeys typical regex capture group numbering where 0 is the entire match, 1 is the first set of parens etc.  I can pull it to a constant for better local contextualization

foundation/foundation-upgrade/src/test/resources/test-files/version-specific/MavenBuildCacheMigration/validation/multiple-max-builds.xml
epwilkins: multiple max builds here means a maxBuilds greater than one (probably could have named it `multiple-cached-builds`). It's 1 here because this is the validation file verifying that the migration successfully changes the >1 value to 1.

PR #713, opened 2025-05-22
[#712] improve build stability

### LINKED ISSUE #712: Feature: Resolve CI issues
## Description
Several issues:
1. Docker login is not being used consistently leading to rate limiting
2. Fabric8 needs to pick up the k8s builder created during the build setup
3. GHCR_IO_TOKEN is expired

## DOD
- [x] Figure out how to get the pre-configured builders into the Fabric8 state directory
- [x] Switch to using the built-in `secrets.GITHUB_TOKEN` where possible
- [x] CI build and other GH actions can be run successfully

## Test Strategy/Script
1. Verify build on dev was successful
2. Verify GH Pages publish was successful
3. Run Prune action and verify it succeeds

Add any other context, links, or screenshots about the feature request here.

epwilkins:
- fix logic for copying buildx dir to fabric8 state dir,
 - set snapshot repo to boozallen/aissemble in root pom so that aissemble-root and it's children can be published with the built-in action token
 - authenticate with GH action token instead of manual PAT
 - force Helm 3.17.3 to work-around helm/helm#30873

### DIFF
--- a/.github/actions/install_dependencies/action.yaml
+++ b/.github/actions/install_dependencies/action.yaml
@@ -34,7 +34,7 @@ runs:
         echo "::group::Install Helm"
         curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3
         chmod 700 get_helm.sh
-        ./get_helm.sh
+        ./get_helm.sh -v 3.17.3
         echo "::endgroup::"
     - name: Install Helm Unittest Plugin
       shell: bash
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,11 +1,3 @@
-# This workflow will build a Java project with Maven, and cache/restore any dependencies to improve the workflow execution time
-# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-java-with-maven
-
-# This workflow uses actions that are not certified by GitHub.
-# They are provided by a third-party and are governed by
-# separate terms of service, privacy policy, and support
-# documentation.
-
 name: Build aissemble
 
 on:
@@ -25,7 +17,6 @@ jobs:
   build:
     runs-on: arc-runner-set-aissemble
     env:
-      DOCKER_CONFIG: /home/runner/.docker
       RUNS_ON_S3_BUCKET_CACHE: aissemble-github-cache
       DOCKER_BUILDER_NAME: k8s-multiarch
     steps:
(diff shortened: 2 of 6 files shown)


.github/actions/install_dependencies/action.yaml
```diff
@@ -34,7 +34,7 @@ runs:
         echo "::group::Install Helm"
         curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3
         chmod 700 get_helm.sh
-        ./get_helm.sh
+        ./get_helm.sh -v 3.17.3
```
epwilkins, line 37: I: we should investigate how to incorporate the workaround of logging into `ghcr.io` (without the path component) as noted in the issue, but I couldn't get it to work off the cuff so doing this to triage.

.github/workflows/build.yml
```diff
@@ -120,6 +133,8 @@ jobs:
       - name: Run Archetype Tests
         run: |
           ./mvnw -B clean install -Parchetype-test -pl :foundation-archetype
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
epwilkins, line 137: I: While it's not immediately obvious why this is necessary, without it the docker-maven-plugin attempts to authenticate against ghcr.io with the password `${env.GITHUB_TOKEN}` which Maven then incorrectly assumes is an encrypted password (since the env var isn't available) resulting in a Maven exception.

.github/workflows/publish.yml
```diff
@@ -39,12 +39,12 @@ jobs:
       - name: Install Antora
         run: npm i antora
       - name: Generate Site
-        run: GIT_CREDENTIALS="https://${{ secrets.PAT }}:@github.com" npx antora docs/antora-playbook.yml
```
epwilkins, line 42: I: We didn't have a secret in our repo called `PAT` and this is just a left-over from before we were open-source.

pom.xml
```diff
@@ -32,6 +32,13 @@
         </developer>
     </developers>
 
+    <distributionManagement>
```
epwilkins, line 35: I: Prevents aissemble-root, build-support, and aissemble-enforcer-extension from publishing under `aissemble-parent` instead of `aissemble`, which in turn was causing a 403 error when using the GITHUB_TOKEN

.github/workflows/build.yml
```diff
@@ -100,17 +91,39 @@ jobs:
                                 --driver=kubernetes \
                                 --platform=linux/arm64 \
                                 --driver-opt="nodeselector=kubernetes.io/arch=arm64","image=docker.io/moby/buildkit:v0.19.0"
-          mkdir -p ~/.docker/fabric8/buildx && cp -R ~/.docker/buildx ~/.docker/fabric8/buildx
+          mkdir ~/.docker/fabric8 && cp -R ~/.docker/buildx ~/.docker/fabric8/buildx
```
epwilkins, line 94: I: Previous command was copying `~/.docker/buildx` to `~/.docker/fabric8/buildx/buildx`, resulting in the builders not being picked up.

chang-annie: i'm surprised this was working before without the builders!

.github/workflows/build.yml
```diff
@@ -100,17 +91,39 @@ jobs:
...
       - name: Create settings.xml
         id: create-settings-xml
         run: |
-          echo "<settings><servers><server><id>ghcr.io</id><username>${{ secrets.GHCR_IO_USERNAME }}</username><password>${{ secrets.GHCR_IO_TOKEN }}</password></server><server><id>pypi</id><username>${{ secrets.PYPI_USERNAME }}</username><password>${{ secrets.PYPI_TOKEN }}</password></server><server><id>dev-pypi</id><username>${{ secrets.TEST_PYPI_USERNAME }}</username><password>${{ secrets.TEST_PYPI_TOKEN }}</password></server> </servers></settings>" > $HOME/.m2/settings.xml
+          cat > $HOME/.m2/settings.xml << EOF
+          <settings>
+            <servers>
+              <server>
+                <id>ghcr.io</id>
+                <username>\${env.GITHUB_ACTOR}</username>
```
chang-annie, line 104: can we remove the extra `\` after `<username>` and `<password>`?

.github/workflows/release.yml
```diff
@@ -75,11 +66,35 @@ jobs:
...
+          <settings>
+            <servers>
+              <server>
+                <id>ossrh</id>
+                <username>${{ secrets.SONATYPE_CENTRAL_REPO_TOKEN_USER }}</username>
+                <password>${{ secrets.SONATYPE_CENTRAL_REPO_TOKEN_KEY }}</password>
+              </server>
+              <server>
+                <id>ghcr.io</id>
+                <username>\${env.GITHUB_ACTOR}</username>
```
chang-annie, line 81: same with the `\` here

.github/workflows/release.yml
```diff
@@ -75,11 +66,35 @@ jobs:
...
+                <password>${{ secrets.PYPI_TOKEN }}</password>
+              </server>
+            </servers>
+          </settings>
+          EOF
+          echo "HOME_DIR=$HOME" >> "$GITHUB_OUTPUT"
       - name: Release aiSSEMBLE
         run: | 
-          ./mvnw release:clean release:prepare release:perform -s $HOME/.m2/settings.xml -U -B -Pci,gh-build -DpushChanges=true -DreleaseVersion=${{ env.RELEASE_VERSION }} -DdevelopmentVersion=${{ env.NEXT_DEVELOPMENT_VERSION }} -Dgpg.keyname=aissemble@bah.com -DsignTag=true -Dmaven.build.cache.enabled=false
+          ./mvnw release:clean release:prepare release:perform -U -B -Pci,gh-build -DpushChanges=true -DreleaseVersion=${{ env.RELEASE_VERSION }} -DdevelopmentVersion=${{ env.NEXT_DEVELOPMENT_VERSION }} -Dgpg.keyname=aissemble@bah.com -DsignTag=true -Dmaven.build.cache.enabled=false
```
chang-annie, line 95: i'm not too familiar with our release process, do we need to specify `--settings $HOME/.m2/settings.xml` back in here?

chang-annie (approved):

.github/workflows/build.yml
epwilkins: The backslash is required to escape the `$` for bash, since this is inside a heredoc

.github/workflows/release.yml
epwilkins: That's the default location that Maven looks at to find `settings.xml`, so it's redundant That's why I removed it.

.github/workflows/build.yml
epwilkins: It wasn't. 😅 Technically the builds _should_ be able to succeed without the k8s builders, as it should fall back to `docker-container` which is just super slow because it has to emulate the non-native architecture. But `aissemble-jenkins-agent` has something that the Qemu emulator really doesn't like so it does fail.

Prior to the failures this week, the build was picking up the builders, but it wasn't authenticating with DockerHub for image pulls, resulting in us hitting the docker rate limit again ~Tuesday.  Copying the buildx dir to a separate directory and having fabric8 use that separate directory is how we preserve both the builders and login credentials.

PR #734, opened 2025-06-04
[#733] incorporate dependapot suggestions

### LINKED ISSUE #733: Feature: Incorporate new Dependabot suggestions
## Description
Pull in package updates suggested by Dependabot

## DOD
- [x] Libraries are updated and new dependabot suggestions are resolved

## Test Strategy/Script
1. Verify the build passes

epwilkins:
Primarly just touches Maven plugins like the archetype. Because we've add Habushu as a dependency in order to help with build cache calculations, a transitive dependency there was making dependabot think we were using a vulnerable version of a library in our application code. I added exclusions to the dependencies we added for this purpose just to reduce the noise.

This exposed a few places where we were actually relying on the transitive dependencies brought in by those for, namely, Krausening and aether-api.

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -20,7 +20,7 @@
         <!-- **************** -->
         <!-- Plugins versions -->
         <!-- **************** -->
-        <version.archetype.plugin>3.2.1</version.archetype.plugin>
+        <version.archetype.plugin>3.3.0</version.archetype.plugin>
         <version.assembly.plugin>3.7.1</version.assembly.plugin>
         <version.buildhelper.plugin>3.6.0</version.buildhelper.plugin>
         <version.buildnumber.plugin>3.2.1</version.buildnumber.plugin>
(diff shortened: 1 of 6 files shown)


build-parent/pom.xml
```diff
@@ -185,18 +185,36 @@
             <artifactId>fermenter-mda</artifactId>
             <version>${version.fermenter}</version>
             <scope>provided</scope>
+            <exclusions>
```
epwilkins, line 188: I: Exclusions added here to reduce noise from false positives.

extensions/extensions-data-delivery/extensions-data-delivery-spark-postgres/pom.xml
```diff
@@ -19,6 +19,10 @@
             <artifactId>postgresql</artifactId>
             <version>${version.postgresql}</version>
         </dependency>
+        <dependency>
+            <groupId>org.technologybrewery.krausening</groupId>
+            <artifactId>krausening</artifactId>
```
epwilkins, line 24: I: Missing declared dependency that was removed from transitives after exclusions above.

foundation/foundation-archetype/pom.xml
```diff
@@ -60,9 +66,9 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>net.masterthought</groupId>
-            <artifactId>cucumber-reporting</artifactId>
-            <version>${version.cucumber.reporting.plugin}</version>
+            <groupId>org.eclipse.aether</groupId>
+            <artifactId>aether-api</artifactId>
```
epwilkins, line 70: I: Missing declared dependency that was removed from transitives after exclusions above. Re-organized to put cucumber-reporting above because the "Supports archetype" comment applies to wagon and aether.

csun-cpointe (approved):

cwoods-cpointe (approved):

Issue #736, opened 2025-06-05
Feature: Update downstream projects to use version property for aissemble dependencies


cwoods-cpointe:
## Description
Downstream projects generated in 1.12 have aissemble dependencies in their root pom that have a hard coded version. They need to be updated to the version.aissemble property. The archetype generation for 1.13 has already been updated so we just need a Baton migration to update existing projects

## DOD
- [x] Baton migration exists to update root pom aissemble dependencies version to the `version.aissemble` property
- [x] Release draft is update to include the baton migration

### Feature Scenarios
```feature
Scenario: A project that has aissemble dependencies with hard coded version are migrated to use maven properties
  Given a project with hardcoded aissemble versions in the root pom
  When the migration is ran
  Then the aissemble versions are update to use the maven property

Scenario: A project that has aissemble dependencies that are use the maven property are not migrated
  Given a project with aissemble versions using the maven property in the root pom
  When the migration is ran
  Then the dependency version migration is skipped
```

### Test steps
1. Create a downstream project
```
mvn archetype:generate '-DarchetypeGroupId=com.boozallen.aissemble' \
                           '-DarchetypeArtifactId=foundation-archetype' \
                           '-DarchetypeVersion=1.12.1' \
                           '-DgroupId=org.test' \
                           '-Dpackage=org.test' \
                           '-DprojectGitUrl=test.org/test-736-migration' \
                           '-DprojectName=test-736-migration' \
                           '-DartifactId=test-736-migration' \
    && cd test-736-migration
```
2. Build the project - `./mvnw clean install`
3. Update the root pom build parent to use aissemble version `1.13.0-SNAPSHOT`
4. Run the migrations - `./mvnw org.technologybrewery.baton:baton-maven-plugin:baton-migrate`
5. Verify these root pom dependencies versions are using the property `version.aissemble`:
    1. artifacts-maven-plugin
    2. foundation-model-plugin
    3. mda-maven-plugin
    4. foundation-mda
    5. foundation-upgrade
6. Verify the dependencies are valid via a successful build -`./mvnw clean install`
7. In the output verify the migration is processed successfully:
```
[INFO] --- baton:1.1.1:baton-migrate (default-cli) @ test-736-migration ---
[INFO] Loading migrations from: jar:file:/blah/.m2/repository/com/boozallen/aissemble/foundation-upgrade/1.13.0-SNAPSHOT/foundation-upgrade-1.13.0-SNAPSHOT.jar!/migrations.json
[...]

PR #737, opened 2025-06-06
[#736] Create Baton migration for projects root aissemble depenencies

cwoods-cpointe:
The migration replaces the defined version for aissemble dependencies in the root pom with the maven property.


foundation/foundation-upgrade/src/main/java/com/boozallen/aissemble/upgrade/migration/version_specific/RootDependencyVersionMigration.java
```diff
@@ -0,0 +1,76 @@
...
+     * Will execute if the pom is the root pom file
+     * and if the aissemble dependency versions have not already been updated
+     *
+     * @param pomFile file to check
+     * @return true if the migration should execute
+     */
+    @Override
+    protected boolean shouldExecuteOnFile(File pomFile) {
+        Model projectModel = getMavenProject().getModel();
+        if (projectModel != null && projectModel.getParent() != null && AISSEMBLE_PARENT.equals(projectModel.getParent().getArtifactId())) {
```
epwilkins: A: isn't this just getRootProject().equals(getMavenProject())?

epwilkins: Oh testing is harder. I think you can extend the migration to override these methods just to trivially return the same empty project, or set the parent value like we did in some of the other migration test extensions in this version

epwilkins (approved):

Issue #780, opened 2025-09-23
Feature: Update Habushu to v3.3.0-SNAPSHOT


lemuelwatts:
## Description
Update aissemble to use latest Habushu version.

## DOD
Acceptance criteria required to realize the requested feature
- [x] Update aiSSEMBLE to use Habushu 3.3.0-SNAPSHOT
- [x] Update the release notes to reflect the new Habushu version
- [x] Ensure all pom templates reflect new Habushu v3.3.0-SNAPSHOT

## Test Strategy/Script
1. Switch to feature branch ([feature/habushu-3-3-0-upgrade](https://github.com/boozallen/aissemble/tree/feature/habushu-3-3-0-upgrade)) and verify you have the latest changes
2. Build aissemble `mvn clean install`
3. Run archetype test with `mvn clean install -pl :foundation-archetype -Parchetype-test`

lemuelwatts:
DOD with @epwilkins

lemuelwatts:
OTS with @epwilkins

csun-cpointe:
Final test passed! 
- **Baseline build passed**

<img width="187" height="510" alt="Image" src="https://github.com/user-attachments/assets/13dcf5cf-113d-4188-ae72-433bf3bfbbec" />

- **Archtype test passed**

<img width="425" height="76" alt="Image" src="https://github.com/user-attachments/assets/2eca564a-892b-4c47-8e5c-5f051b6308d8" />

Issue #779, opened 2025-09-23
BUG: Habushu 3 updates left some pyproject.toml files misconfigured


epwilkins:
## Description
The updates made to update to Habushu 3+ and Poetry 2+ (including #687 and #708) resulted in some misconfigurations around packaging.  Largely, this is due to the change in behavior of the "include" mechanism so that files are never included in wheels unless the format is explicitly declared.

## Definition of Done
 - [x] update aissemble-foundation-messaging-python-client to include generated classpath in wheel
 - [x] update aissemble-foundation-data-lineage-python to include default_properties in wheel
 - [x] update pyproject.toml velocity templates to include generated source in wheel
 - [x] update test-mda-model pyproject.toml files to include generated source in wheel

## Test Steps
1. Build foundation-messaging and data-lineage
    ```sh
    mvn clean install -pl :aissemble-foundation-messaging-python-client,:aissemble-foundation-data-lineage-python
    ```
2. List contents in messaging wheel and verify classpath jars exist
    ```sh
    unzip -l foundation/foundation-messaging/foundation-messaging-python/aissemble-foundation-messaging-python-client/dist/aissemble_foundation_messaging_python-1.13.0.dev0-py3-none-any.whl
    ```
3. List contents in data-lineage wheel and verify default_properties exist
    ```sh
    unzip -l foundation/foundation-lineage/foundation-data-lineage/aissemble-foundation-data-lineage-python/dist/aissemble_foundation_data_lineage_python-1.13.0.dev0-py3-none-any.whl
    ```
4. Build the test modules
    ```sh
    mvn clean install -pl :test -amd
    ```
5. List the contents of the test wheels and verify they have source code
    ```sh
    unzip -l test/test-mda-models/aissemble-test-data-delivery-pyspark-model/dist/aissemble_test_data_delivery_pyspark_model-1.13.0.dev0-py3-none-any.whl
    unzip -l test/test-mda-models/test-machine-learning-model/aissemble-machine-learning-training/dist/aissemble_machine_learning_training-1.13.0.dev0-py3-none-any.whl
    unzip -l test/test-mda-models/test-machine-learning-model/aissemble-machine-learning-inference/dist/aissemble_machine_learning_inference-1.13.0.dev0-py3-none-any.whl
    ```
6. Run the archetype test
    ```sh
    mvn clean install -pl :foundation-archetype -Parchetype-test
    ```
7. Navigate to the root of the test generator project
    ```sh
[...]

PR #789, opened 2025-09-25
[#788] migrate poetry includes

### LINKED ISSUE #788: Feature: Migrate generated source configuration for Python modules
## Description
Poetry 2 changed the behavior of `tool.poetry.include`.  Previously, if the include was a directory it was only included in sdist, and if it was a file/files in sdist and wheel. Now, include is always only included in sdist by default.

We updated the baseline source and templates to account for this in #779, but we need a migration for existing projects that upgrade.

## DOD
- [x] A migration exists that replaces the old default include to one that specifies formats
- [x] New migration is added to draft release notes 
- [x] Conditional step for projects that have customized the include config is added to draft release notes

## Test Strategy/Script
1. Create a 1.12.1 PySpark or ML project
2. Set the build-parent version to 1.13.0-SNAPSHOT in the root pom.xml
3. Run the upgrade migrations
    ```sh
    ./mvnw org.technologybrewery.baton:baton-maven-plugin:baton-migrate
    ```
4. Verify the pyproject.toml files were updated with the new format
    ```diff
    -include = ["src/module_name/generated/**/*"]
    +include = [{path = "src/module_name/generated/**/*", format = ["sdist", "wheel"]}]
    ```

epwilkins:
This also includes the Habushu 3.3 feature branch changes as a separate commit since the release has been cut.

- (26fbdd9e) [[#780](../issues/780)] upgrade habushu to v3.3.0
- (fa5b11d4) [[#788](../issues/788)] add migration to specify format in tool.poetry.include
    This migration only looks for include lines that are fairly close to the
    default that is generated by aiSSEMBLE.  This should cover a large
    majority of cases, but for the edge cases a new conditional step has
    been added to the upgrade instructions telling projects to account for
    the new Poetry 2 behavior manually.

    Also fixes the template updates from [#779](../issues/779) to account for the fact that
    the default project layour is under `src`.


foundation/foundation-mda/src/main/resources/templates/data-delivery-data-records/combined.pyproject.toml.vm
```diff
@@ -9,7 +9,7 @@ authors = [{name = "Your Name", email = "<you@example.com>"}]
 requires-python = ">=3.9"
 
 [tool.poetry]
-include = [{path = "${artifactIdPythonCase}/generated/**/*.py", format = ["sdist", "wheel"]}]
+include = [{path = "src/${artifactIdPythonCase}/generated/**/*.py", format = ["sdist", "wheel"]}]
```
epwilkins, line 12: I: bug fix for #779

foundation/foundation-upgrade/src/test/resources/test-files/version-specific/Poetry2IncludeMigration/migration/pyproject.toml
```diff
@@ -0,0 +1,36 @@
...
+
+[project]
+name = "my-pipeline"
+version = "1.0.0.dev"
+description = "${pipeline.description|'Description of package'}"
+authors = [{name = "Your Name", email = "<you@example.com>"}]
+requires-python = ">=3.9"
+
+[tool.poetry]
+include = ["src/my_pipeline/generated/**/*"]
```
epwilkins, line 11: I: this is the line that changes after the migration

cwoods-cpointe (approved):

PR #792, opened 2025-09-26
[#791] fix issues with parallel build

### LINKED ISSUE #791: Feature: Enable parallel builds
## Description
To improve developer efficiency and enable the use of `mvnd`, we should make our build thread-safe so that parallel builds can be used.

## DOD
- [x] Project can build successfully on mulitple threads (mvnd, or `-T`)
- [x] Archetype test is setup to use parallel builds
- [x] CI build is setup to use parallel builds

## Test Strategy/Script
1. Verify [the build](https://github.com/boozallen/aissemble/actions/runs/18042451994) succeeds and is running on multiple threads
    - The build command in CI should include `-T8` to enable 8 build threads
2. Verify the archetype test succeeds

epwilkins:
- remove docker cache save/restore in CI
 - Add parallel build flag to CI build
 - Add extra replicas to docker k8s builder
 - Upgrade krausening, baton, and build-cache for better parallel support
 - Deconflict ITs by using dynamic ports
 - Add parallel build flag to archetype test
 - Make MDA model managers thread-safe
 - Ensure no baton migrations are needed for pyproject.toml templates
 - Remove spurious `RUN chown` in Spark worker docker template
 - Move report aggregator to separate profile to support parallel builds

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,16 +41,6 @@ jobs:
         with:
           docker-username: ${{ secrets.DOCKERHUB_USERNAME }}
           docker-token: ${{ secrets.DOCKERHUB_TOKEN }}
-      #NB: We restore/save cache manually so that we save the cache even if the build fails
-      - name: Load docker build cache
-        id: cached-docker-build
-        uses: runs-on/cache/restore@v4
-        if: ${{ ! github.event.schedule }}
-        with:
-          path: ~/.docker/cache
-          key: docker-cache-${{ hashFiles('**/Dockerfile') }}
-          restore-keys: |
-            docker-cache-
       - name: Load m2 repository cache # Manually caching .m2 repo as the setup-java caching isn't falling back to older caches
         id: cached-m2-repo
         uses: runs-on/cache/restore@v4
@@ -75,12 +65,6 @@ jobs:
           key: poetry-cache-${{ hashFiles('**/pyproject.toml') }}
           restore-keys: |
             poetry-cache-
-      - name: Hash starting Docker cache
-        id: docker-cache-start
-        uses: actions/github-script@v7
-        with:
-          result-encoding: string
-          script: await glob.hashFiles(`~/.docker/cache/**`, require('os').homedir())
       - name: Provision Docker builder
         id: provision-docker-builder
         run: |
@@ -88,13 +72,13 @@ jobs:
                                 --bootstrap \
                                 --driver=kubernetes \
                                 --platform=linux/amd64 \
-                                --driver-opt="nodeselector=kubernetes.io/arch=amd64","image=docker.io/moby/buildkit:v0.19.0"
+                                --driver-opt="nodeselector=kubernetes.io/arch=amd64","image=docker.io/moby/buildkit:v0.19.0",replicas=3
           docker builder create --name=$DOCKER_BUILDER_NAME \
                                 --append \
                                 --bootstrap \
                                 --driver=kubernetes \
                                 --platform=linux/arm64 \
-                                --driver-opt="nodeselector=kubernetes.io/arch=arm64","image=docker.io/moby/buildkit:v0.19.0"
+                                --driver-opt="nodeselector=kubernetes.io/arch=arm64","image=docker.io/moby/buildkit:v0.19.0",replicas=3
           mkdir ~/.docker/fabric8 && cp -R ~/.docker/buildx ~/.docker/fabric8/buildx
       # Generate the settings.xml for ghcr.io, pypi, & dev-pypi server profiles
       - name: Create settings.xml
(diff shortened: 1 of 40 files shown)


.github/workflows/build.yml
```diff
@@ -75,26 +75,20 @@ jobs:
           key: poetry-cache-${{ hashFiles('**/pyproject.toml') }}
           restore-keys: |
             poetry-cache-
-      - name: Hash starting Docker cache
```
epwilkins, line 78: I: There is some bug causing the docker cache to just never save/restore (I think the formatting of the `hashFiles` command) but when the cache is populated it takes 10-15 mins to restore and save. If our build time is only 47 mins from scratch that adds a lot of churn for not a ton of benefit, especially since the maven build cache will skip most of the docker builds most off the time.

foundation/foundation-archetype/src/main/resources/archetype-resources/.mvn/maven-build-cache-config.xml
```diff
@@ -78,6 +78,12 @@
                         <goal>deploy</goal>
                     </goals>
                 </goalsList>
+                <!-- necessary for calculating the Python version string for use in Docker -->
```
epwilkins, line 81: I: This property actually isn't needed at all anymore with Habushu containerization, but it was too entangled to change in this ticket.  Also, I'm not sure what changed between build cache 1.1 and 1.2 but in 1.2 this is required where it wasn't in 1.1

foundation/foundation-archetype/test-project-archetype.sh
```diff
@@ -27,7 +27,7 @@ cd target/temp
 #---
 function exit {
   if [ "$1" -ne 0 ]; then
-    rm -rf ~/.m2/build-cache/v1/com.bah.aiops
+    rm -rf ~/.m2/build-cache/v1.1/com.bah.aiops
```
epwilkins, line 30: I: Confusingly, this is the 1.2 cache location

foundation/foundation-configuration-store/src/main/resources/application.properties
```diff
@@ -9,4 +9,4 @@
 ###
 
 quarkus.http.port=8080
-quarkus.http.test-port=14800
+quarkus.http.test-port=0
```
epwilkins, line 12: I: `0` means choose a port dynamically

foundation/foundation-mda/src/main/java/com/boozallen/aiops/mda/metamodel/DictionaryModelInstanceManager.java
```diff
@@ -24,7 +24,7 @@
  */
 class DictionaryModelInstanceManager extends AbstractMetamodelManager<Dictionary> {
 
-    private static final DictionaryModelInstanceManager instance = new DictionaryModelInstanceManager();
+    private static final ThreadLocal<DictionaryModelInstanceManager> instance = ThreadLocal.withInitial(DictionaryModelInstanceManager::new);
```
epwilkins, line 27: I: Allows Fermenter with `foundation-mda` to run concurrently across modules.

pom.xml
```diff
@@ -266,6 +263,12 @@
         </pluginRepository>
     </pluginRepositories>
     <profiles>
+        <profile>
+            <id>create-test-report</id>
```
epwilkins, line 267: I: It really didn't make sense to me to list every single module as a dependency of the aggregator module and hope that it stays in sync. Instead we just move it to a profile that can be run separately after the main build completes. We aren't actually using these reports in CI atm, but if we ever did we'd just need to add a build step to run this profile.

extensions/extensions-data-lineage/extensions-data-lineage-http-consumer-service/src/test/resources/application.properties
```diff
@@ -8,9 +8,8 @@
 # #L%
 ###
 quarkus.rest-client."com.boozallen.aissemble.datalineage.consumer.HttpProducerService".uri=http://127.0.0.1:15100/endpoint
-quarkus.amqp.devservices.image-name=quay.io/artemiscloud/activemq-artemis-broker:1.0.25
-#Artemis image requires --user arg to startup and this fixes the issue, see https://github.com/quarkusio/quarkus/issues/36190
-quarkus.amqp.devservices.extra-args=--no-autotune --mapped --no-fsync --relax-jolokia
-quarkus.kafka.devservices.image-name=redpandadata/redpanda:v24.1.7
+quarkus.rabbitmq.devservices.enabled=false
+quarkus.kafka.devservices.enabled=false
+quarkus.amqp.devservices.enabled=false
```
epwilkins, line 13: I: None of these services were needed for the ITs, and it caused a ton of churn on my machine trying to start them all during a parallel build with 15 threads (which is admittedly probably too much parallelization on my machine)

aaron-gary (approved):

